PR #1, opened 2023-03-17
Draft: Add workflow for github actions

eulovi:


.github/workflows/testing.yml
```diff
@@ -0,0 +1,159 @@
...
+          source ${PYTHON_VENV}/bin/activate
+          # We have to update pip, otherwise there is a bug in the compilation of the cython code
+          pip install --upgrade pip
+          # Install meshpy
+          pip install .[CI-CD]
+          # Install cubitpy
+          cd $CUBITPY_REPO
+          git checkout master
+          git pull
+          pip install .
```
isteinbrecher, line 60: The repository of cubitpy seems to cause issues. You can try to simply pull cubitpy from the github repostory here (I would assume the runner has access rights).

This should look something like this:
```
cd $MESHPY_PATH
git clone git@github.com:imcs-compsim/cubitpy.git
pip install cubitpy/
```

.github/workflows/testing.yml
```diff
@@ -0,0 +1,159 @@
...
+          coverage run --rcfile=coverage.config testing_main.py
+          coverage html
+          coverage report
+          coverage-badge -o htmlcov/coverage.svg
+          # Check codestyle
+          cd $MESHPY_PATH
+          black . --check --exclude="${PYTHON_VENV}" && exit 0
+          # If we did not exit earlier, raise an error here
+          exit 1
+  geometric-search-testing:
```
isteinbrecher, line 76: Add a newline here for better readability

.github/workflows/testing.yml
```diff
@@ -0,0 +1,159 @@
...
+          # Install meshpy, it is not yet possible to globally install the ArborX binaries, so we
+          # use the developement mode here.
+          pip install -e .
+          # Print information on the python environment
+          python --version
+          pip list
+          # Run tests
+          cd tests
+          python3 testing_geometric_search.py
+  performance-test:
```
isteinbrecher, line 113: See above

.github/workflows/testing.yml
```diff
@@ -0,0 +1,159 @@
...
+      - name: For the master branch, add the coverage report and the website
+        run: |
+          mkdir public
+          mkdir public/htmlcov
+          mv tests/htmlcov/ public/
+          cd website
+          bundle exec jekyll build -d public
+          cd public
+          mv * ../../public
+
```
isteinbrecher, line 159: This can be removed since it is related to the meshpy website. Maybe we can just create a similar website here on Github?

eulovi:
Currently, the testing for meshpy is done with its corresponding workflow in `testing.yml `. The build and deployment of the pages is done in `github-pages.yml`. It's not possible to check if the deployment is done since this branch doesn't have permissions for deployment: only the `main` branch can do this. But I believe this should work when this is merged to `main`.

PR #16, opened 2023-10-11
Add beam contact boundary condition

maxfirmbach:
Adding boundary condition description and routine for beam to beam contact. This makes the creation of respective scenarios a little bit more convenient (hopefully).


maxfirmbach:
Functionality can also be provided in a less hard coded way, thus closing.

PR #42, opened 2024-01-05
Implement the base of a sphere and test it

eulovi:
A patch describing a section of the surface of a sphere is implemented. This patch can be rotated to create a complete sphere.


meshpy/mesh_creation_functions/nurbs_geometries.py
```diff
@@ -218,6 +218,125 @@ def create_nurbs_flat_plate_2d(width, length, *, n_ele_u=1, n_ele_v=1):
     return surf
 
 
+def create_nurbs_base_sphere_patch_surf(radius, n_ele_u=1, n_ele_v=1):
+    """
+    This function does a base patch to create a sphere.
+    This patch represents a surface.
```
isteinbrecher, line 224: A little better description is needed here, maybe:
"""
Generates a patch of a sphere as a NURBS surface. This function constructs a segment of a spherical surface using Non-Uniform Rational B-Splines (NURBS) based on the specified radius and the number of elements in the parametric u and v directions. 
"""

meshpy/mesh_creation_functions/nurbs_geometries.py
```diff
@@ -218,6 +218,125 @@ def create_nurbs_flat_plate_2d(width, length, *, n_ele_u=1, n_ele_v=1):
...
+    n_ele_v: int
+        number of elements in the parametric v-direction
+
+    Return
+    ----
+    surf: geomdl object
+        geomdl object that contains the surface information
+    """
+
+    # Create a NURBS surface instance
```
isteinbrecher, line 241: As far as I can tell this function creates a patch with the segment angle 90 and 90, i.e., 6 patches are needed to build a full sphere, right?

Is it also possible to let the user specify the angle?

Dis you come up with this implementation or is this taken from some literature, forum, ...? If so, please add a short reference.

tests/testing_nurbs.py
```diff
@@ -258,6 +259,34 @@ def test_nurbs_couple_nurbs_meshes(self):
         # Compare with the reference file
         compare_test_result(self, input_file.get_string(header=False))
 
+    def test_create_nurbs_base_sphere_patch_surf(self):
```
isteinbrecher, line 262: Use the same name here as the function will have.

tests/testing_nurbs.py
```diff
@@ -258,6 +259,34 @@ def test_nurbs_couple_nurbs_meshes(self):
...
+        """Test the creating of the base patch of the surface of a sphere."""
+
+        # Create input file
+        input_file = InputFile()
+
+        # Create the base of a sphere
+        surf_obj = create_nurbs_base_sphere_patch_surf(1, n_ele_u=3, n_ele_v=2)
+
+        # Create first patch set
+        mat = MaterialStVenantKirchhoff(youngs_modulus=430, nu=0.10, density=4.2e-7)
```
isteinbrecher, line 272: You can leave this empty, as these values are not relevant for the test and the functionality of `MaterialStVenantKirchhoff` is already tested.

isteinbrecher (commented):
Thanks @eulovi for the contribution, I added some comments.

meshpy/mesh_creation_functions/nurbs_geometries.py
eulovi: Exactly, each segment has an angle 90 and 90 and with 6 patches a full sphere is built. I took the implementation from the code of the `NURBS-generator` so that I can add it as a reference. For now, the user can't specify the angle, as I'm not sure how to determine the weights of the control points depending on the angle.

meshpy/mesh_creation_functions/nurbs_geometries.py
```diff
@@ -218,6 +218,125 @@ def create_nurbs_flat_plate_2d(width, length, *, n_ele_u=1, n_ele_v=1):
     return surf
 
 
+def create_nurbs_base_sphere_patch_surf(radius, n_ele_u=1, n_ele_v=1):
```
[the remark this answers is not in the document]
eulovi: Thanks for the comments! I changed it to `create_nurbs_sphere_surface`.

tests/testing_nurbs.py
eulovi: I changed it to `test_nurbs_sphere_surface_reference`.

isteinbrecher (approved):

PR #52, opened 2024-04-03
Generalize nurbs

isteinbrecher:
This merge allows to set the "element string" for a NURBS patch directly in the script, i.e., we don't have to implement all relevant cases.
Similarly, now there is a class `MaterialString` which allows to directly set a material from the calling script so we don't have to implement each material case in MeshPy.

### DIFF
--- a/meshpy/__init__.py
+++ b/meshpy/__init__.py
@@ -50,6 +50,7 @@ from .material import (
     MaterialKirchhoff,
     MaterialReissner,
     MaterialReissnerElastoplastic,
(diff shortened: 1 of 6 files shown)


isteinbrecher:
@eulovi this might also be a good way to use the MeshPy NURBS with the new solid elements.

eulovi:
Thanks for implementing this @isteinbrecher! I think this is a nice solution to add the new solid elements :)

eulovi (dismissed):

PR #60, opened 2024-04-22
Add helical beam basic geometry function

davidrudlstorfer:
To enable the usage of meshpy within my own input file creation repository it is helpful to create a simple helical beam base geometry. This pull request introduces a new helical mesh creation function with the following parameters (crucial for the helical geometry):

1. **Axis**: Central axis of helix
2. **Axis point**: Point on central axis which is does not have to be aligned with the bottom plane of the helical geometry
3. **Start point**: Start point of the helical geometry. Also determines the radius of the geometry.
4. **Twist angle**: Angle of helical geometry (also known as pitch angle)
5. Either **height of helix** or **number of turns**: Within my repository the total height of the helix is an important quantity. As an alternative I also added the number of turns as a second parameterization option.

I tested the mesh creation function in general orientations and setups and everything worked as expected. If the general approach is correct I would also add a test.


meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,171 @@ def create_beam_mesh_arc_at_node(
...
+        np.dot(start_point_origin_vec, start_point_origin_vec_simple_rot), -1
+    ):
+        # vectors point in opposite direction
+        R_start_vec = np.asarray([[-1, 0, 0], [0, 1, 0], [0, 0, -1]])
+    else:
+        v = np.cross(start_point_origin_vec, start_point_origin_vec_simple_rot)
+        c = np.dot(start_point_origin_vec, start_point_origin_vec_simple_rot)
+        v_x = np.asarray([[0, -v[2], v[1]], [v[2], 0, -v[0]], [-v[1], v[0], 0]])
+        R_start_vec = np.eye(3) + v_x + np.matmul(v_x, v_x) / (1 + c)
+
```
davidrudlstorfer, line 645: Here, I need to determine the rotation between the simple helix and the wanted helical geometry. I used the approach described here: https://math.stackexchange.com/questions/180418/calculate-rotation-matrix-to-align-vector-a-to-vector-b-in-3d. There's definitely a shorter and more elegant solution than the long and elaborate way I did it using quaternions. I was not able to find something predefined within meshpy so if someone has a suggestion on how to improve this I would be more than happy to know.

isteinbrecher: I think this can be simplified drastically:
```python
rotation = Rotation.from_basis(axis, start_point_origin_vec)*Rotation([0,1,0],np*0.5)*Rotation([0,0,1],np.pi*0.5)
mesh.rotate(rotation)
```
Be aware of the singularity if `norm(start_point_origin_vec) == 0`, related to the question above regarding the case `radius == 0`.

isteinbrecher (commented):
Thanks @davidrudlstorfer for the contribution! The interface looks good, I left some comments on the implementation of the function itself. As you already mentioned, a test case (or cases to cover some corner cases) should be added, at some point in this pull-request.

Although there should be no issues at the moment, I am starting a pipeline.

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,171 @@ def create_beam_mesh_arc_at_node(
...
+    **kwargs (for all of them look into create_beam_mesh_function)
+    ----
+    n_el: int
+        Number of equally spaced beam elements along the line. Defaults to 1.
+        Mutually exclusive with l_el.
+    l_el: float
+        Desired length of beam elements. This requires the option interval_length
+        to be set. Mutually exclusive with n_el. Be aware, that this length
+        might not be achieved, if the elements are warped after they are
+        created.
```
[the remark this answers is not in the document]
isteinbrecher: Nevermind, this is done automatically in the create_beam_mesh_line function.

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: Wow that's a really nice solution! See my comment down below.

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,171 @@ def create_beam_mesh_arc_at_node(
...
+
+    Return
+    ----
+    return_set: GeometryName
+        Set with the 'start' and 'end' node of the line. Also a 'line' set
+        with all nodes of the line.
+    """
+
+    if not 0 < twist_angle < 2 * np.pi:
+        raise ValueError("Twist angle of helix must be between 0 and 2pi!")
```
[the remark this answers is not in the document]
davidrudlstorfer: Thanks for the hint - you're correct and I've removed the check entirely.

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,171 @@ def create_beam_mesh_arc_at_node(
...
+    if height_helix:
+        end_point = np.array([radius, height_helix / np.tan(twist_angle), height_helix])
+    elif turns:
+        end_point = np.array(
+            [
+                radius,
+                2 * np.pi * radius * turns,
+                2 * np.pi * radius * turns * np.tan(twist_angle),
+            ]
+        )
```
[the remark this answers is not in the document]
davidrudlstorfer: Sounds like a very good plan. I'm currently trying your supposed approach and unfortunately I am not able to get it working.

I definitely want to allow the case of `radius == 0`. Therefore, I want to return a straight geometry if the input results in a straight line. If I pursue your supposed approach and create a mesh line and wrap it around a cylinder it results in an error as the `radius_phi` within the `wrap_cylinder` function is zero. The same happens if i define the `radius=0.0` for the `wrap_cylinder` function.

If it's not a problem I would just keep the special case in which the straight line is returned.

Regarding the triads: As far as I understand the triad orientation does not result in any differences down the road (simulation)? Is it a problem if they are just "random"?

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: Whoops - wasn't aware that at Github the comments are connected and if i resolve the first one - this one is resolved as well - Sorry!

I've actually simplified your approach further to only use one rotation in the latest commit. Just one theoretical question: Here, we get the rotation necessary to rotate the helical coordinate system into the global "simple" coordinate system. Why do we not need to inverse the rotation/rotation matrix here? Initially I thought we need to reverse the rotation because we transform the simple geometry into the defined axis/start point.

davidrudlstorfer:
During the changes of your review the following two questions came up:

1. If the warning is printed with `warnings.warn` the following commandline output is written out:

```
/home/rudlstorfer/01_work/01_workspace/meshpy/meshpy/mesh_creation_functions/beam_basic_geometry.py:566: UserWarning: Radius of helix is 0 or twist angle is 90 degrees! Simple line geometry is returned!
  warnings.warn(
  ```
  
  At the end there's always an unnecessary `warnings.warn(`. Have you encountered that problem?
  
  2. I use black v22.12.0 to format my code. Yesterday you started one pipeline run and it failed due to unformatted code. Are there any specific settings within MeshPy I haven't found?

File: meshpy/mesh_creation_functions/beam_basic_geometry.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (isteinbrecher) [reply meshpy/mesh_creation_functions/beam_basic_geometry.py:590]: You are right, the `wrap_cylinder` function will complain for a radius of 0. Then I suggest you leave this as is, this will also resolve the issue with the "undefined" triads.

Regarding your question, if the beams cross-section is double-symmetric the orientation of the triad does not change the results - in the continuous setting, in the discrete setting there might be some (small) differences.

File: meshpy/mesh_creation_functions/beam_basic_geometry.py [hunk 5]
```diff
@@ -482,3 +483,128 @@ def create_beam_mesh_arc_at_node(
...
+    # determine radius of helix
+    axis = np.array(axis) / np.linalg.norm(np.array(axis))
+    origin = axis_point + np.dot(
+        np.dot((np.array(start_point) - np.array(axis_point)), axis), axis
+    )
+    start_point_origin_vec = start_point - origin
+    radius = np.linalg.norm(start_point_origin_vec)
+
+    # return line if radius of helix is 0 or twist angle is np.pi/2
+    if np.isclose(radius, 0) or np.isclose(twist_angle, np.pi / 2):
```
isteinbrecher, line 558: If general angles are allowed this should be:
```python
if np.isclose(radius, 0) or np.isclose(np.cos(twist_angle), 0.0):
```

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,128 @@ def create_beam_mesh_arc_at_node(
...
+            "Radius of helix is 0 or twist angle is 90 degrees! "
+            + "Simple line geometry is returned!"
+        )
+
+        return create_beam_mesh_line(
+            mesh,
+            beam_object,
+            material,
+            start_point=start_point,
+            end_point=start_point + height_helix * axis,
```
isteinbrecher, line 576: To be consistent with the case of "negative" slopes, this should be
```python
end_point=start_point + height_helix * axis * np.sin(twist_angle),
```

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -482,3 +483,128 @@ def create_beam_mesh_arc_at_node(
...
+        end_point=end_point,
+        **kwargs,
+    )
+
+    mesh.wrap_around_cylinder()
+
+    # rotate and translate simple helix to align with neccessary axis and starting point
+    mesh.rotate(
+        Rotation.from_basis(start_point_origin_vec, axis)
+        * Rotation([1, 0, 0], -np.pi * 0.5)
```
isteinbrecher, line 606: Nice rotation simplification!

tests/testing_mesh_creation_functions.py
```diff
@@ -737,6 +738,99 @@ def line(t):
         # Check the output.
         compare_test_result(self, input_file.get_string(header=False))
 
+    def test_mesh_creation_functions_helix_no_rotation(self):
+        """Create a helix and compare it with the reference file."""
+
+        # Create mesh.
+        mesh = Mesh()
```
isteinbrecher, line 745: For such simple test cases you don't need to define the mesh, simply pass `input_file` to `create_beam_mesh_helix` (also for the other test cases)

tests/testing_mesh_creation_functions.py
```diff
@@ -737,6 +738,99 @@ def line(t):
...
+        create_beam_mesh_helix(
+            mesh,
+            Beam3rHerm2Line3,
+            mat,
+            [1.0, 1.0, 1.0],
+            [-1.0, -1.0, -1.0],
+            [1.0, 1.0, 1.0],
+            np.pi / 6,
+            height_helix=80.0,
+            n_el=10,
```
isteinbrecher, line 825: You should use fewer elements for tests, 3-4 elements for the helix should be sufficient (also for the previous test cases).

tests/testing_mesh_creation_functions.py
```diff
@@ -737,6 +738,99 @@ def line(t):
...
+            l_el=5.0,
+        )
+
+        # Add mesh.
+        input_file.add_mesh(mesh)
+
+        # Check the output.
+        compare_test_result(self, input_file.get_string(header=False))
+
+    def test_mesh_creation_functions_helix_radius_zero(self):
```
isteinbrecher, line 803: Should we add an additional test case for `angle=np.pi/2`?

tests/testing_mesh_creation_functions.py
```diff
@@ -737,6 +738,99 @@ def line(t):
...
+            mesh,
+            Beam3rHerm2Line3,
+            mat,
+            [0.0, 0.0, 1.0],
+            [0.0, 0.0, 0.0],
+            [2.0, 0.0, 0.0],
+            np.pi / 4,
+            height_helix=80.0,
+            l_el=5.0,
+        )
```
isteinbrecher, line 764: You get return values from `create_beam_mesh_helix` (node sets). To also test that they are correct, write the following (this will require an update of the reference files):
```python3
helix_set = create_beam_mesh_helix(
            input_file,
            Beam3rHerm2Line3,
            mat,
            [0.0, 0.0, 1.0],
            [0.0, 0.0, 0.0],
            [2.0, 0.0, 0.0],
            np.pi / 4,
            height_helix=80.0,
            l_el=5.0,
        )
input_file.add(helix_set)
```

meshpy/mesh_creation_functions/beam_basic_geometry.py
isteinbrecher: The way you initially create the helix, the helix axis will be in the z-direction and the starting point will lie on the x-axis. Thus we need to create a rotation that maps the z-axis onto the desired helix axis and the x-axis onto the `start_point_origin_vec`. No need to invert anything here.

meshpy/mesh_creation_functions/beam_basic_geometry.py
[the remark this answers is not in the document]
davidrudlstorfer: Thanks for the explanation regarding the triad orientation! Makes sense now

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: I am unsure why you would add the `sin` here? Probably you meant `np.sign`. Added in latest commit

isteinbrecher: Since this is only computed for `cos(twist_angle) == 0` you don't need the `sign` here, but I am fine if we add this for clarity.

tests/testing_mesh_creation_functions.py
davidrudlstorfer: I've reduced all setups to 3-4 elements

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: ooh yea you're totally right. removed the sign

tests/testing_mesh_creation_functions.py
davidrudlstorfer: Done

davidrudlstorfer:
Thanks for your review/feedback. I've added everything and sorry for the long discussions. Locally all tests succeed.

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: During further testing on own examples, the signum is actually needed. We also create a straight beam geometry for `radius == 0`, there `twist_angle != np.pi/2` and the geometry is artificially lengthened or shortened

I pushed the fix

isteinbrecher (approved):
Thanks again for the contribution @davidrudlstorfer everything looks good to me. If you want you can add yourself to the `CONTRIBUTORS.md` and the website.

isteinbrecher:
@davidrudlstorfer I just merged #65 which might require a reordering of the created sets in the reference file for this merge request.

davidrudlstorfer:
I've now done the following things which finalize the PR from my side:

- I've fixed the bugs regarding the twist angle. Now a twist angle between `-inf` and `inf` can be utilized. Additionally, the height and twist can either be positive or negative. Tested it manually for angles in all 4 quadrants as well as angles smaller than 0 and angles larger than `2pi`. This now enables all thinkable sorts of input resulting in correct geometries.

- Made the warning of returning a straight line optional and disabled it within the tests.

- Added myself to the contributors - thanks for the option that means a lot to me🙏🏻

- Finally, I've rebased on main and updated the reference files - again all tests pass locally and I hope that the tests now also pass within the pipeline. That's the output of running `python tests/testing_main.py`

```
/home/rudlstorfer/01_work/01_workspace/meshpy/tests/testing_utility.py:81: UserWarning: Path to baci-release not found. Did you set the environment variable BACI_RELEASE?
  warnings.warn(
...sssss.ss.ss.ss.ss.....................................s................................................ssmesh_1 number of nodes: 13
mesh_1 number of elements: 6
Point couplings before "couple_nodes": 0
Point couplings after "couple_nodes": 6
.
----------------------------------------------------------------------
Ran 109 tests in 2.527s

OK (skipped=16)
```

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -91,7 +92,10 @@ def create_beam_mesh_line(
     t1 = direction / line_length
 
     # Check if the z or y axis are larger projected onto the direction.
-    if abs(np.dot(t1, [0, 0, 1])) < abs(np.dot(t1, [0, 1, 0])):
+    # The tolerance is used here to ensure that round-off changes in the last digits of
+    # the floating point values don't switch the case. This increases the robustness in
+    # testing.
+    if abs(np.dot(t1, [0, 0, 1])) < abs(np.dot(t1, [0, 1, 0])) - mpy.eps_quaternion:
```
isteinbrecher, line 98: Some changes in the last digits of the vectors caused a switch in this `if` clause, thus the failing tests. Should now be working more robustly. 

@davidrudlstorfer I had to change the reference results, please check if the created mesh still looks as you expect it (especially the rotation of the triad in the cross section direction). Otherwise, we have to something different.

isteinbrecher (approved):

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: The changes and resulting geometry looks good! Thanks for digging deeper and helping out here 🙏🏻

isteinbrecher (approved):

PR #69, opened 2024-04-24
Add parallel projection and cosys in pyvista

davidrudlstorfer:
Add option for pyvista visualization to enable parallel projection and option to display small cosys widget.

Looks like this:

Screen Shot 2024-04-24 at 14 27 25 PM


meshpy/mesh.py
```diff
@@ -779,6 +779,8 @@ def display_pyvista(
         beam_radius_for_display=0.1,
         resolution=20,
         is_testing=False,
+        parallel_projection=False,
+        add_cosys=False,
```
isteinbrecher, line 783: I am fine with having this on by default, what do you think? @davidrudlstorfer

davidrudlstorfer: I am totally fine with that. Do you want to enable it always (remove flags) or keep the flags and just set them to true?

meshpy/mesh.py
```diff
@@ -808,6 +808,10 @@ def display_pyvista(
         """
 
         plotter = pv.Plotter()
+
+        plotter.enable_parallel_projection()
```
isteinbrecher, line 812: Sorry, there was a misunderstanding, the `enable_parallel_projection` should be optional (turned off by default). And the axes can be set to on all the time.

Issue #72, opened 2024-04-26
MeshPy with Potential Interactions


davidrudlstorfer:
Currently I am further investigating the usage of MeshPy for the creation of multiple helices (based on #60) within a single input file including beam potential interactions.

In a simplified setting I have multiple helices with different start points, different axes, different twist angles and differing heights which are combined into a single 4C input file. Further, I want to enable potential interactions between individual fibers which results in the necessity to define each helix as an individual line.

As a simple proof of concept I use the following Python Code. 

<details>
  <summary>Code</summary>
  

  ```python
import numpy as np

from meshpy import (
    Mesh,
    InputFile,
    MaterialReissner,
    Beam3rHerm2Line3,
)

from meshpy.mesh_creation_functions.beam_basic_geometry import create_beam_mesh_helix


def main():
    """Minimal example of multiple helical beams"""

    # define input file
    input_file = InputFile()

    # define material
    mat = MaterialReissner(youngs_modulus=1000, radius=0.5, shear_correction=1.0)

    # define radii of differing helical segments
    radii = np.array([0, 3, 6, 9, 12, 15, 18])

    # determine twist angles if helical segments
    outermost_twist_angle = np.deg2rad(45)
    twist_angles = []
    for radius in radii:
        if np.isclose(radius, 0.0):
            twist_angles.append(np.pi / 2)
            continue
        twist_angles.append(
            np.arctan((radii[-1] / radius) * np.tan(outermost_twist_angle))
        )

    # create helices
    for radius, twist_angle in zip(radii, twist_angles):

        tmp_mesh = Mesh()

        helix_set = create_beam_mesh_helix(
            tmp_mesh,
            Beam3rHerm2Line3,
            mat,
            [0, 0, 1],  # axis
            [0, 0, 0],  # axis_point
            [radius, 0, radius],  # start_point
            twist_angle,  # twist_angle
            height_helix=2 * np.pi * 10,
            n_el=20,
            warning_straight_line=False,
        )

        input_file.add_mesh(tmp_mesh)
        input_file.add(helix_set["line"])

    input_file.write_input_file("_test_output/test.dat", header=False)
    input_file.write_vtk("test", "_test_output/")

    input_file.display_pyvista(
        beam_nodes=True,
        beam_tube=True,
        beam_cross_section_directors=True,
        resolution=50,
        parallel_projection=True,
    )


if __name__ == "__main__":

    main()

  ```
</details>

[...]

PR #73, opened 2024-05-03
Fix mesh creation for helical geometries

davidrudlstorfer:
See #72

Add a temporary mesh to the helical mesh creation function to simplify the mesh creation procedure


isteinbrecher:
Looks good to me. Can you also add this to the testing? Should be sufficient if you simply add a line to the mesh in all the test cases before `create_helix` function is called.

davidrudlstorfer:
Thanks for your review, I've added a simple line geometry to the four corresponding helix test cases. I've also checked the resulting reference files and .vtk output to verify the correctness.

PR #85, opened 2024-06-18
Add curved beams in vtk

isteinbrecher:
Add the option to visualise beams with more than a linear interpolation. Concrete, Hermite shape functions are used to visualise a "smooth" beam. In general this does NOT match the exact interpolation used in FE solvers for higher oder beam elements.


mayrmt:
@isteinbrecher Would it be possible to use the exact FE interpolation also for visualization?

davidrudlstorfer:
Just out of curiosity and to better understand the entire beam framework in 4C and meshpy. As far as I understand you now added the option to visualize the beam geometry with more than one segment within the `.vtu` output.

- This now basically resembles the procedure for the visualization within 4C?
- This entire procedure is done because ParaView does not support Hermitian shape functions?

Thanks in advance!

isteinbrecher:
@mayrmt Possible yes, but in my opinion not feasible, as one would end up implementing the beam formulation from the solver in MeshPy just to get a better visualisation.

@davidrudlstorfer This pull request simply introduces the option to get a better visual approximation of the geometry in MeshPy. Hermite shape functions are used for convenience, not because this now matches the interpolation in 4C.

Take for example a simple half circle modelled with 2 beam elements. The vtk visualisation in MeshPy used to always look like this (now with `beam_centerline_visualization_segments=1`)
Screenshot 2024-06-19 at 10 12 45
Now it can look like this (with `beam_centerline_visualization_segments=5`)
Screenshot 2024-06-19 at 10 12 31
In the near future, I envision to set the default value of `beam_centerline_visualization_segments` to 5.

With newer VTK versions ParaView supports Hermite interpolations, but this comes with a lot of other drawbacks. From my experience the "piecewise poly line" visualisation is the best solution to display 1D beam elements in ParaView.

PR #91, opened 2024-07-23
Create clean up function for simulation manager

knarfnitram:
Add a simple clean up function for the simulation manager, which simply deletes all folders and files within the main folder (Place where the simulation is performed/written) - Has not been tested so far for the cluster.

### DIFF
--- a/meshpy/utility.py
+++ b/meshpy/utility.py
@@ -143,21 +143,22 @@ def get_min_max_coordinates(nodes):
     return min_max
 
 
-def clean_simulation_directory(sim_dir, ask_before_clean=False):
+def clean_simulation_directory(sim_dir, *, ask_before_clean=False):
     """
-    If the simulation directory exists, the user is asked if the contents
-    should be removed. If it does not exist, it is created.
+    Clear the simulation directory. If it does not exist, it is created.
+    Optionally the user can be asked before a deletion of files.
     Args
     ----
     sim_dir:
-        Path to a directory .
+        Path to a directory
     ask_before_clean: bool
-        flag which indicates whether the user must confirm the removal.
+        Flag which indicates whether the user must confirm removal of files and directories
     """
 
     # Check if simulation directory exists.
     if os.path.exists(sim_dir):
-        print(f'Path "{sim_dir}" already exists')
+        if not ask_before_clean:
+            print(f'Path "{sim_dir}" already exists')
         while True:
             if not ask_before_clean:
                 answer = input("DELETE all contents? (y/n): ")
--- a/meshpy/utility.py
+++ b/meshpy/utility.py
@@ -143,17 +143,26 @@ def get_min_max_coordinates(nodes):
     return min_max
 
 
-def clean_simulation_directory(sim_dir):
+def clean_simulation_directory(sim_dir, ask_before_clean=False):
     """
     If the simulation directory exists, the user is asked if the contents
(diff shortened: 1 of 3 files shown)


knarfnitram:
@isteinbrecher any ideas why the performacne tests are failing?

isteinbrecher:
@knarfnitram there already exists a `clean_simulation_directory` function in `utility` that basically contains the same code (looks like we found the same stackexhange post). Additionally this existing function asks the user if the directory should be cleaned.

I think the best way to go forward is to add an optional parameter to the existing function `ask_before_clean` to turn on/off if the user should be asked. Also maybe `ensure_clean_simulation_directory` would be a better name. What do you think?

Regarding the performance tests, depending on the overall load on the test machine they might not pass (does not hinder the progress of the pull request). I had a look and everything looks fine.

knarfnitram:
Makes totally sense to reuse the clean functionality, especially if we were inspired by the same stack exchange post. 

Yes, the name sounds better. 
I have renamed and updated the simulation manager and also added the option `ask_before_clean`.

Regarding the performance checks, I thought that this might be the reason, but thanks for clearifying that!

tests/testing_simulation_manager.py
```diff
@@ -256,6 +256,24 @@ def test_simulation_manager_batch_file(self):
             additional_identifier=2,
         )
 
+    def test_simulation_manager_cleaned_start_directory(self):
```
isteinbrecher, line 259: Please add a doc string shortly describing what this test tests.

tests/testing_simulation_manager.py
```diff
@@ -256,6 +256,24 @@ def test_simulation_manager_batch_file(self):
...
+        manager = self.create_simulations(convergence_base_dir)
+
+        # Ensure clean state before adding next simulation
+        manager.ensure_clean_simulation_directory()
+
+        # check if the folder still exists
+        self.assertTrue(os.path.isdir(manager.path))
+
+        # check if folder is empty
+        self.assertFalse(len(os.listdir(manager.path)))
```
isteinbrecher, line 275:
```suggestion
        self.assertFalse(len(os.listdir(manager.path))==0)
```

meshpy/utility.py
```diff
@@ -143,17 +143,27 @@ def get_min_max_coordinates(nodes):
...
+    ----
+    sim_dir:
+        Path to a directory
+    ask_before_clean: bool
+        Flag which indicates whether the user must confirm removal of files and directories
     """
 
     # Check if simulation directory exists.
     if os.path.exists(sim_dir):
-        print(f'Path "{sim_dir}" already exists')
```
isteinbrecher, line 154: Shouldn’t this be the other way? If ask_before_clean is True we want to show the message and ask the user. Now it is the other way around.

isteinbrecher:
@knarfnitram what is the state of this pull-request?

meshpy/utility.py
knarfnitram: yes, you are right, i have inverted the flag.

isteinbrecher (approved):

isteinbrecher:
@knarfnitram looks good to me, feel free to merge

knarfnitram:
So unfortunally it says:
You’re not [authorized](https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches) to merge this pull request.

PR #93, opened 2024-07-24
Improve helix mesh creation

### LINKED ISSUE #92: Improve helix mesh creation parameters
A the moment the helix mesh creation function always takes the `twist_angle` and then requires either the `height` or the number of `turns`. In general two of these three parameters are needed to fully define the wanted helix (wile some corner cases might require a specific combination). The API to the function should represent this to ease the user the creation of a helix in whichever variant is best suited for the problem at hand.

In this discussion the "missing" parameter conversion is explained: https://github.com/imcs-compsim/meshpy/issues/87#issuecomment-2242452732

davidrudlstorfer:
Closes #92 

- Renames `twist_angle` in helix mesh creation function to `helix_angle` and synonyms are added in the docstring
- Introduces the option to create a helix with either combination of [helix angle, height, turns]
- Extends tests (utilizes the same reference file to check that all three combinations of input parameters are valid


davidrudlstorfer:
@isteinbrecher would be ready for review, unfortunately it took longer than expected with all edge cases 😅

meshpy/mesh_creation_functions/beam_basic_geometry.py
```diff
@@ -516,14 +516,18 @@ def create_beam_mesh_helix(
...
+        angle). Exactly two arguments of [helix_angle, height_helix, turns]
+        must be provided.
     height_helix: float
-        Height of helix. Mutually exclusive with number of turns.
+        Height of helix. Exactly two arguments of [helix_angle, height_helix,
+        turns] must be provided.
     turns: float
-        Number of turns. Mutually exclusive with height of helix.
+        Number of turns. Exactly two arguments of [helix_angle, height_helix,
+        turns] must be provided.
```
isteinbrecher, line 528: I would just put `Exactly two arguments of [helix_angle, height_helix, turns] must be provided.` in the main description and not in each argument.

isteinbrecher (commented):
Thanks a lot @davidrudlstorfer for this change, and the added tests. I think this really helps to further improve the usability of this function.

meshpy/mesh_creation_functions/beam_basic_geometry.py
davidrudlstorfer: Thanks for the review and good idea! I've just pushed changes to improve the docstring and ease of the helix function.

isteinbrecher (approved):

PR #96, opened 2024-09-04
Locsys condtion for beams

knarfnitram:
Adds the `LOCSYS` condition to meshpy. 

I have created only a test case for a `POINT LOCSYS` with beams. 
@isteinbrecher  Do we need additional tests for: 

- line 
- surface 
- volume ?


meshpy/boundary_condition.py
```diff
@@ -80,6 +80,7 @@ def from_dat(cls, bc_key, line, **kwargs):
 
         if bc_key in (
             mpy.bc.dirichlet,
+            mpy.bc.locsys,
```
isteinbrecher, line 83: Ordering should match the one in `meshpy/conf.py`. I would put it at the end.

tests/testing_meshpy.py
```diff
@@ -1943,6 +1943,55 @@ def test_meshpy_display_pyvista(self):
...
+                bc_type=mpy.bc.dirichlet,
+            )
+        )
+
+        # Add local sys condition.
+        input_file.add(
+            BoundaryCondition(
+                beam_set["end"],
+                "ROTANGLE 0.0 0.0 0.1 FUNCT 0 0 0 USEUPDATEDNODEPOS 0 USECONSISTENTNODENORMAL 0",
+                bc_type=mpy.bc.locsys,
```
isteinbrecher, line 1988: Since we have quite an elaborate large rotational implementation within meshpy, I suggest a different approach here:
Lets create a class `LocSysCondition` (preferably in the `four_c` subdirectory) that takes a `Rotation` as input argument.

Should look something like this:
```python
class LocSysCondition(BoundaryCondition):

    def __init__(
        self,
        geometry,
        rotation,
        *,
        **kwargs,
    ):
        super().__init__(geometry,
             "ROTANGLE {} {} {} FUNCT 0 0 0 USEUPDATEDNODEPOS 0 USECONSISTENTNODENORMAL 0".format(*rotation.get_rotation_vector()),
             bc_type=mpy.bc.locsys, **kwargs)
```

meshpy/boundary_condition.py
knarfnitram: Thanks, for catching that.

tests/testing_meshpy.py
```diff
@@ -1943,6 +1944,49 @@ def test_meshpy_display_pyvista(self):
         mesh = self.create_beam_to_solid_conditions_model()
         _plotter = mesh.display_pyvista(is_testing=True, resolution=3)
 
+    def test_meshpy_locsys_condition(self):
```
isteinbrecher, line 1947: Docstrings should start directly after the three quotations:
```
        """Test case for point locsys condition for beams.
        The testcase is similar to beam3r_herm2line3_static_locsys.dat, but with simpler material.
        """
```

tests/testing_meshpy.py
```diff
@@ -1943,6 +1944,49 @@ def test_meshpy_display_pyvista(self):
...
+            )
+        )
+
+        input_file.add(
+            BoundaryCondition(
+                beam_set["end"],
+                "NUMDOF 9 ONOFF 1 0 0 0 0 0 0 0 0 VAL 1.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 FUNCT 1 0 0 0 0 0 0 0 0",
+                bc_type=mpy.bc.dirichlet,
+            )
+        )
```
isteinbrecher, line 1982: Why do we need these here?

knarfnitram: I recreated the test `beam3r_herm2line3_static_locsys` just with a simpler material, therefore I added the additional bc on the same points (to see if this is a problem, but it isn't).
One can delete it, but then I would also remove the reference at the top to the `beam3r_herm2line3_static_locsys`.

isteinbrecher: Then lets leave this and add a comment stating that the other BC are there to check that the combination works

isteinbrecher:
@knarfnitramwhat is the status of this merge request? Is something still missing?

knarfnitram:
I have updated the branch, so I think it's ready to be merged.

isteinbrecher (approved):

meshpy/four_c/locsys_condition.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+
+# MeshPy modules
+from ..conf import mpy
+from ..boundary_condition import BoundaryCondition
+
+
+class LocSysCondition(BoundaryCondition):
+    """
```
isteinbrecher, line 42: „This object represents a locsys condition in 4C. It allows to rotate the local coordinate system used to apply Dirichlet boundary conditions“

tests/testing_four_c.py
```diff
@@ -175,6 +179,48 @@ def create_model():
...
+            )
+        )
+        # Add additional dirichlet boundary condtion to check if combination with locsys condition works.
+        input_file.add(
+            BoundaryCondition(
+                beam_set["end"],
+                "NUMDOF 9 ONOFF 1 0 0 0 0 0 0 0 0 VAL 1.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 FUNCT 1 0 0 0 0 0 0 0 0",
+                bc_type=mpy.bc.dirichlet,
+            )
+        )
```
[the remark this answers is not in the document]
isteinbrecher: Ignore this commen.

meshpy/four_c/locsys_condition.py
isteinbrecher: Line formatting seems off here. I think once this is fixed we can merge.

knarfnitram: Thanks, I thought that black would have fixed this.

isteinbrecher: `black` does not change doc strings.

isteinbrecher (approved):

isteinbrecher:
@knarfnitram you should also be able to merge in this repository now.

Issue #76, opened 2024-05-08
Define the python version


eulovi:
While trying to run Meshpy, I noticed that Meshpy is now using Python 3.12. In the Installation instructions, it is unclear if this Python version is used, since with `python3 -m venv meshpy-env` a python environment is created with the local Python3 version, which can be different from Python 3.12.   
Should the installation instructions be updated?

davidrudlstorfer:
It would be nice to just switch to a conda environment and maybe update the install instructions accordingly?

Something like

`conda create -n meshpy python=3.12`

`conda activate meshpy`

`pip install -e .`

isteinbrecher:
Thanks for the suggestion. We should definitely adapt the script so it works with the `main` version of `meshpy`.
@davidrudlstorfer I am against an installation instruction completely relying on `conda`. But maybe something like:


## Installation

MeshPy is developed with `python3.12`. Other versions of Python might lead to issues. It is recommended to use a python environment container such as `venv` or `conda`.
- `venv`: Chose an appropriate directory for this, e.g., `/home/user/opt`. A virtual environment can be created with (make sure that python3 points to a python3.12
  ```bash
  python3 -m venv meshpy-env
  ```
  The created virtual environment can be loaded with
  ```bash
  source meshpy-env/bin/activate
  ```
- `conda`: A condo environment can be created with
  ```
  conda create -n meshpy python=3.12
  conda activate meshpy
  ```

From now on we assume that the previously created enviroment is loaded. To install `meshpy` go to the repository root directory
```bash
cd <path_to_meshpy>
```
...

isteinbrecher:
Closed with #109

PR #105, opened 2024-10-02
Black for pre-commit hook

### LINKED ISSUE #104: Add automatic code formatting to (pre) commit
For most of the software projects I am currently working with, automatic code formatting on committing is standard.
Of course, one can shift this duty to the IDE(click on save, plugins etc.), but since the CI is already checking whether the code is formatted with `black`.  I think it would be reasonable to add it in a pre commit as well. 

Are there any objections to add code formatting on (pre) commit ?

knarfnitram:
Sets `black` as default style guide via the `pre-commit` option before contributing to `MeshPy`.

closes #104 .


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 23.9.1  # Use the latest release of black
+    hooks:
+      - id: black
+        language_version: python3  # only choose python 3
```
knarfnitram, line 6: Do we need to further specify the python version, or is python3 enough?

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dependencies = [
   "pyvista_utils@git+https://github.com/isteinbrecher/pyvista_utils.git@86210ab",
   "scipy",
   "splinepy",
-  "vtk"
```
isteinbrecher, line 42: The dependencies here are in alphabetical order.

.pre-commit-config.yaml
knarfnitram: Yeah, so good idea @davidrudlstorfer, I think we can add (or forbid) some more functionalities. However, I would try not overcomplicate things too much, but your repo provides a good start. 
@isteinbrecher any additional features you like ?

isteinbrecher: I am fine with having `black` in the default settings (especially the increased number of characters per line).

@davidrudlstorfer is the `python3.12` actually specify a python version that is used to run the pre-commit hook or is this sort of an option for black to know how to format the code?

Thanks for the link @davidrudlstorfer! I don't know most of the tools used there, but I am sure some could be very useful for MeshPy. I will open up an issue to gather ideas on possible advances in that direction.

isteinbrecher (commented):
Thanks @knarfnitram for adding this, this is for sure very useful for development.

.pre-commit-config.yaml
davidrudlstorfer: @isteinbrecher the `language_version: python 3.xx` should coincide with the installed/used python version. 

Therefore, it would be beneficial to initially solve #76 and once and for all define a specific Python version for MeshPy before adapting any pre-commit hooks

isteinbrecher: I opened a pull request to resolve #76. Then we can put python3.12 here.

isteinbrecher: #76 is solved, so I think we can proceed here.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 24.4.2
+    hooks:
+      - id: black
+        language_version: python3.12
+        args: ["--line-length=79"]
```
isteinbrecher, line 7: This will cause issues since we test the code formatting with the default `black` settings, which means a line length of 88.

knarfnitram: I see, I have not tested any large lines, I changed it to default settings with no arguments.

.gitignore
```diff
@@ -41,3 +41,4 @@ website/public/
 
 # Don't ignore directory of Github workflows
 !.github
+!.pre-commit-config.yaml
```
davidrudlstorfer, line 44: Why do you add the pre-commit config.yaml to the gitignore file? AFAIK this file should be tracked?

davidrudlstorfer: Sorry for my misunderstanding, didn't catch that earlier on😅 Thanks for the explanation!

knarfnitram: No problem - critically questioning my contributions is always a good thing :wink:

isteinbrecher: FYI, even if you exclude all hidden files with `.*`, you can add them to the repo with `git add -f .pre-commit-config.yaml` then it will be tracked even tough it is excluded in `.gitignore`, that's also how the `.gitignore` file is tracked.

PR #109, opened 2024-10-02
Add to install instructions

isteinbrecher:


davidrudlstorfer:
Looks good 👍

Do the sections regarding the cython code need an update as well? I actually never used cython but probably the `python3` commands are not up to date when using a conda environment without any `python3` presence?

isteinbrecher:
Are there conda environments without defining `python3`? For me it works per default (on macOS).

davidrudlstorfer:
Yea as far as I know `python3` is installed in macOS by default. Could the remaining commands just use `python` within the conda or virtual env?

isteinbrecher:
I had a look, I am using `conda` on macOS and linux and `python3` always points to the correct path. I would keep it for now, if this is an issue in the future, we can adapt the `README`.

PR #115, opened 2024-10-15
Draft: Add  FD generation with interwoven option

knarfnitram:
To generate the flow diverter, I have added a simple function to back up how this can done. 

Additionally, I started drafting _interwoven_ feature. This allows to move the intersection wires just slightly above such that the wires do not intersect anymore. 
In the following picture you can clearly see it, the first node is inside and at the next intersection it has moved outside. 

The question is, can we generalize this algorithm a little bit more?

Also, for high element numbers it's not yet quite optimal, since it adjusts only the intersecting elements, but not the position of the surrounding neighbors. Therefore, the curvature is not optimal anymore. 

Screenshot from 2024-10-15 23-17-01


davidrudlstorfer:
Since the specific "flow diverter" geometry is added to mehspy, in my opinion it would make sense to briefly describe what a flow diverter actually is or for what this geometry is intended? As a person who has no additional knowledge on this type of geometry it would be very helpful (also regarding the abbreviation `fd`)

The interwoven feature sounds very interesting!

meshpy/mesh.py
```diff
@@ -632,6 +642,100 @@ def couple_nodes(
             for node_list in partner_nodes:
                 self.add(coupling_factory(node_list, coupling_type, coupling_dof_type))
 
+    def interwove_nodes_of_z_cylinder(self, *, nodes=None, fibers, beam_radius=None):
```
isteinbrecher, line 645: This is a highly problem specific function and does not belong as a member of `Mesh`.

meshpy/mesh_creation_functions/beam_fd.py
```diff
@@ -0,0 +1,80 @@
...
+# SOFTWARE.
+# -----------------------------------------------------------------------------
+"""
+This file has functions to creates a flow diverter
+"""
+from meshpy.mesh_creation_functions import create_beam_mesh_helix
+import numpy as np
+
+
+def create_beam_flow_diverter(
```
isteinbrecher, line 38: If we add additional geometry we have to make sure to document this.

isteinbrecher (commented):
Thanks @knarfnitram for the contribution.

Two general remarks:

1. If is difficult to draw a sharp line on whether or not to include geometry into MeshPy itself or leave it in the calling scripts. Currently the FD geometry is simply two nested loops that create a helix. I personally would not check that into MeshPy at the moment. However, since this will likely become more complex in the future, I am fine with adding this also now.

2. I assume this `interwove` functionality is to create some sort of "webbing" between the FD wires. What you are doing now is to modify the reference configuration so you can initialise beam-to-beam contact, correct? I think that is a bad idea, as even very small modifications of the reference geometry of a thin structure will change that structures overall behaviour in a significant way. In my opinion you have two options here: i. move the nodes apart via a non-homogeneous DBC or ii. apply loads that will separate the beams (as is done in Chapter 4.5.3.3 in Christoph Meier's dissertation). Then you can add beam-to-beam contact and remove the initial BC.

isteinbrecher:
Closed as this branch will be removed from this repo, will live on in the fork of @knarfnitram

PR #117, opened 2024-10-30
Improve time step setting options in header functions

isteinbrecher:


meshpy/header_functions.py
```diff
@@ -376,6 +377,8 @@ def set_header_static(
         Time increment per step.
     n_steps: int
         Number of time steps.
+    total_time: float
+        Total time of simulation
```
davidrudlstorfer, line 381: Here the optional requirement of only two parameters could be added to the docstring?

meshpy/header_functions.py
```diff
@@ -425,6 +428,19 @@ def set_header_static(
...
 
+    given_time_arguments = [arg for arg in (time_step,n_steps,total_time) if arg is not None]
+    if len(given_time_arguments) < 2:
+        raise ValueError('At least two of the following arguments "time_step", "n_steps" or "total_time" are required')
+
+
+    if time_step is None:
+        time_step = total_time/n_steps
+    elif n_steps is None:
+        n_steps = total_time/time_step
```
davidrudlstorfer, line 439: Should this be casted to an int?

```suggestion
        n_steps = int(total_time/time_step)
```

isteinbrecher: Good catch, I think `round` makes the most sense here.

davidrudlstorfer (approved):

PR #122, opened 2024-11-11
Update pyvista_utils version

isteinbrecher:
Update the `pyvista_utils` version.

@all if you are running the MeshPy tests locally you might have to reinstall MeshPy inside `pip` for the testing to work.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
   "numpy<2",
   "pre-commit",
   "pyvista",
-  "pyvista_utils@git+https://github.com/isteinbrecher/pyvista_utils.git@86210ab",
+  "pyvista_utils@git+https://github.com/isteinbrecher/pyvista_utils.git@3aeb8329bd28b55c01977a10c8eae3a9b2f67442",
   "scipy",
   "splinepy",
   "vtk"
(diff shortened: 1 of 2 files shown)


davidrudlstorfer:
Unfortunately I am not able to get the tests running even after reinstalling MeshpPy inside pip (`pip install -e .`)

`ModuleNotFoundError: No module named 'vtk_utils'`

Are there any additional steps you executed?

davidrudlstorfer:
When i exchange the import from `vtk_utils.compare_grids` to `pyvista_utils` it works as expected

isteinbrecher:
I am not sure why `pip install -e .` does not work. Can you check the state of `pyvista_utils` in your local environment?

The code is correct (at least I am pretty sure). With the latest changes, the repository _pyvista_utils_ comes with two modules `pyvista_utils` and `vtk_utils`. I am pretty sure this is a bad design choice, but I wanted to emphasise the parts of `pyvista_utils` that actually depend on `pyvista` and the ones that only depend on `vtk`. Probably the best solution would be to rename `pyvista_utils` to `visualisation_utils` or something like that. But still that does not explain why the pip reinstall does not work for you.

davidrudlstorfer:
Sorry for the confusion was a problem on my side - problem solved -> #125

PR #126, opened 2024-11-13
Interpolate dbc monitor input

knarfnitram:
This MR adds the functionality to extract all nodal force values of a previous simulation and adds them to a new simulation as Neuman condition. With the previous functionality, only the last values of the force were used.

_Note:_
This MR has been split into #127 and #128 to reduce the complexity.


meshpy/four_c/dbc_monitor.py
```diff
@@ -91,7 +93,7 @@ def read_dbc_monitor_file(file_path):
 def dbc_monitor_to_input(input_file, file_path, step=-1, function=1, n_dof=3):
     """
     Convert the Dirichlet boundary condition monitor log to a Neumann
-    boundary condition input section.
+    boundary condition input section. Uses only the last force value.
```
knarfnitram, line 96: Should I remove this function?

meshpy/geometry_set.py
```diff
@@ -218,8 +218,10 @@ def get_all_nodes(self):
         elif self.geometry_type is mpy.geo.line:
             nodes = []
             for element in self.geometry_objects[mpy.geo.line]:
-                nodes.extend(element.nodes)
-            return list(set(nodes))
+                for node in element.nodes:
+                    if node not in nodes:
+                        nodes.append(node)
+            return nodes
         else:
```
knarfnitram, line 225: Can we speed up these lines here?

tests/testing_meshpy.py
```diff
@@ -82,6 +82,7 @@
 )
 from meshpy.mesh_creation_functions.beam_honeycomb import create_beam_mesh_honeycomb
 from meshpy.mesh_creation_functions.beam_curve import create_beam_mesh_curve
+from meshpy.utility import check_node_by_coordinate
```
isteinbrecher, line 85: Why do we need this additional import?

tests/testing_meshpy.py
```diff
@@ -119,14 +120,15 @@ def create_test_mesh(mesh):
...
-    create_beam_mesh_line(
+    beamset = create_beam_mesh_line(
         mesh,
         Beam3rHerm2Line3,
         material,
         [100 * random.uniform(-1, 1) for _i in range(3)],
         [100 * random.uniform(-1, 1) for _i in range(3)],
         n_el=3,
     )
+    return beamset
```
isteinbrecher, line 131: Why do we need to change the signature of this function?

tests/testing_meshpy.py
```diff
@@ -541,6 +543,41 @@ def test_meshpy_geometry_sets(self):
         # Check the output.
         compare_test_result(self, mesh.get_string(header=False))
 
+    def test_unique_ordering_of_line_condition(self):
+        """
+        This test case ensures that the ordering of nodes defined within the line condition of the input file is unique.
```
isteinbrecher, line 548: This test ensures that ordering of the nodes returned from the `get_all_nodes` function for line sets is unique.

tests/testing_meshpy.py
```diff
@@ -541,6 +543,41 @@ def test_meshpy_geometry_sets(self):
...
+        """
+
+        # set up a beam mesh with material
+        input_file = InputFile()
+        mat = MaterialReissner(youngs_modulus=1.0, radius=1)
```
isteinbrecher, line 553: You can leave the values empty here.

tests/testing_meshpy.py
```diff
@@ -541,6 +543,41 @@ def test_meshpy_geometry_sets(self):
...
+        beam_set = create_beam_mesh_line(
+            input_file, Beam3rHerm2Line3, mat, [0, 0, 0], [2, 0, 0], n_el=10
+        )
+
+        # apply different dirichlet to all nodes within this condition
+        for i, node in enumerate(beam_set["line"].get_all_nodes()):
+
+            # do not constraint middle nodes
```
isteinbrecher, line 561: We don’t need to filter the middle nodes for this test case.

tests/testing_meshpy.py
```diff
@@ -541,6 +543,41 @@ def test_meshpy_geometry_sets(self):
...
+            if not node.is_middle_node:
+                # add different condition value for each node
+                input_file.add(
+                    BoundaryCondition(
+                        GeometrySet(node),
```
isteinbrecher, line 566: We can use a “dummy” string here for testing purposes.

tests/testing_utility.py
```diff
@@ -34,9 +34,11 @@
 
 # Python imports
 import unittest
+import numpy as np
+import numpy.testing as npt
```
isteinbrecher, line 38: What do we use from `numpy.testing`?

tests/testing_utility.py
```diff
@@ -75,6 +77,101 @@ def test_is_node_on_plane(self):
             )
         )
 
+    def test_linear_time_transformation_scaling(self):
+        """
+        test the scaling of the intervall the function
```
isteinbrecher, line 82: General remark, do strings in MeshPy start with an upper case letter and directly after the tripple quotes (no newline).

tests/testing_four_c_simulation.py
```diff
@@ -109,6 +115,29 @@ def run_four_c_test(self, name, mesh, n_proc=2, restart=[None, None], **kwargs):
         )
         self.assertEqual(0, return_code, msg="Test {} failed!".format(name))
 
+    def create_cantilver_model(self, n_steps, time_step=0.5):
```
isteinbrecher, line 118: Can we move this out of the unit test class as a free function?

meshpy/geometry_set.py
isteinbrecher: I have an idea how to make this a little bit more efficient, I will push these changes in the upcoming days.

meshpy/four_c/dbc_monitor.py
isteinbrecher: Yes, I think we should aim at having a “full” function that can also handle this case.

meshpy/utility.py
```diff
@@ -387,3 +387,63 @@ def get_env_variable(name, *, default="default_not_set"):
     elif default == "default_not_set":
         raise ValueError(f"Environment variable {name} is not set")
     return default
+
+
+def linear_time_transformation(
```
isteinbrecher, line 392: Does this really belong into the general utility file? I think this is so closely related to the DBC monitor that we should add it as a free function there.

isteinbrecher (commented):
Thanks for the contribution @knarfnitram . I left some first comments.

.github/workflows/testing.yml
```diff
@@ -20,6 +20,8 @@ env:
   # Some tests require the 4C executable to check if the created input file
   # works with 4C.
   MESHPY_FOUR_C_EXE: /home_local/github-runner/testing_lib/four_c_master/release/4C
+  # Specify the testing directory
```
isteinbrecher, line 23: Let's put this change to the testing workflow into an own pull request.

Can we avoid adding the new parameter `MESHPY_TESTING_DIR`? As far as I know, we install meshpy into `${{ github.workspace }}`. Then we should be able to specify the path to the testing directory as `${{ github.workspace }}/tests/testing-tmp/`.

.github/workflows/testing.yml
```diff
@@ -72,6 +75,12 @@ jobs:
           # If we did not exit earlier, raise an error here
           exit 1
           ls
+      - name: Upload Results
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{github.job}}-${{ github.run_number }}
```
isteinbrecher, line 82: Is there a reason that once there is a space after the `{{` and once there is not?

.github/workflows/testing.yml
knarfnitram: Yes, I wanted to do this initially and just test it here.

.github/workflows/testing.yml
knarfnitram: seems not to be the case, i'll fix it.

tests/testing_meshpy.py
knarfnitram: I used to coordinate in x direction, so that one can easily track which points are changing.

tests/testing_meshpy.py
knarfnitram: Because I forgot to delete it - is removed.

tests/testing_meshpy.py
knarfnitram: we do not need it.

tests/testing_four_c_simulation.py
knarfnitram: Yes, I would actually like to move it into a complete new file named something as `mesh_generations_for_testing.py`. I think it would be good to store some "default meshes" for testing in one file.

knarfnitram:
Pipeline or local testing is still failing due to wrong ordering of nodes within `DNODE-NODE TOPOLOGY` , so it seems that #128 does not cover this example?
Values left are the one from the test machine. Value's at right are chosen from my local machine: 
Screenshot from 2024-11-18 18-23-36
By adapting the variables within commit [bcbb8f0](https://github.com/imcs-compsim/meshpy/pull/126/commits/bcbb8f031d39c68b5789ae1c52ed45b3467cc79e), the pipelines are passing. 
However, I am not happy with the overall design of the test, since it seems to be machine dependant.

meshpy/utility.py
knarfnitram: I moved it to `function_utility`, I think this is a better place.

knarfnitram:
@isteinbrecher Thanks for the hint with `os.walk` now everything is working as supposed.
Again Ready for review.

isteinbrecher (commented):
Sorry @knarfnitram for the long wait. I think I now understand what you want to do with this function and I agree with you initial assessment that a general interface might be to convoluted here. Maybe we can do the following 
- Add a "base" method with an interface like this:
  ```
  def create_point_neuman_condition(mesh, nodeID , list(3xtime_functions), list(3xforce values), n_dof)
  ```
- The "old" function then simply calls this like:
  ```
  create_point_neuman_condition(mesh, nodes[0] , [function, function, function], force[step])
  ```
- The "new" functionality then would create the needed `function_x`, `function_y`, `function_z` functions (they now also include the force values) and call this like:
  ```
  create_point_neuman_condition(mesh, nodes[0] , [function_x, function_y, function_z], [1,1,1])
  ```

Does this make sense? We can talk about this.

Please have a look at my comments and apply the ones that are obvious, the ones that don't apply to this proposed solution can simply be ignored.

Also should we add a check that there is exactly one node? Because I think other cases don't make sense.

knarfnitram:
I have added the functionalities, only regarding the testcases for  `add_point_neuman_condition_to_input_file` I am a little bit uncertain @isteinbrecher .

If I understood you correctly, you would like to offer the possibility to use the 3 key functionalities of `all_dbc_monitor_values_to_input` by yourself, namley: 

1) Read forces and switch sign 
`nodes, time, force, _ = read_dbc_monitor_file(file_path)` 
`force *= -1.0`

2) create custom functions (since this should be flexible to problem, without using the time transformations or hat functions) 

3) add_point_neuman_condition_to_input_file(myfile, nodes, ...)

Should I add another testcase to cover these steps explicitly or is it alright to test them "under the hood" , namely within the testcases of `all_dbc_monitor_values_to_input` ?

meshpy/four_c/dbc_monitor.py
```diff
@@ -36,8 +36,73 @@
 # Python modules.
 import numpy as np
 
+import meshpy
```
isteinbrecher, line 39: This looks weird. Can't you use `from meshpy import ...`?

meshpy/four_c/dbc_monitor.py
```diff
@@ -88,44 +153,203 @@ def read_dbc_monitor_file(file_path):
...
+        values to scale the function array with
     n_dof: int
         Number of DOFs per node.
     """
 
-    nodes, _, force, _ = read_dbc_monitor_file(file_path)
+    # Add the function to the input file, if they are not previously added.
+    for function in function_array:
+        if function not in input_file.functions:
+            input_file.add(function)
```
isteinbrecher, line 185:
```suggestion
    input_file.add(function_array)
```
The check is performed under the hood in `Mesh.add`.

meshpy/four_c/dbc_monitor.py
```diff
@@ -88,44 +153,203 @@ def read_dbc_monitor_file(file_path):
...
-    # The forces are the negative reactions at the Dirichlet boundaries.
-    force *= -1.0
+    # check if the dimensions of force and functions match
+    if force.size != 3:
+        raise ValueError(
+            f"The forces vector must have dimensions [3x1] not [{force.size}x1]"
+        )
+
+    if len(function_array) != 3:
+        raise ValueError(f"The function array must have length 3 not {force.size}.")
```
isteinbrecher, line 194: Lets move these checks to the beginning of the function.

meshpy/four_c/dbc_monitor.py
```diff
@@ -88,44 +153,203 @@ def read_dbc_monitor_file(file_path):
...
+            "{edz} FUNCT {{}} {{}} {{}}{edz}"
+        ).format(n_dof=n_dof, data=force, edz=extra_dof_zero),
         bc_type=mpy.bc.neumann,
-        format_replacement=[function],
+        format_replacement=function_array,
     )
     input_file.add(bc)
+
+
+def all_dbc_monitor_values_to_input(
```
isteinbrecher, line 214: Maybe my last comment was not clear. I meant to keep the "original" `dbc_monitor_to_input` here, since it requires a much simpler interface compared to the more general version `all_dbc_monitor_values_to_input` (like you initially mentioned and I didn't think though 😬). So here we would simply have:
```python
def dbc_monitor_to_input(input_file, file_path, *, step=-1, function=1, **kwargs):
    """
    Convert the Dirichlet boundary condition monitor log to a Neumann
    boundary condition input section.

    Args
    ----
    input_file: InputFile
        The input file where the created Neumann boundary condition is added
        to. The nodes referred to in the log file have to match with the ones
        in the input section. It is advisable to only call this function once
        all nodes have been added to the input file.
    file_path: str
        Path to the Dirichlet boundary condition log file.
    step: int
        Step values to be used. Default is -1, i.e. the last step.
    function: Function
        Function for the Neumann boundary condition.
    """

    nodes, _, force, _ = read_dbc_monitor_file(file_path)

    # The forces are the negative reactions at the Dirichlet boundaries.
    force *= -1.0

    add_point_neumann_condition_to_input_file(
        input_file, nodes, [function] * 3, force[step], **kwargs
    )
```
This allows to keep the simple interface of the old functionality, while the actual code for adding the BC to the mesh is the same for the "old" and "new" function.

tests/testing_utility.py
```diff
@@ -75,6 +77,106 @@ def test_is_node_on_plane(self):
             )
         )
 
+    def test_linear_time_transformation_scaling(self):
```
isteinbrecher, line 80: Really nice unit tests! Let's move them to the 4C test file.

tests/testing_four_c_simulation.py
```diff
@@ -669,7 +762,7 @@ def create_model(n_steps):
             "TIMES 1.0 11.0 VALUES 1.0 0.0"
         )
         restart_simulation.add(function_nbc)
-        dbc_monitor_to_input(
+        all_dbc_monitor_values_to_input(
```
isteinbrecher, line 765:
```suggestion
        dbc_monitor_to_input(
```

tests/testing_four_c_simulation.py
```diff
@@ -678,7 +771,8 @@ def create_model(n_steps):
                 f"{initial_run_name}_102_monitor_dbc.csv",
             ),
             n_dof=9,
-            function=function_nbc,
+            time_span=[10 * 0.5, 21 * 0.5],
+            fun_array=[function_nbc, function_nbc, function_nbc],
```
isteinbrecher, line 775: With my comment above, we should be able to use exactly the same call as we used to here, i.e., without having to manually specify the time points.

tests/meshes_for_testing.py
```diff
@@ -0,0 +1,68 @@
...
+    set_header_static,
+    InputFile,
+    Function,
+    MaterialReissner,
+    Beam3rHerm2Line3,
+)
+from meshpy.mesh_creation_functions import create_beam_mesh_line
+
+
+def create_cantilver_model(n_steps, time_step=0.5):
```
isteinbrecher, line 45: Can you move this to the test 4C file, as it is only used there for now? We can think of gathering all functionality we have in testing to create models in a single file (there should be a lot). But I would like to do that in a separate PR.

tests/testing_four_c_simulation.py
```diff
@@ -598,26 +604,113 @@ def test_four_c_simulation_dirichlet_boundary_to_neumann_boundary(self):
...
+        function_nbc = Function(
+            """SYMBOLIC_FUNCTION_OF_TIME nbc_value
+            VARIABLE 0 NAME nbc_value TYPE linearinterpolation NUMPOINTS 2 """
+            "TIMES 1.0 11.0 VALUES 1.0 0.0"
+        )
+        restart_simulation.add(function_nbc)
+        all_dbc_monitor_values_to_input(
+            restart_simulation,
+            os.path.join(
+                testing_temp,
```
isteinbrecher, line 664: Where does all this come from? Is it possible that you added the same test again (the two tests look like they have the same name)

meshpy/four_c/dbc_monitor.py
knarfnitram: I removed it, this was not necessary at all.

meshpy/four_c/dbc_monitor.py
knarfnitram: Ok so I brought the function `dbc_monitor_to_input` back to life and provided 3 tests cases.
1) `test_four_c_simulation_dbc_monitor_to_input`  - tests the old functionality with the interface how it used to be. **Important:** Only one force value is extracted.  
2) `test_four_c_simulation_all_dbc_monitor_values_to_input` tests the same as test case 1) to ensure that the new function is capable of doing the same functionality.
3) `test_four_c_simulation_dirichlet_boundary_to_neumann_boundary_with_all_values` -  tests the new functionality as a function by selected force values is created and added.

tests/testing_four_c_simulation.py
```diff
@@ -59,10 +59,42 @@
...
 )
-from meshpy.four_c import dbc_monitor_to_input, run_four_c
 from meshpy.mesh_creation_functions.beam_basic_geometry import create_beam_mesh_line
 from meshpy.mesh_creation_functions.beam_honeycomb import create_beam_mesh_honeycomb
+from meshpy.utility import check_node_by_coordinate
+
+
+def create_cantilver_model(n_steps, time_step=0.5):
+    """
+    Create a simple cantilver model.
```
isteinbrecher, line 76:
```suggestion
    Create a simple cantilever model.
```

tests/testing_four_c_simulation.py
```diff
@@ -59,10 +59,42 @@
...
+    dbc_monitor_to_input,
+    run_four_c,
 )
-from meshpy.four_c import dbc_monitor_to_input, run_four_c
 from meshpy.mesh_creation_functions.beam_basic_geometry import create_beam_mesh_line
 from meshpy.mesh_creation_functions.beam_honeycomb import create_beam_mesh_honeycomb
+from meshpy.utility import check_node_by_coordinate
+
+
+def create_cantilver_model(n_steps, time_step=0.5):
```
isteinbrecher, line 74:
```suggestion
def create_cantilever_model(n_steps, time_step=0.5):
```

tests/testing_four_c_simulation.py
```diff
@@ -592,32 +624,110 @@ def test_four_c_simulation_rotated_beam_axis(self):
         self.run_four_c_test("rotated_beam_axis", input_file)
         self.run_four_c_test("rotated_beam_axis", input_file, nox_xml_file="xml_name")
 
-    def test_four_c_simulation_dirichlet_boundary_to_neumann_boundary(self):
+    def test_four_c_simulation_dbc_monitor_to_input(self):
```
isteinbrecher, line 627: I think I now understand the added test cases, so sorry if my previous comment caused some confusion. We should unify `test_four_c_simulation_dbc_monitor_to_input` and `test_four_c_simulation_all_dbc_monitor_values_to_input` as they produce exactly the same files. So basically we should create the following structure:
```python
    def xtest_four_c_simulation_dbc_monitor_to_input_linear(self, *, type=None):
        ...
        if type == "dbc_monitor_to_input":
            dbc_monitor_to_input(
                restart_simulation,
                os.path.join(
                    testing_temp,
                    initial_run_name,
                    f"{initial_run_name}_monitor_dbc",
                    f"{initial_run_name}_102_monitor_dbc.csv",
                ),
                n_dof=9,
                function=function_nbc,
            )
        elif type == "dbc_monitor_to_input_all":
            all_dbc_monitor_values_to_input(
                restart_simulation,
                os.path.join(
                    testing_temp,
                    initial_run_name,
                    f"{initial_run_name}_monitor_dbc",
                    f"{initial_run_name}_102_monitor_dbc.csv",
                ),
                n_dof=9,
                time_span=[10 * 0.5, 21 * 0.5],
                fun_array=[function_nbc, function_nbc, function_nbc],
            )
        ...

    def test_four_c_simulation_dbc_monitor_to_input_linear(self):
        test_four_c_simulation_dbc_monitor_to_input_linear(self, type="dbc_monitor_to_input")

    def test_four_c_simulation_dbc_monitor_to_input_all_linear(self):
        test_four_c_simulation_dbc_monitor_to_input_linear(self, type="dbc_monitor_to_input_all")
```
This should save a lot of duplicated code.

isteinbrecher (commented):
@knarfnitram my last round of comments. I suggest you add the comments, squash and then rebase on the current `main`, I think some code checks might fail with this state. Thanks for the effort! Let's get this merged!

meshpy/four_c/dbc_monitor.py
```diff
@@ -88,9 +151,210 @@ def read_dbc_monitor_file(file_path):
     return nodes, data[:, 1], data[:, 4:7], data[:, 7:]
 
 
-def dbc_monitor_to_input(input_file, file_path, step=-1, function=1, n_dof=3):
+def add_point_neuman_condition_to_input_file(
+    input_file: InputFile,
```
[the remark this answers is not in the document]
knarfnitram: Thanks, I hope this helps to reuse these functions.

tests/testing_four_c_simulation.py
knarfnitram: Yes, this is now better to understandable. I just used the name of the folder(`initial_run_name`) to distinguish between them.

knarfnitram (commented):
Thanks for the comments and help @isteinbrecher , I hope that I have now fixed all your points!

meshpy/four_c/dbc_monitor.py
```diff
@@ -37,7 +37,70 @@
...
+        # add first coordinate again at the beginning of the array
+        if len(values.shape) == 1:
+            values = np.append(values[0], values)
+        else:
+            values = np.append(values[0], values).reshape(
+                values.shape[0] + 1, values.shape[1]
+            )
+
+    # repeat last value at provided time point
+    if valid_start_and_end_point and len(time_span) > 2:
```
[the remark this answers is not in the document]
isteinbrecher: Sounds good!

isteinbrecher:
Looks good to me! The only thing missing is a failing code check!

isteinbrecher:
Sorry, one more thing, is it possible that two of the checked in reference files are identical? If so, we should only check in the file once. If that is the case and if you want, I can take care of this once the code checks are fixed.

knarfnitram:
@isteinbrecher all checks finally pass :tada: 
I also removed the additional file, so we really test the functions with the same file. :smiley:

isteinbrecher (approved):
Thanks for the contribution!

PR #127, opened 2024-11-14
Add upload folder on failure

knarfnitram:
During #126 I encountered the problem that my changes work locally, but for some reason the pipeline fails.

I therefore added the automatic upload of the testing folder.
Additionally, I introduced the environment variable `MESHPY_TESTING_DIR`, with which allows to specify the location of the testing folder.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -72,6 +72,12 @@ jobs:
           # If we did not exit earlier, raise an error here
           exit 1
           ls
+      - name: Upload Test Results
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{github.job}}-${{github.run_number}}
+          path: ${{github.workspace}}/tests/testing-tmp/
 
   geometric-search-testing:
     name: geometric-search-testing


tests/utils.py
```diff
@@ -61,7 +62,14 @@ def skip_fail_test(self, message):
...
 testing_input = os.path.join(testing_path, "reference-files")
-testing_temp = os.path.join(testing_path, "testing-tmp")
+# if testing directory is specified use it
+testing_temp = get_env_variable("MESHPY_TESTING_DIR")
+if testing_temp == "default_not_set":
+    # create default one if not specified
+    testing_temp = os.path.join(testing_path, "testing-tmp")
+else:
+    # convert string to path
+    testing_temp = Path(testing_temp)
```
knarfnitram, line 72: @isteinbrecher 
I think it's on the long run cleaner to specify the directory from a variable and provide a default location.

.github/workflows/testing.yml
```diff
@@ -72,6 +75,12 @@ jobs:
           # If we did not exit earlier, raise an error here
           exit 1
           ls
+      - name: Upload Results
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{github.job}}-${{github.run_number}}
+          path: ${{env.MESHPY_TESTING_DIR}}
```
isteinbrecher, line 83: Can't you just set this to `path: ${{ github.workspace }}/tests/testing-tmp/`? Then all the other changes are not needed.

knarfnitram: I adapted it accordingly.

knarfnitram:
I have added a second commit, where I add the upload on failing.  
Everything is working as intended:
[Here](https://github.com/imcs-compsim/meshpy/pull/127/checks) the artifacts were uploaded (so we know that the path is correct) with the second commit, we have no upload since the pipeline passed.

.github/workflows/testing.yml
```diff
@@ -72,6 +74,12 @@ jobs:
...
```
isteinbrecher, line 82: Sorry to be nitpicking, but can't we just do it like this:
```suggestion
      - name: Upload Results
        if: failure()
        uses: actions/upload-artifact@v3
        with:
          name: ${{github.job}}-${{github.run_number}}
          path: ${{github.workspace}}/tests/testing-tmp/
```
Then we don't have to introduce the new variable `MESHPY_TESTING_DIR` which is just used for GitHub testing.

isteinbrecher (commented):
@knarfnitram there seem to be some not needed changes in the `README.md` and `utils.py` files.

tests/utils.py
```diff
@@ -33,15 +33,15 @@
...
+import os
 import shutil
 import subprocess
-import warnings
-import xml.etree.ElementTree as ET
 import vtk
 from vtk.util import numpy_support as vtk_numpy
 from vtk_utils.compare_grids import compare_grids
+import warnings
+import xml.etree.ElementTree as ET
```
isteinbrecher, line 44: I think these changes are not needed, correct? (are you using a tool that orders the imports?)

knarfnitram: No they are not needed. removed them as well.

isteinbrecher:
@knarfnitram thanks for taking care of this, I run an additional pipeline to show the upload still works as expected. Everything looks good, once the tests pass here, we can merge this.

knarfnitram:
Thanks for your help and additional testing @isteinbrecher .

PR #129, opened 2024-11-14
Make tests fail

isteinbrecher:
Test PR to check if the uploading of artefacts works.


isteinbrecher:
@knarfnitram artefacts are working as expected. Can you please squash everything in a single commit, then we can proceed with #127

PR #131, opened 2024-11-15
Sync the version of black

isteinbrecher:
Sync the version of black defined in pyptoject.toml and the pre-commit hooks


isteinbrecher:
@davidrudlstorfer I discovered this version mismatch, please correct me if there is another reason for the specific version specified in the pre-commit hooks.

davidrudlstorfer (approved):
As far as I know this should be the same version 👍

PR #134, opened 2024-11-18
Rename test

isteinbrecher:
Now the renamed test comes with a renamed reference file that should fix the failing pipeline.


isteinbrecher:
@knarfnitram sorry for the confusion, now the `main` branch should be fine.

Issue #119, opened 2024-11-05
Port testing to pytest


davidrudlstorfer:
To simplify the current testing setup and to benefit from additional pytest features the testing suite should be moved to pytest.

isteinbrecher:
Just to mention it, we have some "tests" that start with `xtest_*` that are used to define a base test version that can then be called with different parameters. This construct very likely simply be replaced with `pytest` parametrize  fixtures.

davidrudlstorfer:
Just to collect the overall ideas for PyTest in an official way:

- Introduce pytest fixtures for globally utilized variables and functions (i.e., reference file path, result comparison functions, ...) (see #148 for more details)
- Introduce tagging for tests (https://pypi.org/project/pytest_tagging/) to differentiate between normal tests, performance tests and 4C tests
- Overhaul/update tutorial to Jupyter Notebook and test in separate CI job

davidrudlstorfer:
#157 introduced the necessary infrastructure and #170 is the issue to port the remaining tests

I'll close this one and open an issue for the tutorials

PR #136, opened 2024-11-25
Proof of concept for yaml output

davidrudlstorfer:
This is just a pretty rough and quick draft to convert MeshPy output from .dat to .yaml.

There are several things which need to be reworked for a final version. This should demonstrate an overall working versions to get a feeling for the necessary changes - and also to look at a .yaml file.

This MR produced the following exemplary .yaml file (saved as .txt due to Github not supporting the upload of .yaml files):

[test_input.4C.txt](https://github.com/user-attachments/files/17903787/test_input.4C.txt)

As a direct comparison, this would be the regular 4C .dat input file (again formatted as .txt to enable the upload to Github)

[test_input.txt](https://github.com/user-attachments/files/17903794/test_input.txt)

I quite like the look and feel of the .yaml file, and especially the syntax highlighting in VSCode. Try it for youself!

@isteinbrecher I would be interested in you opinion

### DIFF
--- a/meshpy/conf.py
+++ b/meshpy/conf.py
@@ -177,9 +177,6 @@ class MeshPy(object):
         self.geometric_search_max_nodes_brute_force = 1000
         self.geometric_search_binning_n_bin = 10
 
-        # Values for the formating of the input file.
-        self.dat_len_section = 80
-
         # Import meshes as pure dat or import the geometry.
         self.import_mesh_full = False
 
--- a/meshpy/inputfile.py
+++ b/meshpy/inputfile.py
@@ -37,6 +37,7 @@ import sys
 import os
 import datetime
 import re
+import yaml
 from _collections import OrderedDict
 
 # Meshpy modules.
@@ -52,12 +53,9 @@ from .utility import get_git_data
 from .nurbs_patch import NURBSPatch
 
 
-def get_section_string(section_name):
-    """Return the string for a section in the dat file."""
-    return (
-        "".join(["-" for _i in range(mpy.dat_len_section - len(section_name))])
-        + section_name
-    )
+class CustomDumper(yaml.Dumper):
+    def increase_indent(self, flow=False, *args, **kwargs):
+        return super().increase_indent(flow=flow, indentless=False)
 
 
 class InputLine:
@@ -218,7 +216,7 @@ class InputSection:
     def get_dat_lines(self):
         """Return the lines for this section in the input file."""
 
-        lines = [get_section_string(self.name)]
+        lines = []
         lines.extend([str(line) for line in self.data.values()])
         return lines
 
@@ -301,6 +299,7 @@ class InputFile(Mesh):
             mpy.geo.surface,
         ): "DESIGN SURF MORTAR CONTACT CONDITIONS 3D",
     }
+
     geometry_counter = {
         mpy.geo.point: "DPOINT",
         mpy.geo.line: "DLINE",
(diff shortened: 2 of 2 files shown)


davidrudlstorfer:
I forgot to mention a few small things:

- I would also adjust the testing workflow (replace `input_file.get_string()` with `input_file.get_dict()`) in combination with #119 
- The read in mechanism of input files needs to be updated - currently not supported on this branch.
- The entire header could be added as a METADATA field to the yaml file once this is supported on the 4C side

isteinbrecher:
@davidrudlstorfer thanks a lot for the draft. Here are my first thoughts:

I like the idea to use a standard format for input to 4C. One mayjor drawback at the moment is that 4C does not have a state of the art way to read “large data”, e.g., node coordinates and connectivity. Before we merge something like this, I would like to see that the yaml approach is at least not much slower than the current approach. I know that there is the possibility to link certain input sections in yaml to point to a dat file, but this seems like a suboptimal solution as it requires handling of multiple files. Before we do something like that I would wait until we have a binary approach.

I think the approach you took is going in the right direction. Maybe the best way would be to fill up key-value pairings (in a yaml object or directly in a dictionary) and then write it at the end. A general remark on the stuff that is in `input_file.py`, e.g., InputLine, InputSection, this comes from the very first days of MeshPy (I think it wasn’t even called MeshPy back then) and can likely be much simplified or even replaced by something like a dictionary.

In the long run I would prefer to only support one input file format. If we go with yaml we need to figure out a way how to handle cases like
```python
input_file.add("""
    —Section
    parameter1 a
    parameter2 b
""")
```
Do we want to to keep this? Do we want to define a dict here? Do we want to input a yaml string?
This is a very commonly used construct at the moment and changing this behavior will likely break most of the existing scripts (which does not mean we can’t do it).

davidrudlstorfer:
@isteinbrecher thanks for your first evaluation and detailed answer - I simply answer them point by point

1. I would also oppose the idea of splitting the input file into a `.yaml` file and a `.dat` file for the geometry. As you mentioned this would only make sense once 4C supports binary geometry data.

2. Do you think that the speed of the `.yaml` file could be a problem? I do not know all of the current use cases of MeshPy - therefore, I am not in a situation to evaluate the possible problem of input speed. Are there any current and planned future applications with MeshPy which create such large input files? As far as I have seen so far the problems are rather "small" and we do not get into input file sizes of multiple GB's where this could be a problem. Please correct me here

3. Yes - I also think that the best approach would be to switch the entire logic to dictionaries. This would greatly simplify the data handling in MeshPy

4. Yes - I've already looked into the `InputLine`, `InputSection`, etc. and I think they can be removed entirely. I haven't adjusted the logic yet.

5. Regarding data entry/handling

I would replace this entire construct with a dictionary for easier handling

```
input_file.add("""
    —Section
    parameter1 a
    parameter2 b
""") 
```

for example

```
data = {"Section": ["parameter1 a", "parameter2 b"]}
input_file.add(data)
```

Details could be discussed along the way.

The final question is whether I should further pursue the `.dat` to `.yaml` switch now or if there are any reasons to omit it for now and maybe pursue it on a later date. I think it would be a good idea to start with this before overhauling the testing framework to pytest but I leave the decision up to you @isteinbrecher

davidrudlstorfer:
As discussed today with @isteinbrecher we will first compare the performance between the `.yaml` and `.dat` based input files regarding creation (writing out from MeshPy) and read-in speed (in 4C).

If the performance is on the same level or better with `.yaml` we will continue.

Additionally the `cubitpy` files still need to be read in => consider it in this PR

isteinbrecher:
Thanks for documenting this!

Also just to mention this, there are currently two ways of reading `.dat` files. One where we basically only store the text and then write it again, and one where we convert the dat file to a MeshPy mesh. @davidrudlstorfer do you know if with the current state of this PR, that functionality is still possible?

davidrudlstorfer:
@isteinbrecher As far as I know both options to read a .dat file are currently not working within this PR. But I'll update the state of this PR to be functional again

davidrudlstorfer:
Once the most important issues are resolved we can revisit the change to .yaml again.

If the layout of MeshPy is changed (#159) the minimal changes within this PR have to be rebased anyways. Since this involves a lot of effort I'll close this PR

PR #137, opened 2024-12-02
Allow to set file path for result compare tests

isteinbrecher:
This PR allows to explicitly set the name for the reference file when comparing MeshPy meshes during testing.

@davidrudlstorfer is this similar to what you were mentioning when we last talked about the switch to `pytest`?

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -131,26 +131,39 @@ def compare_string_tolerance(
 
 
 def compare_test_result(
-    self, result_string, *, extension="dat", additional_identifier=None, **kwargs
+    self,
+    result_string,
+    *,
+    extension="dat",
+    reference_file_name=None,
+    additional_identifier=None,
+    **kwargs,
 ):
-    """
-    Compare a created string in a test with the reference results. The reference results
-    are stored in a file made up of the test name.
+    """Compare a created string in a test with the reference results. The reference
+    results are stored in a file made up of the test name. The filename will always
+    end with "_reference".
 
     Args
     ----
+    result_string: str
+        String to compare with a reference file
+    reference_file_name: str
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (dismissed):
Thanks for adding it - this is what I had in mind and it looks very good!

Would it make sense to maybe drop the mandatory `_reference` ending? For example if I provide `reference_file_name="test2.yaml"` the `_reference` is always added.

Otherwise I would have to provide three two arguments, i.e. `reference_file_name=test2` and `extension="yaml"` which results in `"test2_reference.yaml"`

But this is only a minor idea

isteinbrecher:
Yes, we can definitely drop the `_reference` going further (maybe in a follow up PR).

I get your point regarding the arguments, but providing the full name (with extension) is not possible because some tests have the `additional_identifier` argument, which allows to test multiple files in a single test while still having the test name as the first part of the reference file. Maybe the naming of the argument can be improved: `reference_file_name` -> `reference_file_base_name`.

Or do you have another idea on how to pass the arguments here?

davidrudlstorfer:
> Yes, we can definitely drop the `_reference` going further (maybe in a follow up PR).
> 
> I get your point regarding the arguments, but providing the full name (with extension) is not possible because some tests have the `additional_identifier` argument, which allows to test multiple files in a single test while still having the test name as the first part of the reference file. Maybe the naming of the argument can be improved: `reference_file_name` -> `reference_file_base_name`.
> 
> Or do you have another idea on how to pass the arguments here?

The renaming of the argument sounds like a good option for now to improve the clarity of the variable. So let's rename it and lets get it merged :rocket:

davidrudlstorfer (approved):

PR #138, opened 2024-12-02
Rework geometry set

isteinbrecher:
This PR reworks the `GeometrySet` object (one of the more messier parts of MeshPy).

We used to store the geometry (nodes, elements) contained in a geometry set in a `list`. To ensure unique entries in this list we checked if the list contains a single item before adding it. This works as expected, however, for a large number of contained items, the check gets increasingly slow and this scales with approximately $\mathcal{O}(n^2)$. This was noticed when modifying the performance tests to also include a volume set with approximately 100,000 nodes. The runtime for the performance test `meshpy_load_solid_full` increased from 4sec to 40sec.

In this PR the internal storage is changed to a `dict` (with the keys being the geometry items and the values always being `None`). This mimics an ordered set using only types from the standard python library. This reduced the complexity to $\mathcal{O}(n)$. The previously mentioned tests runtime now actually decreased (with the additional volume set).

### DIFF
--- a/meshpy/geometry_set.py
+++ b/meshpy/geometry_set.py
@@ -82,9 +82,7 @@ class GeometrySetBase(BaseMeshItemFull):
                 output so we can color the nodes which are part of element sets.
         """
         if link_to_nodes == "explicitly_contained_nodes":
-            node_list = self.get_nodes_explicitly_contained_in_set(
-                fail_if_not_point_set=False
-            )
+            node_list = self.get_node_dict().keys()
         elif link_to_nodes == "all_nodes":
             node_list = self.get_all_nodes()
         else:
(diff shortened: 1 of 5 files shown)


meshpy/geometry_set.py
```diff
@@ -70,41 +70,53 @@ def __init__(self, geometry_type, name=None, **kwargs):
...
-                break
-        else:
-            raise ValueError(
-                "The node that should be replaced is not in the current node set"
-            )
+        """Replace old_node with new_node."""
+
+        explicit_nodes_in_this_set = self.get_node_dict()
+        explicit_nodes_in_this_set[new_node] = None
+        del explicit_nodes_in_this_set[old_node]
```
davidrudlstorfer, line 107: I've actually never seen this `del` syntax.

Maybe

```suggestion
        explicit_nodes_in_this_set.pop(old_node)
```

This is more pythonic/expressive

davidrudlstorfer (commented):
Similarly to the previous PR - nice changes!

I have one question which you already shortly discussed in the PR description: Is a `dict` the right object for this use case? If all values are `None` it doesn't make too much sense to me. Maybe we can simply switch it to a set (https://www.w3schools.com/python/python_sets.asp)? This also ensures unique entries.

isteinbrecher:
The main problem with a standard `set` is, that if we put python objects in a `set` the ordering inside the set is non-deterministic (I think the elements in the set are ordered according so some hash that contains the object ID which changes from run to run). In the case of the `GeometrySet` that would still be sufficient in almost all cases, since we sort the node IDs before writing to the input file. The only one case where we have problems with this approach is if one loops over the nodes in a set and expects the ordering to be deterministic. @knarfnitram had this problem and it was quite hard to track.

So we need a data type that only contains unique entries but keeps the ordering of the entries added. Turns out that since python3.7 this is exactly the behaviour of the default `dict` (before that it was called `OrderedDict`). I think the overhead of storing the `None` keys is neglectible for the considered cases (although a true `OrderedSet` would still be nice to have in python).

meshpy/geometry_set.py
isteinbrecher: I would keep it as it is, as `del` does exactly what is expected here, `pop` would additionally return the value of the deleted key, which is not needed here. See also [this thread](https://stackoverflow.com/questions/5713218/what-is-difference-between-pop-vs-del-for-deleting-a-dictionary-item-python).

davidrudlstorfer (approved):
> The main problem with a standard `set` is, that if we put python objects in a `set` the ordering inside the set is non-deterministic (I think the elements in the set are ordered according so some hash that contains the object ID which changes from run to run). In the case of the `GeometrySet` that would still be sufficient in almost all cases, since we sort the node IDs before writing to the input file. The only one case where we have problems with this approach is if one loops over the nodes in a set and expects the ordering to be deterministic. @knarfnitram had this problem and it was quite hard to track.
> 
> So we need a data type that only contains unique entries but keeps the ordering of the entries added. Turns out that since python3.7 this is exactly the behaviour of the default `dict` (before that it was called `OrderedDict`). I think the overhead of storing the `None` keys is neglectible for the considered cases (although a true `OrderedSet` would still be nice to have in python). This is also described [in this post](https://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set).

Very interesting topic - I just played around a while with dicts and it's really not possible to extend or update a dictionary without a value but just the key although its possible to create the dict with just keys and no values at all.

Thanks for the  heads-up and explanation

If @knarfnitram needs the functionality to keep the order then lets merge this

davidrudlstorfer:
As a supplement for the discussion: https://www.geeksforgeeks.org/add-key-to-dictionary-python-without-value/

PR #139, opened 2024-12-03
Use 4C docker image for testing

isteinbrecher:
This PR updates the CI of MeshPy. The main testing focus is switched to GitHub hosted runners. As default docker image we take an image with a pre-build 4C version. This allows to do all testing that does not require Cubit on GitHub hosted runners.

### DIFF
--- a/.github/actions/build_arbrox_geometric_search/action.yml
+++ b/.github/actions/build_arbrox_geometric_search/action.yml
@@ -1,0 +1,15 @@
+name: build_arborx_geometric_search
+description: Build the geometric search functionality with ArborX
+runs:
+  using: composite
+  steps:
+    - name: Build ArborX
+      shell: bash
+      env:
+        kokkos_dir: /opt/4C-dependencies/lib/cmake/Kokkos
+      run: |
+        cd ${GITHUB_WORKSPACE}
+        mkdir -p build/arborx
+        cd build/arborx
+        cmake -DKokkos_DIR=${{ env.kokkos_dir }} ../../meshpy/geometric_search/src -G Ninja
+        ninja
--- a/.github/actions/code_check/action.yml
+++ b/.github/actions/code_check/action.yml
@@ -1,0 +1,13 @@
+name: code_check
+description: Perform MeshPy code checks
+runs:
+  using: composite
+  steps:
+    - name: Code Checks
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}
+        source python-testing-environment/bin/activate
+        pip install -e .[CI-CD]
+        black . --check --exclude=python-testing-environment && exit 0
+        exit 1
--- a/.github/actions/run_performance_tests/action.yml
+++ b/.github/actions/run_performance_tests/action.yml
@@ -1,0 +1,18 @@
+name: run_performance_tests
+description: MeshPy performance test suite
+runs:
+  using: composite
+  steps:
+    - name: MeshPy testing
+      shell: bash
+      env:
+        CUBIT_ROOT: /imcs/public/compsim/opt/cubit-15.2
+        PERFORMANCE_TESTING_HOST: github-sisyphos-docker
+      run: |
+        cd ${GITHUB_WORKSPACE}
+        source python-testing-environment/bin/activate
+        pip install .[CI-CD]
+        python --version
+        pip list
+        cd tests
+        python performance_testing.py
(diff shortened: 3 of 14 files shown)


isteinbrecher:
@davidrudlstorfer could you have a look? This should now move the "main" testing to GitHub hosted runners. I will think of a way to also move the cubit testing there.

If you have an idea on how to also move the pages job to GitHub, please feel free to add this to the PR.

.github/actions/config_python_venv/action.yml
davidrudlstorfer, line 1: Do we actually need to create a python venv for testing? As far as I know (and always did in my own and other repos) we could just directly install the requirements into the Docker container?

Maybe you could also briefly explain the different testing strategies. In my opinion it would be more than enough to just install the dependencies via editable mode (`pip install -e . `) and do not test the standard approach. If it works in the editable mode it also must work in the standard mode as far as I know. This is only important if files are changed and, here, the latest code is always pulled from the repo without any pending changes.

davidrudlstorfer (commented):
Regarding the move from IMCS runners to Github hosted runners for the docu I would be more than happy to do that. Maybe it would make sense to do this in a separate PR to keep the changes here manageable.

isteinbrecher:
So things turned out a little bit more difficult than I expected. The main problem is that MeshPy needs Cubit for some tests (one test in the test suite and the performance tests). It is not possible to setup Cubit on the container since, one would need to sign in with an account providing a sufficient license, so we still need to run these tests on our local testing machine.  It was also not possible to run them in the docker image since the Cubit version used for testing is not compatible with Ubuntu24.04. Regarding the virtual environment, yes, once we are solely on docker we can remove this part. Regarding the two types of installing MeshPy, we used to have some issues that files were not copied properly when installing MeshPy with `pip install .`, so I like that this is now also tested. However, ArborX only works when installing MeshPy with `pip install -e .`. Since we need two main testing runs anyway (one with 4C on the docker image and one with Cubit on our local machine), I decided to add both install options during testing.

To summarise the new testing jobs:
- Code check: Check if MeshPy matches the `black` formatting
- Run MeshPy test suite: Run the MeshPy test suite on a IMCS testing machine, with the Cubit tests, but without ArborX and 4C.
- Run MeshPy test suite ArborX: Run the MeshPy test suite on a GitHub hosted runner using the latest 4C docker image. This tests the 4C compatibility and also the ArobrX geometric search. Here, MeshPy is installed in editable mode.
- Run MeshPy performance test suite: Run the MeshPy performance test suite on a IMCS testing machine

@davidrudlstorfer thanks for the offer, to add this to the documentation. We should definitely do this. Maybe it makes sense to wait a little bit, as we will be migrating to Ubuntu24.04 in the near future, so the Cubit thing might change (maybe we can even replace it for the MeshPy testing with a mocking approach). But you can open an issue so we can track this.

davidrudlstorfer (approved):
@isteinbrecher thanks for the nice PR and especially for the lengthy reasoning on the approach. Slowly but surely I better understand all the aspects of MeshPy.

Could you maybe specify the problems you encountered when using `pip install .` - actually never heard something like that?

I really like the new approach - particularly the split into the different action files which makes everything a lot clearer. From my side I found no comments regarding the code and would merge everything as it is.

PS. Also very nice that it now works with the official 4C docker image 🥳

isteinbrecher:
As far as I know, the default `pip install .` only installs python files. In case other files also needed reactive to the python files, they have to be specified explicitly in the `pyproject.toml` file. With MeshPy we used to have a couple of such files (e.g., slurm config files). That is the reason why I think it makes sense to also test the default `pip` installation as well.

Ps: I had a look and we actually still have one file that gets copied when installing MeshPy, `4C_input_solid_tutorial.dat` (one could argue if this really needs to be installed...).

davidrudlstorfer:
> As far as I know, the default `pip install .` only installs python files. In case other files also needed reactive to the python files, they have to be specified explicitly in the `pyproject.toml` file. With MeshPy we used to have a couple of such files (e.g., slurm config files). That is the reason why I think it makes sense to also test the default `pip` installation as well.
> 
> Ps: I had a look and we actually still have one file that gets copied when installing MeshPy, `4C_input_solid_tutorial.dat` (one could argue if this really needs to be installed...).

Thanks for the explanation! As we've already discussed in person we could switch the tutorial to a jupyter notebook and host it on Github in the future. So let's get this current status merged (maybe this also fixes the pipeline which failed after my last merge)🚀

isteinbrecher:
Sorry, the reason for the failing test is that I already removed the pre-build 4C on our testing machines. Should be fixed with this merge.

PR #142, opened 2024-12-05
Drop _reference from reference file names

### LINKED ISSUE #141: Remove the trailing `_reference` from reference result files
The trailing `_reference` at the end of the filename of reference result files does not really serve a purpose. Let's remove it.

davidrudlstorfer:
Closes #141


tests/testing_meshpy.py
```diff
@@ -300,7 +300,9 @@ def create_comments_in_solid(self, full_import):
         # Convert the solid mesh to meshpy objects.
         mpy.import_mesh_full = full_import
 
-        solid_file = os.path.join(testing_input, "test_meshpy_comments_in_solid.dat")
+        solid_file = os.path.join(
+            testing_input, "test_meshpy_comments_in_solid_initial.dat"
```
isteinbrecher, line 304: This is changed, because otherwise the "new" reference file would have the same name, right?

isteinbrecher (approved):
Thanks a lot @davidrudlstorfer, looks good to me!

tests/testing_meshpy.py
davidrudlstorfer: Yes exactly - I actually thought too long about the name - we could also take something like `base` or `original`. If you'r fine with `initial` I would merge

isteinbrecher: `initial` is fine with me. I just wanted to double check. Let's merge this!

PR #146, opened 2024-12-06
Add macOS and Windows testing

isteinbrecher:
This PR adds macOS and Windows CI testing. We now also test different python versions (3.10-3.12) in combination with the different OS. There were some numerical issues with the Windows version, these are resolved by making the `Rotation.get_rotation_vector` method more robust (several results had to be adapted).


davidrudlstorfer (approved):
Very nice changes and it's really cool that it works on different platforms and versions!

Overall I think this PR introduces a lot of complexity with the different OS versions and Python versions - is it really necessary to test 9 different combinations? Personally, the Windows version could be dropped - similarly the Python 3.11 version could be dropped. If Python 3.10 and 3.12 works it would be hard to believe for me that errors occur in Python 3.11.

Of course the decision is up to you but I think the complexity should not be artificially increased just because the runners are free.

isteinbrecher:
I agree that we can drop python 3.11, as with 3.10 and 3.12 we basically test the "supported" version range. I like having the Windows test cases as this should help keeping the code more general (requiring the Windows test cases to pass already led to an improvement of the rotation vector evaluation). If the Windows compatibility turns out to lead to an increased development effort we can remove it again.

I will push a fix that removes python3.11.

One thing I discovered, currently we get problems if test cases fail since the artefacts of different jobs have the same name and thus only the artefact of one of the test cases is uploaded. Do you know how to add the "job name" to the artefact name:
```yaml
name: ${{github.job}}-${{github.run_number}}
```

isteinbrecher:
The issue with the artifact name should be fixed. Also python3.11 testing is removed.

davidrudlstorfer (approved):
Thanks for the update! Even with one less Python version the actions/checks are clearer.

Let's get this merged 🚀

PR #152, opened 2024-12-11
Move dependencies into file and update pyproject

davidrudlstorfer:
While setting up the pre-commit hooks i moved the dependencies into a separate `requirements.in` file so the requirements can be automaticall set by `pip`. This pins all versions and improves the stability for version changes.

I also cleaned up the pyproject.toml file along the way and updated numpy (this was set to numpy<2?)


isteinbrecher:
I am not sure if I fully understand this PR (my knowledge on python packaging is a little rusty). What is the advantage of moving the requirements to the `requirements.txt` file? Also, is it really necessary to keep the exact version of the packages? Wouldn't it make sense to always test against the "default" version that comes with `pip`?

Regarding the point coupling conditions, I think that might be related to numpy>2, as far as I know there are some incompatibilities with `pybind11` and the `cpp` ArborX implementation.

davidrudlstorfer:
> I am not sure if I fully understand this PR (my knowledge on python packaging is a little rusty). What is the advantage of moving the requirements to the `requirements.txt` file? Also, is it really necessary to keep the exact version of the packages? Wouldn't it make sense to always test against the "default" version that comes with `pip`?

In my opinion it makes sense to pin the versions to ensure that everybody uses the same package - if not - users may have other versions locally and experience bugs which appear in newer versions (this is somewhat avoided by the nightly test pipeline - but users may still use locally cached versions and experience problems).

In the future this could be used on Github with (e.g. dependabot) to automatically update the dependencies and manage it better.

The opinion across the web is split into pin vs no-pin, i.e. https://stackoverflow.com/questions/28509481/should-i-pin-my-python-dependencies-versions

I really like the approach because packages can simply be updated with the command which is within the `requirements.txt` file. This also ensures that the versions across each other are compatible.

Additionally, I just "copied" this approach from other repositories, i.e. also queens => https://gitlab.lrz.de/queens_community/queens - they have a very solid python setup
 
> Regarding the point coupling conditions, I think that might be related to numpy>2, as far as I know there are some incompatibilities with `pybind11` and the `cpp` ArborX implementation.

You are right! That's a new one - this should be investigated to be able to update to the newer Numpy versions.

Additionally, black was needed as a requirement? This should be handled by the pre-commit hook.

isteinbrecher:
I am a little bit torn here. I understand the appeal of fixing the dependences, but I also like the current approach in just taking what `pip` gives you.

As far as I understand (and please correct me if I am wrong), the main advantage of fixing the dependencies is that we get exactly what we developed/tested with when we install MeshPy. But how would that that work if we install MeshPy together with another package that fixes the dependencies (e.g., QUEENS). Then as far as I know, pip would take the versions of the dependencies of the package that was installed last. Thus meaning we overwrite the fixed dependencies of the previously installed package. I personally view MeshPy as a library that one can install in an existing environment with other already configured packages (which might habe a more complex/restrictive dependency management). Thus I would prefer to stay with the current approach, and maybe reconsider this if we run into issues at some point. We currently print `pip list` (is this the same as `requirements.txt`?) in each CI test, so if something breaks it should be easy to track down.

Please let me know you opinions on this, I am also happy to talk about it.

Regarding ArborX, I will have a look, likely an update of pybind11 should to the trick, otherwise I would stick with the current numpy version for now, as the future of the ArborX functionality is not clear.

`black` can be removed, I just had it there so it was available when I develop MeshPy in my IDE.

knarfnitram:
I just made a checkup regarding the concerns of different python versions and mixing multiple projects:
Wouldn't it be the right approach to create first a combined `requirements.txt` for all python packages  environment when trying to combine multiple python packages ?

So the workflow would be to combine queens with meshpy would be to create a `combined-requirements.txt ` in with: 
`pip-compile --output-file=combined-requirements.txt queens/requirements.in meshpy/requirements.in`

and from those requirements we create the `.in` file, where the versions are set explicitly.
 `pip-compile --all-extras --output-file=combined-requirements.txt requirements-combinded.in`

As long as we do not select the exact versions in the `meshpy/requirements.txt`, I suppose that this approach should choose the python packages to the default ones provided by pip(or the other package) or result in a version conflict if verisons do not match.

davidrudlstorfer:
@knarfnitram thanks for the update and solution to this! (for clarification - I think you once switched up `.txt` with `.in`

> and from those requirements we create the ~~.in~~ **.txt** file, where the versions are set explicitly.
> pip-compile --all-extras --output-file=combined-requirements.txt requirements-combinded.in

This would exactly be the approach which should be taken for multi project examples. In my opinion this is still the wanted approach - but I am also happy to further discuss this or close the PR

knarfnitram:
@davidrudlstorfer Oh yes, I messed it up once. Thanks for clarifying!

@isteinbrecher are you familiar with `pip-compile` to combine the requirements? 
So requirements are actually defined within the `requirements.in` file and if you have a look at [queens](https://gitlab.lrz.de/queens_community/queens/-/blob/main/requirements.in?ref_type=heads), they really specify the versions for some packages. 
I interpret the `resulting.txt` file as a snapshot of versions with which it definitely should work and can be used for testing.  

I think this is a good approach @davidrudlstorfer, and I am also in favour of merging it.

isteinbrecher (commented):
Sorry @davidrudlstorfer , somehow I did not publish the review last week.

I am still not sure if I like the approach of fixing all dependency versions for MeshPy directly in the repository. I had a look at some other python packages that I like, e.g., dask, spack, splinepy, numpy-quaternion, and none do it this way (I think the philosophy behind MeshPy and splinepy / numpy-quaternion is quite similar, in that we want to provide a library that can simply be added to any existing python environment). Maybe lets me ask this question, are there any disadvantages of sticking to the current approach? The (nightly) test suite should discover broken compatibility right away.

However we decide, the `pyproject.toml` file can definitely use an overhaul, so I left some comments there.

isteinbrecher:
Lets close this PR and continue the discussions in separate issues/PRs in the future.

#159

PR #153, opened 2024-12-11
Add first version of pre-commit

### LINKED ISSUE #108: Discuss possible code checks to be used in meshpy
This issue shall serve as a spot to discuss possible code checks and pre-commit hooks to be used within MeshPy. Some possible choices can be seen here https://github.com/davidrudlstorfer/pyskel/blob/main/.pre-commit-config.yaml (thanks to @davidrudlstorfer).

Possbile tools:
- [x] `pylint` currently we have `8.93/10` for `meshpy/` and `8.55/10` for `tests/`
- [x] Ruff
- [x] ~Code formatter for markdown~ could not be found
- [x] Code formatter for c++
- [x] ~Code formatter for Cython. Does such a tool exist? Do we need this as the Cython code likely will not change on a regular basis.~ Deemed not worth the effort
- [x] Typo check (see #163)

Other code format to enforce:
- [x] Require docstrings for functions
- [ ] Permit relative imports depends on #159

davidrudlstorfer:
Add first version of more exhaustive pre-commit hook


davidrudlstorfer:
Notes for myself:

- Enfore absolute imports
- Enforce docstrings? Increase docstring coverage? If 100% => create docu
- type hinting?

davidrudlstorfer:
I've now also changed the `check-code` option to not execute the `no-commit-to-branch` on the main branch

I've also changed the `check-code` option to just use an Ubuntu image? Prior these checks were done within the 4C image although this is not needed?

isteinbrecher:
Thanks a lot @davidrudlstorfer for these changes! I will have a look, but it might take until next week.

Regarding your questions:
> We could (and should) enforce absolute imports, i.e., meshpy.rotation.blablub rather than ..rotation.blablub

Sounds good to me

> We could (and in my opinion should) enforce the docstrings, currently we have 96% coverage. This would improve the docu which could be created from the docstrings

Sounds good to me

> We could further enforce the docstring style, i.e. args to be consistent across the code base

I would like to see how that looks like, but in general I am for something like this

> We could add type hinting - I think we briefly talked about this and you had some points against that?

I remember we talked about this and you mentioned, that this only makes sense if it is done throughout the code base (and type are also enforced?). I am for adding type hints (the `cosserat_curve` module already comes with quite some), I am just not sure if we want/can do it everywhere.

> We should enforce that each source code file contains the license header

We already have that, the test `tests/testing_header.py` (should probably be moved to the code checks instead of being run inside the test suite)

meshpy/four_c/solid_shell_thickness_direction.py
```diff
@@ -204,7 +207,10 @@ def set_solid_shell_thickness_direction(
     for element in elements:
         is_hex8 = isinstance(element, VolumeHEX8)
         if is_hex8:
-            is_solid_shell = "SOLIDSH8" in element.dat_pre_nodes
+            is_solid_shell = (
+                hasattr(element, "dat_pre_nodes")
```
isteinbrecher, line 211: Why do we need this? Is this added automatically?

isteinbrecher:
And yes, the default ubuntu image is fine for the code checks

meshpy/four_c/solid_shell_thickness_direction.py
davidrudlstorfer: Otherwise mypy would have thrown the following error:

```
Run mypy (static type checker for python)............................................Failed
- hook id: mypy
- exit code: 1

meshpy/four_c/solid_shell_thickness_direction.py:210: error: "Element" has no attribute "dat_pre_nodes"  [attr-defined]
Found 1 error in 1 file (checked 67 source files)
```

.pre-commit-config.yaml
```diff
@@ -1,8 +1,92 @@
...
+        name: Check for trailing whitespaces
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.8.0
+    hooks:
+      - id: bandit
+        name: Run bandit (security linter for python)
+        args: ["-c", "pyproject.toml", "-r", "-ll", "meshpy/"]
+        additional_dependencies: ["bandit[toml]"]
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v19.1.5
```
isteinbrecher, line 46: Did you try to apply this version to the C++ files in the repo? Maybe we need to "update" them.

.pre-commit-config.yaml
```diff
@@ -1,8 +1,92 @@
...
+    hooks:
+      - id: clang-format
+        name: Run clang-format (C/C++ formatter)
+  - repo: https://github.com/PyCQA/docformatter
```
isteinbrecher, line 50: As fas as I can tell, I really like this. It looks like it could improve writing documentation, which is a pain until now.

.pre-commit-config.yaml
```diff
@@ -1,8 +1,92 @@
...
+    rev: "eb1df347edd128b30cd3368dddc3aa65edcfac38" # change back to version once bug is fixed in latest version to be compatible with pre-commit
+    hooks:
+      - id: docformatter
+        name: Run docformatter (formatter for docstrings)
+        args: ["--in-place", "--config=./pyproject.toml", "-r", "meshpy/"]
+  - repo: https://github.com/econchick/interrogate
```
isteinbrecher, line 56: Very nice, once we have that we can also enforce this in testing with a future PR (I think it should not be too much effort to add all the currently missing docstrings to get to a 100%).

.pre-commit-config.yaml
```diff
@@ -1,8 +1,92 @@
...
+    hooks:
+      - id: mypy
+        name: Run mypy (static type checker for python)
+        args: ["--install-types", "--non-interactive", "--ignore-missing-imports", "--follow-imports=silent"]
+  - repo: https://github.com/asmeurer/removestar
+    rev: "1.5.2"
+    hooks:
+      - id: removestar
+        name: Run removestar (remove wildcard imports)
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
isteinbrecher, line 73: Just so I understand, this is compatible with the `black` code style?

isteinbrecher (commented):
Thanks a lot @davidrudlstorfer for taking a look into this, this will definitely put the code quality of MeshPy to another level. I left some comments regarding the configuration. I have two more questions:
- Is it possible to have one commit that applies all the automatically done stuff and one that adds the manually required changes for everything to pass?
- For practical purposes, if I want to commit something and I get an error message in the pre commit hook, do I get a command that I can use to apply the suggested changes or are they applied automatically, how does that work?

.pre-commit-config.yaml
```diff
@@ -1,8 +1,92 @@
...
-  - repo: https://github.com/psf/black
-    rev: 22.12.0
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
     hooks:
-      - id: black
-        language_version: python3.12
-
-
+      - id: check-added-large-files
```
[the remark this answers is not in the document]
davidrudlstorfer: Exactly. In case it is really necessary we can simply add an exclude to that specific hook ,i.e., 

`exclude: path/to/too_large_file.exe`

If the file should be much larger than 1MB  we should anyways think about using gitLFS.

.pre-commit-config.yaml
davidrudlstorfer: The C++ files are checked and everything is correct. I've added "unformatted" changes to the .cpp/.h files and they were formatted correctly.

If the files would not adhere to the formatting - the pipeline would fail as it checks all files within the repo

.pre-commit-config.yaml
davidrudlstorfer: Nice that you support the docstring changes 🙏🏻

.pre-commit-config.yaml
davidrudlstorfer: Yes - that should also be a separate/future PR

.pre-commit-config.yaml
davidrudlstorfer: Yes - ruff more or less applies the black formatting style, per their docs:

> The initial goal of the Ruff formatter is not to innovate on code style, but rather, to innovate on performance, and provide a unified toolchain across Ruff's linter, formatter, and any and all future tools.
> 
> As such, the formatter is designed as a drop-in replacement for [Black](https://github.com/psf/black), but with an excessive focus on performance and direct integration with Ruff. Given Black's popularity within the Python ecosystem, targeting Black compatibility ensures that formatter adoption is minimally disruptive for the vast majority of projects.
> 
> Specifically, the formatter is intended to emit near-identical output when run over existing Black-formatted code. When run over extensive Black-formatted projects like Django and Zulip, > 99.9% of lines are formatted identically. (See: [Black compatibility](https://docs.astral.sh/ruff/formatter/#black-compatibility).)

Given this focus on Black compatibility, the formatter thus adheres to [Black's (stable) code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html), which aims for "consistency, generality, readability and reducing git diffs"

davidrudlstorfer (commented):
> Thanks a lot @davidrudlstorfer for taking a look into this, this will definitely put the code quality of MeshPy to another level. I left some comments regarding the configuration. I have two more questions:
> 
> - Is it possible to have one commit that applies all the automatically done stuff and one that adds the manually required changes for everything to pass?

Unfortunately, I've combined both changes into a single commit and I would prefer not to split them up again as it was a lot of manual work 🥶 if you're ok with it I would leave it as is

> - For practical purposes, if I want to commit something and I get an error message in the pre commit hook, do I get a command that I can use to apply the suggested changes or are they applied automatically, how does that work?

That depends on the specific hook - all formatting changes are applied completely automatically if they fail, i.e., formatting any sort of code (C++, Python) or formatting docstrings.

Code related problems are not automatically adjusted - there the user has to intervene manually, i.e., `mypy` and `bandit` and the tools provide an error where type problems may arise or sometimes also show one or more solutions on how to solve the problem

.pre-commit-config.yaml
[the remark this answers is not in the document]
isteinbrecher: Perfect, I think that is a very good approach.

.pre-commit-config.yaml
isteinbrecher: Sounds good! Can we somehow easily print for which functions this is missing?

isteinbrecher:
Thanks for the explanations @davidrudlstorfer. I tried to "split" your first commit by first applying just the automatic changes from the pre-commit hooks and then resetting the state to your commit, I think that should have worked (see the branch on my fork https://github.com/isteinbrecher/meshpy/tree/add_pre_commit_hook). Can you have a short look if those are your manual changes (in the commit 16253c457d5e191dd610914bedb2f3e180bb28c4) and if so push this structure? (The actually code in the final state is exactly the same).

.pre-commit-config.yaml
davidrudlstorfer: Yes - once we increase the percent you automatically get an exhaustive overview on which docstrings are missing

davidrudlstorfer:
@isteinbrecher I've now rebased everything onto main - as there were also some changes which were necessary to the cosserat module.

I've also split up everything into logical commits and also split the automatic and manual changes as you suggested.

I encounter one problem which I was not able to solve directly - maybe you could have a brief look:

The code check now fails with a problem in the new cosserat module:

```bash
meshpy/cosserat_curve/cosserat_curve.py:391: error: Unsupported operand types for <= ("list[float]" and "float")  [operator]
meshpy/cosserat_curve/cosserat_curve.py:393: error: Unsupported operand types for >= ("list[float]" and "float")  [operator]
meshpy/cosserat_curve/cosserat_curve.py:398: error: Unsupported operand types for - ("list[float]" and "float")  [operator]
```

(see actions output)

Unfortunately, I was not able to quickly test this behaviour as it was not triggered within the testing pipeline (I simply wanted to exit once this part is reached)

Is this really wrong code or is there another problem here?

davidrudlstorfer:
@isteinbrecher the last commit fixes the above mentioned problem in the cosserat module - maybe you could briefly check if these changes make sense - then I would add them to the "manual changes" commit

davidrudlstorfer:
@isteinbrecher so the problem within the cosserat curve was that the variable arc_length was previously initialized as a list, then reset to a float. MyPy uses the first initialisation of a variable => therefore it was wrong

if you are happy with these changes I would combine the last two commits into the manual changes

isteinbrecher:
I will have a look at the other changes now, I think it is best if we combine the commits once everything is resolved.

meshpy/__init__.py
```diff
@@ -28,22 +28,30 @@
...
-
-# 3D rotations for nodes.
-from .rotation import Rotation
-
 # Mesh items.
 from .base_mesh_item import BaseMeshItemString
+
+# Boundary conditions and couplings for geometry in the mesh.
+from .boundary_condition import BoundaryCondition
+from .conf import mpy
```
isteinbrecher, line 39: I really like the automatic ordering of imports!

meshpy/abaqus/beam.py
```diff
@@ -31,10 +31,8 @@
 """This file provides functions to create Abaqus beam element classes to be
 used with MeshPy."""
 
-# Python modules
```
isteinbrecher, line 36: I also really like that these comments (that I introduced back in the day...) are gone!

meshpy/cosserat_curve/cosserat_curve.py
```diff
@@ -440,7 +440,7 @@ def get_pyvista_polyline(self) -> pv.PolyData:
         cell = np.insert(cell, 0, self.n_points)
         poly_line.lines = cell
 
-        base = [[], [], []]
+        base: list[list[np.ndarray]] = [[], [], []]
```
isteinbrecher, line 443: This whole construct was a little bit shady, I replaced this with a `numpy` array.

meshpy/element_beam.py
```diff
@@ -57,14 +56,14 @@ class Beam(Element):
...
-    nodes_create = []
+    nodes_create: Any = []
 
     # A list of valid material types for this element.
-    valid_material = []
+    valid_material: Any = []
 
     # Coupling strings.
-    coupling_fix_string = None
-    coupling_joint_string = None
```
isteinbrecher, line 69: This should stay at `None`, it is later on checked if it is None and an error is thrown (in `get_coupling_string`). I added an `Optional[str]` type hint.

meshpy/four_c/solid_shell_thickness_direction.py
```diff
@@ -207,7 +207,10 @@ def set_solid_shell_thickness_direction(
     for element in elements:
         is_hex8 = isinstance(element, VolumeHEX8)
         if is_hex8:
-            is_solid_shell = "SOLIDSH8" in element.dat_pre_nodes
+            is_solid_shell = (
```
isteinbrecher, line 210: That is the only one I don't understand, since we check there the element is of `VolumeHEX8` type which has the `element.dat_pre_nodes` variable. I added an exception here.

isteinbrecher (dismissed):
Thanks a lot @davidrudlstorfer for the effort put into this. I really like the new look and feel. I had a look at all the changes, and adapted the ones where I had some issues directly - now I am asking for your review on those. If you agree with them, I think we can squash the relevant commits together and merge this!

davidrudlstorfer:
@isteinbrecher also big thanks to the effort you put into the review 🙏🏻

I think this lays a good foundation to further improve the code quality within the repo and for future work.

Thanks for improving the manual code changes necessary to comply with the new pre-commit hook and associated changes.

From my side everything is alright.

As a last step I would squash the last 6 commits into the manual changes and then I would be happy to merge 🚀

isteinbrecher (approved):
Thanks again @davidrudlstorfer! Let's get this merged, great to see MeshPy improving!

davidrudlstorfer:
@isteinbrecher unfortunately you have to merge as I am unable to merge 😅

Screenshot 2024-12-16 at 15 45 14

isteinbrecher:
I am also not allowed to merge, since I added commits and therefore my review is not valid... But I will make use of 
image :)

isteinbrecher:
One thing I just saw in the test report, we don't actually check any `xml` files:
```
Check valid XML syntax...........................................(no files to check)Skipped
...
Run xml-formatter................................................(no files to check)Skipped
```
all the `vtk`, `vtu`, ... are xml based. Should we add them to the checks?

PR #156, opened 2024-12-16
Improve smallest rotation computation

isteinbrecher:
Improve the computation of the smallest rotation mapping in the large rotation framework. Previously we computed the smallest rotation mapping by first computing the smallest rotation triad and then extracting the corresponding quaternion. With the proposed changes, the quaternion representing the smallest rotation mapping can directly be calculated which is cleaner in my opinion and should save about half of the operations according to AceGen (not that it is super important to count operations here).


davidrudlstorfer (approved):
Looks good! I would be happy if this could get merged once the pre-commit hooks are merged. Maybe there is some formatting that needs to be done (although I do not see any problems) and the rebasing of the other PR is quite elaborate

isteinbrecher:
Thanks for the review, we can definitely wait until the pre-commit hooks are merged.

PR #157, opened 2024-12-17
Add first version of pytest infrastructure

davidrudlstorfer:
Includes first fixtures and tagging approach


davidrudlstorfer:
@isteinbrecher this is a work in progess - I've added the first pytest fixtures and also enabled the tagging approach. This turned out to be much more complicated than anticipated. 

I wanted that the "standard" tests would always execute - this approach was not doable with https://pypi.org/project/pytest_tagging/

Therefore, I simply implemented my own version which turns out to be very easy and doable

The tests can now be executed with:

`pytest` => execute tests which are not tagged
`pytest --4C` => execute tests which are tagged with fourc, i.e., `@pytest.mark.fourc`
`pytest --performance` => execute tests which are tagged with fourc, i.e., `@pytest.mark.performance`


It is also possible to add multiple tags to tests and also execute multiple tags at once

I am happy with the current status and I'll add the reference comparison function soon

davidrudlstorfer:
Also pytest coverage is added and works as expected - afterwards we can simply add 

1.  the coverage report of meshpy
2. the coverage badge to the readme

.github/actions/run_tests/action.yml
```diff
@@ -49,5 +49,4 @@ runs:
         coverage html
         coverage report
         coverage-badge -o htmlcov/coverage.svg
```
isteinbrecher, line 51: We should be able to remove the `coverage run --rcfile=${{ inputs.coverage_config }} testing_main.py` call (and the corresponding calls to `coverage`, since `pytest` already picks up all "old" tests.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "numpy-quaternion",
   "pre-commit",
   "pytest",
+  "pytest-cov",
```
isteinbrecher, line 42: Does not have to be decided in this PR, but do we want this in the default installation configuration? Or do we want a special "development" config?

pyproject.toml
```diff
@@ -55,3 +56,11 @@ CI-CD = [
   "cubitpy@git+https://github.com/imcs-compsim/cubitpy.git@main",
   "setuptools" # Needed for coverage-badge
 ]
+
+[tool.pytest.ini_options]
+addopts = "-p pytest_cov --cov-report=term --cov-report=html --cov-fail-under=30 --cov=meshpy/ --cov=tutorial/ --cov-append"
```
isteinbrecher, line 61: Why only 30%? I think we can set that to 90. (With the exception of the performance tests).

Should we also add the `tests/` directory here? I personally think it makes sense to identify untested code in the testing framework.

pyproject.toml
```diff
@@ -55,3 +56,11 @@ CI-CD = [
...
+markers = [
+  "fourc_arborx: tests in combination with 4C and ArborX",
```
isteinbrecher, line 63: Later on this is named `4C_ArborX` -> does this have to be the same string?

tests/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        parser: Pytest parser
+    """
+
+    parser.addoption(
+        "--4C_ArborX",
+        action="store_true",
+        default=False,
+        help="Execute 4C and ArborX based tests.",
```
isteinbrecher, line 52: According to the behaviour I saw, shouldn't this be:
```suggestion
        help="Execute basic tests and 4C and ArborX based tests.",
```

tests/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+        help="Execute performance tests.",
+    )
+
+
+def pytest_collection_modifyitems(config: Config, items: list) -> None:
+    """Filter tests based on their markers and provided command line options.
+
+    Currently configured options:
+        `pytest`: execute tests with no markers
+        `pytest --4C_ArborX`: execute tests with the `fourc_arborx` marker
```
isteinbrecher, line 75: If I read this code correctly, we always execute not tagged tests, right? Then this has to be adapted.

In general I really like the current approach (and kudos to the implementation!) of always executing the base tests and then allowing to selectively add test tags. Two comments from my side:
1. I would split 4C and ArborX, as a few people use MeshPy directly with 4C but don't have ArborX configured.
1. When running the performance tests, I would like to not run the untagged tests, is this possible?

tests/conftest.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns:
+        Path: A Path object representing the full path to the reference file directory.
+    """
+
+    testing_path = os.path.abspath(os.path.dirname(__file__))
+    return Path(testing_path) / "reference-files"
+
+
+@pytest.fixture(scope="function")
+def current_test_name(request: pytest.FixtureRequest) -> str:
```
isteinbrecher, line 122: Is it really necessary to provide the argument `requests` here? I am not sure if we have this available everywhere where we need to call this function. Maybe (for the beginning?) we should stick to the `env` variable approach. What do you think?

tests/test_cosserat_curve.py
```diff
@@ -50,30 +49,27 @@
...
 )
 
 
-def load_cosserat_curve_from_file():
+def load_cosserat_curve_from_file(reference_file_directory):
     """Load the centerline coordinates from the reference files and create the
     Cosserat curve."""
     coordinates = np.loadtxt(
-        os.path.join(testing_input, "test_cosserat_curve_centerline.txt"),
+        os.path.join(reference_file_directory, "test_cosserat_curve_centerline.txt"),
```
isteinbrecher, line 59: Do we want to passt this as a function argument or simply call the `reference_file_directory()` inside this function?

tests/test_dummy.py
```diff
@@ -0,0 +1,79 @@
...
+        current_test_name: name of the current test
+    """
+
+    # approach to get reference_file_directory
+    print("reference_file_directory: ", reference_file_directory)
+
+    # approach to get temporary testing path
+    print(
+        "tmp_path: ", tmp_path
+    )  # pytest automatically keeps the last three runs for debugging and deletes everything that's older
```
isteinbrecher, line 53: Do you envision using this instead of the "testing-tmp" directory? In case CI tests fail we currently upload the `testing-tmp` directory to GitHub, to ease debugging. I would like to remain that possibility.

isteinbrecher (commented):
@davidrudlstorfer thanks a lot for the effort! This looks really great and I really like it. I left some comments. Do you envision switching only a few modules in this PR or more? See my previous comment, we could simply merge the `pytest` config and already execute the "old" tests and the port file by file, but that is just a suggestion.

pyproject.toml
davidrudlstorfer: Good point - didn't think about that. For this PR I would leave it as is and we could adapt in within #152

pyproject.toml
davidrudlstorfer: Locally I only execute the pytest framework and there I have 38% coverage - that's why I set it that low.

Normally the `tests/` is not added as only the coverage of the source code is of intereset

pyproject.toml
davidrudlstorfer: i just differentiated the case style because I liked lowercase better within the decorators. And I thought within the CLI this approach looks better 😅 They do not have to be identical

tests/conftest.py
davidrudlstorfer: Thanks - I changed that

tests/conftest.py
davidrudlstorfer: Good point - I'll split the 4C and ArborX tests

We could simply add a flag if the base tests are executed? Then this could be simply switched on and off for the performance tests as well all other tests

tests/conftest.py
davidrudlstorfer: I think https://stackoverflow.com/questions/17726954/py-test-how-to-get-the-current-tests-name-from-the-setup-method sums it up really nicely

This will also work for parameterized tests (we said we probably want those) and also it works for test functions as well as test methods.

The argument requests is always available once pytest is available.

Note: We do not need to provide the argument request. We simply call it via `current_test_name`.

tests/test_dummy.py
davidrudlstorfer: I think the `tmp_path` approach is more viable as it is maintained through pytest and not our own implementation. 

For the CI we can also simply upload the test directory.

This also provides the benefit of even better management, e.g. see https://github.com/pytest-dev/pytest/issues/8141

We can then also only upload the data of the failed tests and trash the results of the successfull tests

davidrudlstorfer:
@isteinbrecher thanks for your first round of review of the current status

In this PR I only wanted to update the current pytest (cosserat curve) and out of curiosity I've ported one additional feature (rotations). The remaining parts should be done in another PR

The only thing missing here is the implementation of the comparison function. As we've already briefly discussed I want to implement one function which automatically detects if either a string or file is handed over and then automatically compares the files or the string. I'll add that soon

pyproject.toml
isteinbrecher: Sounds like a good plan!

pyproject.toml
isteinbrecher: That makes sense. In the CLI I also prefer `4C`.

tests/conftest.py
isteinbrecher: > Note: We do not need to provide the argument request. We simply call it via current_test_name.

Sorry, I was not aware of that, then this is perfectly fine the way it is.

tests/test_dummy.py
isteinbrecher: Sounds good!

isteinbrecher:
> The only thing missing here is the implementation of the comparison function. As we've already briefly discussed I want to implement one function which automatically detects if either a string or file is handed over and then automatically compares the files or the string. I'll add that soon

@davidrudlstorfer we already have that, the `compare_test_result` (which calls `compare_strings` under the hood) does exactly that. If I remember correctly, we were talking about unifying this for general text based files and vtk files (`compare_vtk`), although I am not sure if this does not introduce more complexity than simply having two functions for fundamental different file types.

tests/conftest.py
davidrudlstorfer: Added an additional flag `--exclude-standard-tests` and it works as expected

`pytest` executes all standard tests
`pytest --4C` executes all standard tests + 4C tests
`pytest --4C --exclude-standard-tests` executes only the 4C tests

4C can be swapped with all other flags or an arbitrary number of flags can be combined, i.e.,

`pytest --4C --ArborX --CubitPy --Performance --exclude-standard-tests` executes only the special tests, i.e. 4C, ArborX, CubitPy and Performance test

Really nice solution in my opinion! Thanks for your input @isteinbrecher

davidrudlstorfer:
@isteinbrecher this would be ready for a review from your side.

In the last commit I've now added my suggestion for the new comparison feature. In my opinion this has a few advantages to the old version:

- one interface for testing comparisons (strings, vtk based objects, ... => easily extendable to other types, i.e. csv, ...)
- the interface is now clearer to the user => input are strings and nothing magical happens within the function (i.e., the retrieval of test names)
- Elaborate error retrieval and processing of errors: The actual differences are now printed for each failing test. This drastically improves the debugging rather than the manual file write out (which was not consistent across tests)

I've tested every path of the implementation and everything works accordingly!

I would be very happy if we could converge on a final pytest testing framework with the current suggested changes.

davidrudlstorfer:
This would now also enable to port everything to the new testing framework - as well as the deletion of the coverage configs and the test utils

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+    if rtol is None and atol is None:
+        diff = list(
+            unified_diff(reference.splitlines(), result.splitlines(), lineterm="")
+        )
+        if diff:
+            print("Difference between reference and result:")
+            print("\n".join(list(diff)))
+            return False
+        else:
+            return True
```
davidrudlstorfer, line 260: As an example this now uses the standard python diff library to create a very nice diff between the strings rather than the old option to just write everything into a file and the user needs to manually compare it again.

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+
+    parser.addoption(
+        "--Performance",
+        action="store_true",
+        default=False,
+        help="Execute standard and performance tests.",
+    )
+
+    parser.addoption(
+        "--exclude-standard-tests",
```
isteinbrecher, line 82: Nice!

.github/actions/run_tests/action.yml
```diff
@@ -46,8 +46,5 @@ runs:
         pip list
         cd tests
         coverage run --rcfile=${{ inputs.coverage_config }} testing_main.py
-        coverage html
-        coverage report
-        coverage-badge -o htmlcov/coverage.svg
-        coverage run --rcfile=${{ inputs.coverage_config }} -m pytest pytest_testing_cosserat_curve.py
-        coverage report
+        cd ..
+        pytest
```
isteinbrecher, line 50: Should we then set the testing flags via an `inputs:` variable? Something like `pytest_flags`. That could then simply be set in the workflows. This would make the `require_*` and the `coverage_config` flags obsolete.

Does it make sense to do this setup already in this PR? I think yes, but I will leave that up to you.

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+            reference: The reference string or path to the reference file.
+            result: The result string or path to the result file.
+            rtol: The relative tolerance.
+            atol: The absolute tolerance.
+
+        Returns:
+            bool: true if comparison is successful, false otherwise
+        """
+
+        if type(reference) is not type(result):
```
isteinbrecher, line 212: This means we have to switch to `Path` objects to describe paths (at least in testing), correct? This is not used at all in the rest of MeshPy. Besides the type checking are there disadvantages of using a string here?

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+            raise TypeError("Reference and result must be of the same type.")
```
isteinbrecher, line 213: We have a lot of tests, where we pass a string that is created during testing and compare it to a string saved in a reference file. This would mean that we have to first manually read the file and save it to a string variable and pass that to the function. That is likely just one or two lines of code, but they will be repeated a lot. Why not allow file vs. string comparisons?

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+) -> bool:
+    """Compare if two strings are identical within a given tolerance. If no
+    tolerance is given strings are compared for equality.
+
+    Args:
+        reference: The reference string.
+        result: The result string.
+        rtol: The relative tolerance.
+        atol: The absolute tolerance.
+        string_splitter: With which string the strings are split.
```
isteinbrecher, line 244: I really like the "new" docstring format when using type hints, that we don't have to specify the type here.

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+        # Loop over each line in the file
+        for line_reference, line_result in zip(lines_reference, lines_result):
+            line_reference_splits = line_reference.strip().split(string_splitter)
+            line_result_splits = line_result.strip().split(string_splitter)
+
+            if len(line_reference_splits) != len(line_result_splits):
+                raise AssertionError(
+                    f"Number of items in reference and result line differ!\n"
+                    + f"Reference line: {line_reference}\n"
+                    + f"Result line:    {line_result}"
```
isteinbrecher, line 283: Really nice, that is super helpful!

tests/test_cosserat_curve.py
```diff
@@ -251,7 +262,9 @@ def test_cosserat_curve_mesh_warp_transform_boundary_conditions():
         ),
     )
 
-    # Compare with the reference result
-    compare_test_result_pytest(
-        mesh.get_string(check_nox=False, header=False), rtol=1e-8, atol=1e-8
+    reference_string = get_string(
+        reference_file_directory / (current_test_name + ".dat")
     )
+    result_string = mesh.get_string(check_nox=False, header=False)
```
isteinbrecher, line 268: Like I mentioned before, do we really want to do this? We would have lines that look like this in almost every test case, producing a lot of duplicated code. Maybe, if we don't want the `check_nox=False, header=False` arguments all the time, we can consider allowing an `InputFile` to be passed to `compare_results`.

tests/test_rotations.py
```diff
@@ -0,0 +1,338 @@
...
+    tan = [2.0, 3.0, -1.0]
+    rot_smallest = smallest_rotation(rot, tan)
+
+    rot_smallest_ref = [
+        0.853329730651268,
+        0.19771093216880734,
+        0.25192421451158936,
+        0.4114279380770031,
+    ]
+    np.testing.assert_array_less(
```
isteinbrecher, line 336: I know this doesn't come from you, but do we want to switch all numpy array comparisons to `allclose` and provide a global standard for `rtol` and `atol`? Does not have to be part of this PR.

tests/test_rotations.py
```diff
@@ -0,0 +1,338 @@
...
+        t2_rot = rot * [0, 1, 0]
+        np.testing.assert_array_less(
+            np.linalg.norm(t1 - t1_rot),
+            mpy.eps_quaternion,
+            "test_rotation_matrix: compare t1",
+        )
+        np.testing.assert_array_less(
+            np.linalg.norm(t2 - t2_rot),
+            mpy.eps_quaternion,
+            "test_rotation_matrix: compare t2",
```
isteinbrecher, line 261: General comment, I used to do this in a lot of cases, write a short text of what is compared here. I stoped doing so a while ago, simply because I couldn't see any benefit of doing so. I would be fine with removing those (unless for some reason they are super meaningful) from the testing framework. Does not have to be part of this PR.

isteinbrecher (commented):
@davidrudlstorfer thanks a lot for the work on the modernisation of the testing framework. I left some comments, maybe it makes sense to discuss them in person at some point. I would just like to get the functionality best fit for the type of testing done in MeshPy. Please don't get discourage by my comments, I really like this whole approach and greatly appreciate your effort!

Other question, if I run `pytest` it takes a couple of seconds to `gather` the tests. Do you have an idea how this could be speed up. This doesn't matter for CI, only for local testing.

.github/actions/run_tests/action.yml
davidrudlstorfer: One thing I have not yet considered is the following:

We have to adjust for the coverage in different pipeline runs. If the performance tests should not include the standard tests the coverage will be lower. Therefore, we somehow have to adjust the necessary coverage for differing runs.

In my opinion we could also just execute the standard tests in every config so the minimum coverage is reached.

tests/conftest.py
davidrudlstorfer: Not really - in my opinion it would make sense to switch to path objects if we are at it. But let's discuss this in person


I'll just leave the first few search hints for "python string vs path" here.

https://medium.com/@guandika8/stop-representing-file-paths-in-python-using-strings-8312791f10f

https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

https://www.btelligent.com/en/blog/best-practice-working-with-paths-in-python-part-2-1/

tests/test_rotations.py
davidrudlstorfer: Definitely - I did not even know that this option exists in numpy before. 

`allclose` makes more sense

davidrudlstorfer:
@isteinbrecher big thanks for your second round of review! Let's discuss the few open points in person to faster converge on one solution :)

> Other question, if I run pytest it takes a couple of seconds to gather the tests. Do you have an idea how this could be speed up. This doesn't matter for CI, only for local testing.

I think this is a combination of many things here. First of all it takes so long because `pytest` resolves all imports prior to running the tests. This takes some time in MeshPy because the whole module structure is not optimal. Once the imports are improved and modules are correctly split this should be faster (i.e. see #159)

https://stackoverflow.com/questions/16417546/how-to-speed-up-pytest

> Pytest imports all modules in the testpaths directories to look for tests. The import itself can be slow. This is the same startup time you'd experience if you ran those tests directly, however, since it imports all of the files it will be a lot longer. It's kind of a worst-case scenario.
> 
> This doesn't add time to the whole test run though, as it would need to import those files anyway to execute the tests.

The total test time should, therefore, stay the same in comparison to the current testing approach. The current testing suite takes around 3s on my machine and the current pytests take around 1s for collecting and running. I think once everything is ported we can look at the total execution time again. Also after restructuring the repo and improving the imports this should be even faster.

tests/test_cosserat_curve.py
davidrudlstorfer: As discussed in person we further use the pytest fixtures

tests/test_cosserat_curve.py
```diff
@@ -221,24 +225,31 @@ def test_cosserat_curve_mesh_warp():
...
+    assert compare_results(reference_string, result_string, rtol=1e-10)
 
 
-def test_cosserat_curve_mesh_warp_transform_boundary_conditions():
+def test_cosserat_curve_mesh_warp_transform_boundary_conditions(
+    reference_file_directory,
+    get_string,
+    current_test_name,
+    compare_results,
+):
```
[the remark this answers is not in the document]
davidrudlstorfer: As discussed in person we want to use the pytest approach

.github/actions/run_tests/action.yml
davidrudlstorfer: I've now adjusted the entire testing workflow on Github. Therefore, I removed all `require_*` flags and replaced them with the new pytest flags/markers

I've also simplified the performance tests and removed the separate workflow action

It should be noted that in the current state only the pytest tests are executed and the "old" manual testing framework is no longer executed

I would stay with this approach as in my opinion it is not straight forward to keep both testing strategies and we should move the old tests ASAP to the new approach once this is merged

davidrudlstorfer: Additional Note: As discussed further above the base tests are always executed except for the performance tests => in this case only the performance tests are executed

tests/test_rotations.py
davidrudlstorfer: I removed the comments which I was able to find quickly

tests/performance_testing.py
```diff
@@ -173,8 +172,7 @@ class TestPerformance(object):
     """A class to test meshpy performance."""
 
     # Set expected test times.
-    expected_times = {}
-    expected_times["github-sisyphos-docker"] = {
```
isteinbrecher, line 177: Good decision to remove this hard coded host name!

Can you then also remove the one reference to this in the `README`?

```
(the reference time values and host name might have to be adapted in the file `<path_to_meshpy>/tests/performance_testing.py`)
```
replace with
```
(the reference time values might not suite your system)
```


Lets not put too much thought into that, as the README will likely be overhauled soon anyway.

tests/test_cosserat_curve.py
```diff
@@ -113,7 +107,7 @@ def test_cosserat_curve_translate_and_rotate():
     def load_compare(name):
         """Load the compare files and return a numpy array."""
         return np.loadtxt(
-            os.path.join(testing_input, f"{get_pytest_test_name()}_{name}.txt")
+            os.path.join(reference_file_directory, f"{current_test_name}_{name}.txt")
```
isteinbrecher, line 110: Could this maybe be the first case where we use the previously mentioned "get_reference_file_name" function?

tests/test_cosserat_curve.py
```diff
@@ -127,25 +121,32 @@ def load_compare(name):
...
 
-    result_name = os.path.join(testing_temp, get_pytest_test_name() + ".vtu")
-    curve = load_cosserat_curve_from_file()
-    pv.UnstructuredGrid(curve.get_pyvista_polyline()).save(result_name)
-    compare_vtk_pytest(
-        os.path.join(testing_input, get_pytest_test_name() + ".vtu"),
-        result_name,
+    reference_path = Path(
+        os.path.join(reference_file_directory, current_test_name + ".vtu")
+    )
```
isteinbrecher, line 131: See above. And also later on ;)

Basically every time we use `current_test_name` at the moment.

tests/test_cosserat_curve.py
```diff
@@ -127,25 +121,32 @@ def load_compare(name):
     )
 
 
-def test_cosserat_curve_vtk_representation():
+def test_cosserat_curve_vtk_representation(
+    tmp_path, reference_file_directory, current_test_name, compare_results
```
isteinbrecher, line 125: The `tmp_path` is provided by pytest, correct? Can we make sure that the contents of this directory get uploaded to GitHub in case of test failure? Does not have to be part of this PR, just to mention it for future reference.

tests/test_dummy.py
```diff
@@ -0,0 +1,88 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# -----------------------------------------------------------------------------
+"""Dummy test to demonstrate and ensure working pytest infrastructure."""
```
isteinbrecher, line 31: Will we delete this once all tests are ported?

tests/utils.py
```diff
@@ -38,65 +38,6 @@
 import vtk
```
isteinbrecher, line 38: Do we still need this file once everything is ported? Would it make sense to move everything here to the `conftest.py` file?

isteinbrecher (commented):
@davidrudlstorfer thanks for the changes, I am not sure if the PR was already ready for review, I sill left some comments 😅.

davidrudlstorfer:
@isteinbrecher sorry if it wasn't clear from my side: I am still working on the changes we discussed yesterday I hope to get them ready this evening or tomorrow morning

They will adress most of the points raised above, i.e., a very fancy way vastly simplify the result comparison with reference files

Currently its a real WIP - once the major things are resolved I'll push it 🚀

tests/conftest.py
davidrudlstorfer: I've added back the option regarding the VSCode diff and extended it to even more usecases

I also added back the option to download the artifacts from Github

More on that in my summary later down below

tests/test_cosserat_curve.py
davidrudlstorfer: I've entirely removed that logic, now we can even put in the mesh and everything is just one line.

More on that later down below in my summary.

tests/test_rotations.py
davidrudlstorfer: I removed a few occurences within this PR and would do the rest on a later occasion

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+            selected_tests.append(item)
+
+        if config.getoption("--ArborX") and "arborx" in markers:
+            selected_tests.append(item)
+
+        if config.getoption("--CubitPy") and "cubitpy" in markers:
+            selected_tests.append(item)
+
+        if config.getoption("--performance-tests") and "performance" in markers:
+            selected_tests.append(item)
```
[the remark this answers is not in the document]
davidrudlstorfer: Really nice suggestion 🙏🏻 applied it in the latest commit

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+@pytest.fixture(scope="session")
+def reference_file_directory() -> Path:
+    """Provide the path to the reference file directory.
+
+    Returns:
+        Path: A Path object representing the full path to the reference file directory.
+    """
+
+    testing_path = os.path.abspath(os.path.dirname(__file__))
+    return Path(testing_path) / "reference-files"
```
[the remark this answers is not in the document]
davidrudlstorfer: Nice thanks for improving the Paths :)

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+
+    compare = compare_grids(
+        get_vtk(reference), get_vtk(result), output=True, rtol=rtol, atol=atol
+    )
+
+    if not compare[0]:
+        raise AssertionError("\n".join(compare[1]))
+
+    return compare[0]
```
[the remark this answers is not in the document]
davidrudlstorfer: This is added within this PR and I will write more on that later down below

tests/performance_testing.py
davidrudlstorfer: Thanks for the hint regarding the README. I totally forgot about that.

I've updated the readme with basic infos regarding the new testing framework - as you say this probably needs an update soon.

isteinbrecher:
I am sorry, I didn’t have a close enough look before reviewing 😅

tests/test_cosserat_curve.py
davidrudlstorfer: I've updated the currently ported tests to the new function which does everything automatically. More on that later down below

tests/test_cosserat_curve.py
davidrudlstorfer: Yes that's also added - more on that later down below (😅 sorry for the repetition)

tests/test_dummy.py
davidrudlstorfer: Yes we can delete that file - I've just created it so all the things are working properly and that the test selection works properly because no other tests are tagged yet.

This was also a little playground to see what is possible and that others (mainly you) can see how things should work out

tests/utils.py
davidrudlstorfer: Yes - all the functionality is 100% ported and we can delete that file as well.

tests/conftest.py
```diff
@@ -0,0 +1,350 @@
...
+    """Return function to compare either string or files.
+
+    Necessary to enable the function call through pytest fixtures.
+    """
+
+    def _compare(
+        reference: Union[Path, str],
+        result: Union[Path, str],
+        rtol: float = 1e-14,
+        atol: float = 1e-14,
```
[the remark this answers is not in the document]
davidrudlstorfer: Very good catch!

I hope I fixed it correctly

davidrudlstorfer:
@isteinbrecher First of all - sorry that this took so long from my side - somehow I wanted to make it really fancy and with the current status of this PR at least my own expectations are met 😅 Even today all the changes took much longer than anticipated to include all the functionality as universally applicable as possible.

I am very curious about your opinion 🫣

1. All points we discussed yesterday in person are now adressed
2. I've addressed all conversations above

I'll quickly summarize the most important features of this PR

1. Tagging/Marking
We can now tag/mark any test we want like
```python
@pytest.mark.fourc
test_4C():
    pass
```
and execute them with
```bash
pytest --4C
```

2. I provided the following easy to use fixtures
    - `reference_file_directory` => returns the path to the reference file directory
    - `current_test_name` => returns the string of the current test name
    - `get_string` => returns the string of any given file
    - `get_corresponding_reference_file_path` => get the path to the corresponding dat file name of the current test
    - `compare_results` => this is the most important and somehow my holy function. With this you can compare anything we (currently) need, i.e., strings + strings, file + string, vtk+vtk, **file + input_file**
    This results now in these simple asserts for each test case:
    ```python
        assert compare_results(
        get_corresponding_reference_file_path, mesh, rtol=1e-8, atol=1e-8
    )
    ```
    Nothing more is needed! `get_corresponding_reference_file_path` automatically provides the path to the identically named .dat file and the string is automatically read from the mesh. In my opinion this is now an easily understandable code part which also does not need many lines.
    
 3. I've switched the testing framework within the github actions to pytest
 
 4. I've updated the README (with very little effort 😅)
 
 5. If a test fails it now looks like this. For each test the following output is written to the terminal (locally and within the pipeline):
 ```bash
 ----------------------------- Captured stdout call -----------------------------
========= Test: 'test_cosserat_curve_mesh_warp_transform_boundary_conditions' =========
String comparison with tolerance failed!
Numbers do not match within given tolerance!
Reference line: MAT 1 MAT_BeamReissnerElastHyper YOUNG -1.0 POISSONRATIO 0.5 DENS 0.0 CROSSAREA 0.007853981633974483 SHEARCORR 1 MOMINPOL 9.817477042468105e-06 MOMIN2 4.9087385212340526e-06 MOMIN3 4.9087385212340526e-06
Result line:    MAT 1 MAT_BeamReissnerElastHyper YOUNG -1.0 POISSONRATIO 0.0 DENS 0.0 CROSSAREA 0.007853981633974483 SHEARCORR 1 MOMINPOL 9.817477042468105e-06 MOMIN2 4.9087385212340526e-06 MOMIN3 4.9087385212340526e-06
Result string saved to: '/tmp/pytest-of-runner/pytest-0/test_cosserat_curve_mesh_warp_0/test_cosserat_curve_mesh_warp_transform_boundary_conditions_result.txt'.
```
Further, within the local IDE (VSCode) the same diff is opened as before. And on Github the temporary testing directory is uploaded (and the new result files can be downloaded) (for example see this run https://github.com/imcs-compsim/meshpy/actions/runs/12677922314/job/35334350367)

6. If I did not forget anything the only thing we have to adjust in the future is the necessary coverage once we have added more tests. We also have to differentiate between the testing machines and their coverage (i.e., the performance test run only executes the performance tests and not the standard tests => therefore, the coverage is lower)

7. And once everything is added we can also delete the old test utils, the dummy test and the dummy test reference file

Concluding - I think this now provides a good base for the complete switch to pytest. Thanks for the time you invested in all the review rounds and into our in-person meeting 🙏🏻

I'm really curious what you have to say 😄

**Comment by davidrudlstorfer** [comment]
I've ported one additional small file and the porting works really really nice

File: tests/conftest.py [hunk 31]
```diff
@@ -0,0 +1,512 @@
...
+        Function to compare results.
+    """
+
+    def _compare(
+        reference: Union[Path, str],
+        result: Union[Path, str, InputFile],
+        rtol: Optional[float] = None,
+        atol: Optional[float] = None,
+        input_file_string_check_nox: bool = False,
+        input_file_string_check_header: bool = False,
```
isteinbrecher, line 237: These two parameters are very specific, I am not sure if they make sense here. Maybe we can control the check via the `mpy` object. Alternatively we could "bundle" those arguments in a dictionary here, e.g.,
```python
    def _compare(
        reference: Union[Path, str],
        result: Union[Path, str, InputFile],
        rtol: Optional[float] = None,
        atol: Optional[float] = None,
        input_file_kwargs: dir={
            "input_file_string_check_nox": False,
            "input_file_string_check_header": False,
       },
        **kwargs,
    )
```
and then pass `input_file_kwargs` to `result.get_string(**input_file_kwargs)`.

tests/conftest.py
```diff
@@ -0,0 +1,512 @@
...
+            elif reference.suffix in [".vtk", ".vtu"]:
+                return compare_vtk_files(reference, result, rtol, atol)
+            else:
+                raise NotImplementedError(
+                    f"Comparison is not yet implemented for {reference.suffix} files."
+                )
+        else:
+            raise TypeError("Reference and result must be either string or Path.")
+
+    return _compare
```
isteinbrecher, line 303: I really like this approach of having one clearly structured function to do all the comparisons!

One thought, in most cases we end up actually comparing strings to each other. Should we consider a slightly different approach that will allow for an easier extension to other types of input (see my previous comment regarding the comparison between two meshes).

Something like (quickly copied together pseudo code)
```python
if ref_is_file and res_is_file:
  # Here we implement true file comparisons
  if ref_type is "vtk/vtu" and res_type is "vtu/vtk:
     return compare_vtk_files(reference, result, rtol, atol)
  else:
     raise error("Comparison for the file types "xxx" is not implemented")
else:
    # Here we have string comparisons
    def get_string2(data):
        """Could also be done with the previously defined get_string but also locally here"""
        if data_type is str:
            return data
        elif data_type is InputFile:
            return data.get_string(**some_args)
        else:
            return get_string(data)

     reference_string, result_string = [get_string2(data) for data in [reference, result])
     try:
          return compare_strings(
          reference_string, result_string, rtol, atol, **kwargs
    )
     except AssertionError as error:
          # Do we want to always call this function to improve the console output, or just in case that reference is a file?
          return handle_unequal_strings(
               error, tmp_path, current_test_name, result_string, reference
          )
        
```

tests/conftest.py
```diff
@@ -0,0 +1,512 @@
...
+        vtk.vtkDataObject: VTK data object.
+    """
+
+    reader = vtk.vtkXMLGenericDataObjectReader()
+    reader.SetFileName(path)
+    reader.Update()
+    return reader.GetOutput()
+
+
+def handle_unequal_strings(
```
isteinbrecher, line 473: Nice that this functionality is retained! Is this also the function that is responsible that the "failed" results get written to disc and then uploaded to GitHub?

isteinbrecher (commented):
Thanks for the effort @davidrudlstorfer! I really like the result. I left some comments.

.github/workflows/testing.yml
```diff
@@ -47,16 +47,12 @@ jobs:
...
-          require_four_c: 0
-          require_arborx: 0
-          require_cubitpy: 0
       - name: Upload test results on failure
         if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: ${{github.job}}-${{ matrix.os-version }}-python${{ matrix.python-version }}-${{github.run_number}}
-          path: ${{github.workspace}}/tests/testing-tmp/
+          path: /tmp/pytest-of-runner/
```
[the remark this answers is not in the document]
davidrudlstorfer: Good catch :eyes: 

Yesterday I read some docs that the path is identical for all Github hosted runners but as one can see here https://github.com/imcs-compsim/meshpy/actions/runs/12677922314 the path actually differs on Mac and Windows

I'll fix it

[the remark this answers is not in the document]
davidrudlstorfer: Now the upload works for all systems - in the pipeline i simply use the provided tmp directory by Github and upload it in the final step

See https://github.com/imcs-compsim/meshpy/actions/runs/12690733332 for all uploaded artifacts

tests/conftest.py
```diff
@@ -0,0 +1,512 @@
...
+
+        corresponding_reference_file_path = reference_file_directory / (
+            current_test_name + ".dat"
+        )
+
+        assert os.path.isfile(corresponding_reference_file_path)
+
+        return corresponding_reference_file_path
+
+    return _get_corresponding_reference_file_path()
```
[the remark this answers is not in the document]
davidrudlstorfer: I've added the old logic to enable editing the corresponding reference file path

tests/conftest.py
davidrudlstorfer: Yes - exactly. The files are written to the tmp directory and the upload just uses the temp directory in the Github Actions

tests/conftest.py
davidrudlstorfer: I really like the approach with the bundled arguments in a dictionary - I've added it in the latest commit

tests/conftest.py
davidrudlstorfer: Thanks for the nice suggestion - this has also minimized duplicated code. I've added the new version in the last commit

Regarding the question in the code - we only use the handle unequal strings function if the reference is a file, because otherwise we could not make a diff in Vscode etc.

In the string-string comparisons the same error is printed to the CLI with the Assertions which are executed - so the output is more or less the same - except for the diff

davidrudlstorfer:
Of course, it took much longer than expected - it felt like 50 pipeline runs :smiling_face_with_tear: 

Now everything works and the artifact upload works for all containers (macOS, Windows, ubuntu, 4C container)

I've adressed all of your comments and this would be ready for your next review @isteinbrecher

davidrudlstorfer:
This would be a failed pipeline run where you can download the artifact from each job

https://github.com/imcs-compsim/meshpy/actions/runs/12692452173

isteinbrecher:
@davidrudlstorfer thanks for the work, I think we are very close to merging this. I took the liberty and added two commits to the branch, as a suggestion:
```
b769dca Allow for general types in compare function
```
This commit allows for comparison of all possible combinations of `Path`, `str`, and `InputFile` with each other.

```
9766fde Refactor assertions and return values from compare functions
```
While adding the changes form the previous commit, I realised that I missed the changes in the `compare_strings_with_tolerance` function which raise an error. This should not be done in this function, as we use this one in some other projects where we rely that the return value is either True or False. I changed that accordingly.

While looking at this, I saw that we had some (at least to me) not so intuitive behaviours in the code: We had functions that return True or raise an error, but never return False. At least to me that seemed weird. I tried to unify this, that the compare functions we actually will call from within the testing framework now either run without complaining or raise an assertion. This also allows to simply call the renamed compare function in the testing with
```python
assert_results_equal(...
```
Without the leading `assert` - I personally like that, since this eliminates the possible error that someone calls the compare function and forgets the `assert`.

I am curios about your opinion on these changes.

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+
+        assert os.path.isfile(corresponding_reference_file_path)
+
+        return corresponding_reference_file_path
+
+    return _get_corresponding_reference_file_path
+
+
+@pytest.fixture(scope="function")
+def assert_results_equal(get_string, tmp_path, current_test_name) -> Callable:
```
davidrudlstorfer, line 235: That's a good idea regarding the name change! Now it better reflects what it does and it makes sense to no longer return true or false.

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+    Args:
+        get_string: Function to get string from file.
+        tmp_path: Temporary path for testing.
+        current_test_name: Name of the current test.
+
+    Returns:
+        Function to compare results.
+    """
+
+    def _compare(
```
davidrudlstorfer, line 249: Sorry that I also made the error - here it would be nice if the inner function is named similarly/identical to the outer one, i.e.,

`_assert_results_equal`

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+        reference: Union[Path, str],
+        result: Union[Path, str, InputFile],
+        rtol: Optional[float] = None,
+        atol: Optional[float] = None,
+        input_file_kwargs: dict = {
+            "check_nox": False,
+            "header": False,
```
davidrudlstorfer, line 256: That's a really nice idea! Improves the readability by a lot

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+        },
+        **kwargs,
+    ):
```
davidrudlstorfer, line 259:
```suggestion
    ) -> None:
```

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+                    return get_string(data)
+                elif isinstance(data, InputFile):
+                    return data.get_string(**input_file_kwargs)
+                else:
+                    raise TypeError(f"Data type {type(data)} not implemented.")
+
+            # retrieve strings to compare
+            [reference_string, result_string] = [
+                get_string_to_compare(data) for data in [reference, result]
+            ]
```
davidrudlstorfer, line 305: Currently I do not like that construct that much. First of all we could completely remove the inner by simply generalizing the `get_string` function. For me the main problem is that on the bottom we retrieve the strings but the input could be a string as well. So we have function calls to convert a string into a string. This entire construct extremely increases the complexity of what is actually happening and I currently see no added benefit.

If you like we can keep it for sure as this is only my opinion

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+            else:
+                raise NotImplementedError(
+                    f"Comparison is not yet implemented for {reference.suffix} files."
+                )
+
+        # String based comparison
+        else:
+
+            def get_string_to_compare(data):
+                """Get the string depending on the data type that will actually
```
davidrudlstorfer, line 291: Here we should also add the type hinting and the current style of docstring with args and return if we decide on keeping it.

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+            [reference_string, result_string] = [
+                get_string_to_compare(data) for data in [reference, result]
+            ]
+
+            # compare strings and handle non-matching strings
+            compare, error_message = compare_strings(
+                reference_string, result_string, rtol, atol, **kwargs
+            )
+            if not compare:
+                if isinstance(reference, Path):
```
davidrudlstorfer, line 312: With the try, except and split into the pure string comparison this nested if construction was not necessary (in the beginning I had the same code but I removed it 😅)

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+    return _compare
+
+
+def compare_strings(
+    reference: str,
+    result: str,
+    rtol: Optional[float] = None,
+    atol: Optional[float] = None,
+    string_splitter: str = " ",
+) -> Tuple[bool, str]:
```
davidrudlstorfer, line 331: This weird return type is what I wanted to prevent - let's just raise an assertion error and catch it above? Then we do not need to return anything in this function as well and can simply reuse it

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+        bool, str: True if comparison is successful, False otherwise. Also return
+        string containing information about failed comparisons.
+    """
+
+    if rtol is None and atol is None:
+        return compare_strings_equality(reference, result)
+    else:
+        return compare_strings_with_tolerance(
+            reference, result, rtol, atol, string_splitter=string_splitter, output=True
+        )  # type: ignore
```
davidrudlstorfer, line 352: Why do we need this ignore?

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+        option is set, also return string containing information about failed
+        comparisons.
+    """
+
+    def get_return_values(flag, message):
+        """Get the data structure that shall be returned from this function."""
+        if output:
+            return flag, message
+        else:
+            return flag
```
davidrudlstorfer, line 407: This entire construct could also be simply removed if we just raise the Assertion Error and catch it above - for me that's a lot of added complexity

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+
+def compare_strings_with_tolerance(
+    reference: str,
+    result: str,
+    rtol: Optional[float],
+    atol: Optional[float],
+    *,
+    string_splitter=" ",
+    output=False,
+) -> Union[bool, Tuple[bool, str]]:
```
davidrudlstorfer, line 385: I think the function should have a fixed return value that would simplify a lot.

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+    return reader.GetOutput()
+
+
+def handle_unequal_strings(
+    error: str,
+    tmp_path: Path,
+    current_test_name: str,
+    result: str,
+    reference_path: Path,
+):
```
davidrudlstorfer, line 509:
```suggestion
) -> None:
```

tests/conftest.py
```diff
@@ -0,0 +1,537 @@
...
+                    raise TypeError(f"Data type {type(data)} not implemented.")
+
+            # retrieve strings to compare
+            [reference_string, result_string] = [
+                get_string_to_compare(data) for data in [reference, result]
+            ]
+
+            # compare strings and handle non-matching strings
+            compare, error_message = compare_strings(
+                reference_string, result_string, rtol, atol, **kwargs
```
davidrudlstorfer, line 309: That's also one point I explicitly wanted to prevent. In my opinion it's not really nice to return a tuple only in the case if it fails. If it succeeds it just returns an empty string which is completely unused.

My thought behind catching the assertion was that the function can also be used entirely on it's own. Now it's only usable in combination with the code down below.

davidrudlstorfer (commented):
@isteinbrecher thanks a lot for extending the functions!

Unfortunately, I do not fully agree with you on all changes. In many cases a lot of complexity and nested functions were added where it is not necessary at all. Also functions should always return one thing and not just empty strings depending on the outcome.

We could probably increase the simplicity of the code by a lot. I also fully agree with you if you decide to merge the current changes.

davidrudlstorfer:
If you like I am more than happy to also provide a suggestion on my comments from above - I think in the end it is worth it if we provide a really nice testing framework 🙏🏻

tests/conftest.py
isteinbrecher: Maybe I can give a little context here. I view this function as a utility that allows to compare strings, where floating point values (or integers) can be compared using numerical tolerances. It is also used in some other projects, e.g., CubitPy. I would like to not throw anything in this function (unless an actually error occurs), because in my opinion the user of this function is responsible to handle the result of the comparison. Now I really like your additional error messages, so I choose an approach very similar to the one in `compare_grids` for `vtk` comparisons (on that note, maybe we also want to retrieve the error message from `compare_grids` and print it for failed tests). If I remember correctly, I saw this approach with the optional information in the return value in some `numpy` functions.

This might not be the best way, but for the pure compare functions I would like to keep that approach (unless there is some nice alternative without assertions).

We can basically decide how to handle the rest in MeshPy and there I am completely open. I am fine with going back to your previous approach with the assertions, then we basically just have to check the return value from `compare_strings_with_tolerance` inside `compare_strings` and throw the respective assertion. Then everything else could go back to the way you had it (maybe without the `return True` in the functions to simply keep them `void` and only throw an assertion in case of a failed comparison).

tests/conftest.py
isteinbrecher: See my previous comment

tests/conftest.py
isteinbrecher: Yes, lets go back to your approach in this function, see my comment below.

tests/conftest.py
isteinbrecher: Yes, lets go back to your approach in this function, see my comment below.

tests/conftest.py
isteinbrecher: Yes, lets go back to your approach in this function, see my comment below.

isteinbrecher (commented):
Thanks for your comments @davidrudlstorfer. I left some explanations. I will add another commit trying to revert my changes to address most of your major points. We can use that as a starting point for further discussion or simply revert it again.

tests/conftest.py
```diff
@@ -0,0 +1,519 @@
...
+    Args:
+        error: Raised error during string comparison
+        tmp_path: Temporary pytest directory
+        current_test_name: Name of the current test
+        result: "New" result string
+        reference_path: Path to "old" reference file
+    """
+
+    print(f"========= Test: '{current_test_name}' =========")
+    print(error)
```
isteinbrecher, line 505: Do we really want to print the error here? Shouldn't that be done by `pytest` itself?

davidrudlstorfer: Yes - we can definetily remove these two lines - I'll remove it

tests/conftest.py
davidrudlstorfer: I am sold - let's keep the now implemented more general way. Big thanks for going the extra mile and improving everything - especially by expanding the `get_string` function

davidrudlstorfer:
@isteinbrecher thanks for adding the changes and explanations!

If I understand you correctly your major concern is that you want to use the function externally as well. What do you think if we just provide a wrapper for the external use?

I added this in the last commit - this function now does exactly the same thing you implemented but is just a wrapper around the standard `compare_strings_with_tolerance` function.

This way the function signature is identical within meshpy for `compare_strings_equality` and `compare_strings_with_tolerance`.

I am open for a better name of the function which also returns stuff 😅

Curious about your opinion

tests/conftest.py
```diff
@@ -0,0 +1,534 @@
...
+        # String based comparison
+        else:
+            # retrieve strings to compare
+            [reference_string, result_string] = [
+                get_string(data, input_file_kwargs) for data in [reference, result]
+            ]
+
+            # compare strings and handle non-matching strings
+            try:
+                return compare_strings(
```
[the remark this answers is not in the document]
davidrudlstorfer: Done - good catch

davidrudlstorfer:
Perfect 🤩 I've addressed the 2 minor remarks. I'll squash everything into a few manageable commits and then force push it

isteinbrecher (approved):
Thanks again for the effort, lets get this merged!

PR #158, opened 2024-12-17
Add missing docstrings

isteinbrecher:
This PR adds all missing dock strings and sets the required doctoring coverage to 100%.

@davidrudlstorfer I added temporary "TODO" dock strings to a lot of different testing functions that will likely be replaced with `pytest` fixtures in the near future.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
     hooks:
       - id: interrogate
         name: Run interrogate (linter for docstrings)
-        args: [--fail-under=95, --ignore-init-module, --style=google, -vv, meshpy/]
+        args: [--fail-under=100, --ignore-init-module, --style=google, -vv]
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.13.0
     hooks:
--- a/meshpy/conf.py
+++ b/meshpy/conf.py
@@ -152,6 +152,7 @@ class MeshPy(object):
         self.vtk_type = VTKType
 
     def set_default_values(self):
+        """Set the configuration to the default values."""
         # Version information.
         self.git_sha = None
         self.git_date = None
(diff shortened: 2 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -58,7 +58,7 @@ repos:
     hooks:
       - id: interrogate
         name: Run interrogate (linter for docstrings)
-        args: [--fail-under=95, --ignore-init-module, --style=google, -vv, meshpy/]
+        args: [--fail-under=100, --ignore-init-module, --style=google, -vv, meshpy/]
```
davidrudlstorfer, line 61: If we want to also check the docstring coverage within the tests we should probably also add the path here?

See https://github.com/davidrudlstorfer/pygebege/blob/ffcb1e370bb087f2bf189bce0e376a7a77e00fdb/.pre-commit-config.yaml#L23

davidrudlstorfer (approved):
Very nice! Thanks for tackling this so fast! I only have a minor comment

.pre-commit-config.yaml
isteinbrecher: I think that is a private repository, the link does not work for me.

davidrudlstorfer: Sorry my fault - https://github.com/davidrudlstorfer/pyskel/blob/a929e2b8f849fb84d19d3a7b9dbd7e575a72d446/.pre-commit-config.yaml#L48

This should work

```suggestion
        args: [--fail-under=100, --ignore-init-module, --style=google, -vv, meshpy/, tests/]
```

Basically that was my idea. I think we also need the path to the test folder if the docstring coverage should be checked?

davidrudlstorfer: Sounds good - if it works out of the box let's drop the path arguments 🤝

isteinbrecher: I tried, it works, a wrong doctoring cause the pipeline to fail. The current state should be good to go.

davidrudlstorfer (approved):

.pre-commit-config.yaml
davidrudlstorfer: Thanks for testing the missing docstring 🙏🏻

PR #160, opened 2024-12-19
Return geometry objects from geometry_set

knarfnitram:
Adds a function to return the geometry object of the geometry set and adds an example how it can be used to add specific elements from a line to a condition.

### DIFF
--- a/meshpy/geometry_set.py
+++ b/meshpy/geometry_set.py
@@ -246,6 +246,10 @@ class GeometrySet(GeometrySetBase):
                 "Currently GeometrySet are only implemented for points and lines"
             )
 
+    def get_geometry_objects(self):
+        """Return a list of the objects with the specified geometry type."""
+        return list(self.geometry_objects[self.geometry_type].keys())
+
 
 class GeometrySetNodes(GeometrySetBase):
     """Geometry set which is defined by nodes and not explicit geometry."""
--- a/tests/testing_meshpy.py
+++ b/tests/testing_meshpy.py
@@ -966,6 +966,32 @@ class TestMeshpy(unittest.TestCase):
             additional_identifier=additional_identifier,
         )
 
+    def test_geometry_set_get_geometry_objects(self):
+        """Test if the geometry set returns the objects(elements) in the
+        correct order."""
+
+        # Initialize material and input file.
+        mat = MaterialReissner()
+        mesh = InputFile()
+
+        # number of elements
+        n_el = 5
+
+        # Create a simple beam.
+        geometry = create_beam_mesh_line(
+            mesh, Beam3rHerm2Line3, mat, [0, 0, 0], [2, 0, 0], n_el=n_el
+        )
+
+        # Get all elements from the geometry set.
+        elements_of_geometry = geometry["line"].get_geometry_objects()
+
+        # Check number of elements.
+        self.assertEqual(len(elements_of_geometry), n_el)
+
+        # Check if the order of the elements from the geometry set is the same as for the mesh.
+        for i_element, element in enumerate(elements_of_geometry):
+            self.assertEqual(element, mesh.elements[i_element])
+
     def test_meshpy_replace_nodes_geometry_set(self):
         """Test case for coupling of nodes, and reusing the identical nodes.
 


tests/testing_meshpy.py
```diff
@@ -966,6 +966,43 @@ def get_arguments_curve(circle_type):
             additional_identifier=additional_identifier,
         )
 
+    def test_get_geometry_set_to_boundary_conditions(self):
+        """Test if the geometry set returns the objects(elements) and if they
+        can be forwarded to a GeometrySet."""
+
+        # Initialize material and input file.
+        mat = MaterialReissner(radius=0.1, youngs_modulus=1)
```
isteinbrecher, line 974: For a test case the radius and Youngs modulus can be left empty.

meshpy/geometry_set.py
```diff
@@ -246,6 +246,10 @@ def get_all_nodes(self):
                 "Currently GeometrySet are only implemented for points and lines"
             )
 
+    def get_geometry_objects(self):
+        """Return the objects of the specified geometry type."""
+        return self.geometry_objects[self.geometry_type]
```
isteinbrecher, line 251: This returns a `dict`, because that is the way we store the data internally due to the ordered set issue discussed in https://github.com/imcs-compsim/meshpy/pull/138. But wouldn't it make more sense to return `list(self.geometry_objects[self.geometry_type].keys())` here? As a user I have no interest in the dictionary, only in a list containing the added elements in the correct order.

tests/testing_meshpy.py
```diff
@@ -966,6 +966,43 @@ def get_arguments_curve(circle_type):
...
+        # First check type.
+        self.assertEqual(type(elements_of_geometry[0]), Beam3rHerm2Line3)
+
+        # Check number of elements.
+        self.assertEqual(len(elements_of_geometry), n_el)
+
+        # Add boundary condition to all elements except the first and last element.
+        mesh.add(
+            BoundaryCondition(
+                GeometrySet(elements_of_geometry[1:-1]),
```
isteinbrecher, line 997: Why not add all elements? That would make the test case even more robust.

tests/testing_meshpy.py
```diff
@@ -966,6 +966,43 @@ def get_arguments_curve(circle_type):
             additional_identifier=additional_identifier,
         )
 
+    def test_get_geometry_set_to_boundary_conditions(self):
```
isteinbrecher, line 969: `test_get_geometry_objects`

meshpy/geometry_set.py
knarfnitram: Yes! This makes sense totally sense, especially since at every point I have converted them afterwards, any time.
Thanks for the hint!

tests/testing_meshpy.py
knarfnitram: I think we should exclude the first and last here:
If we add all elements, the question is, why do we test this at all ? We could simply add all elements by using the Geometry set from `beams["line"]`. 
So this test case is also here to show, how this list can be used, namely select elements as you need them.

tests/testing_meshpy.py
```diff
@@ -966,6 +966,43 @@ def get_arguments_curve(circle_type):
...
+        mesh.add(
+            BoundaryCondition(
+                GeometrySet(elements_of_geometry[1:-1]),
+                "NUMDOF 9 ONOFF 1 1 1 0 0 0 0 0 0 VAL 0 0 0 0 0 0 0 0 0 FUNCT 0 0 0 0 0 0 0 0 0",
+                bc_type=mpy.bc.dirichlet,
+            )
+        )
+
+        # Compare with the reference file.
+        compare_test_result(self, mesh.get_string(header=False))
```
knarfnitram, line 1004: That's why also the name is `test_get_geometry_set_to_boundary_conditions`
For testing only the function, `get_geometry_objects` I can remove the part completely here.

tests/testing_meshpy.py
isteinbrecher: Thinking of it, what you should actually do, is add a boundary condition for each element in `elements_of_geometry`. By doing so, you also check that the order is correct:
```python
for element in elements_of_geometry:
        mesh.add(
            BoundaryCondition(
                GeometrySet(element, ...))

```
If the order of the returned elements is correct, you also know that you can always just drop the first and the last one using slices.

isteinbrecher: Alternatively what you could do (without having to add a boundary conditions or a result file):
```python
for i_element, element in enumerate(elements_of_geometry):
    assert element == mesh.elements[i_element]
```
That would be the clearer approach -> you simply check that the order you get from the element set is the order of the elements in the mesh, which is ultimately what you are interested in and what you want to ensure.

tests/testing_meshpy.py
```diff
@@ -966,6 +966,43 @@ def get_arguments_curve(circle_type):
             additional_identifier=additional_identifier,
         )
 
+    def test_get_geometry_set_to_boundary_conditions(self):
+        """Test if the geometry set returns the objects(elements) and if they
+        can be forwarded to a GeometrySet."""
```
isteinbrecher, line 971:
```suggestion
        """Test if the geometry set returns the objects (elements) in the correct order."""
```

tests/testing_meshpy.py
isteinbrecher: With the approach from the previous comment, the naming should be even more fine if it is just `test_get_geometry_objects` or maybe even better `test_geometry_set_get_geometry_objects`. The naming of and ordering of the tests is something that we will have to tackle at some point anyways 😅

tests/testing_meshpy.py
knarfnitram: Ok, I have removed the file, so let's keep it simple and only test  if the order of the elements is same as for the mesh.

isteinbrecher (approved):

PR #161, opened 2024-12-19
Add 4C beam to beam contact 

knarfnitram:
This MR adds the boundary conditions for beam contact - header functions and example is moved to different MR.


meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,65 @@
...
+from ..conf import mpy
+from ..geometry_set import GeometrySet
+
+
+class BeamtoBeamContactCondition(BoundaryCondition):
+    """This object represents the beam to beam contact condition in 4C.
+
+    It sets the conditions to enable contact between beams.
+    """
+
```
knarfnitram, line 45: Here I have a question:
Currently, I have to add two times the same Boundary Condition.
Wouldn't it make more sense to have something like:
`BeamtoBeamContactCondition(nodes1,nodes2)` 
and under the hood the function create the boundary conditions for `nodes1` and `nodes2` and keeps track automatically of the IDs ?

meshpy/header_functions.py
```diff
@@ -540,3 +535,181 @@ def set_header_static(
 
     # Set the xml content in the input file.
     input_file.nox_xml = nox_xml
+
+
+def set_binning_strategy(input_file, *, binning_size=None, bounding_box=None, **kwargs):
+    """Set Binning options in input file.
```
knarfnitram, line 541: Binning small.

meshpy/four_c/beam_to_beam_contact.py
isteinbrecher: Per definition a boundary condition is always something that acts on a single geometry set in MeshPy. So it is not possible to add two geometry sets to a single BC. What is however possible is to have a factory function or object (similar to what you proposed) what takes `beam_contact_bc_factory(mesh, nodes1, nodes2, bc_description, ...)` and creates the corresponding BCs automatically. The ID count could either be stored in the factory object, a global variable, or be retrieved by checking how many `beam_to_beam_contact` conditions are already in the mesh and dividing that number by 2. @knarfnitram @davidrudlstorfer do you have an idea what would be a Pythonic way to track the number of added IDs? I am not sure if global variables and objects are desirable.

meshpy/conf.py
```diff
@@ -47,6 +47,7 @@ class BoundaryCondition(Enum):
     neumann = auto()
     locsys = auto()
     moment_euler_bernoulli = auto()
+    beam_to_beam_interaction = auto()
```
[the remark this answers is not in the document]
davidrudlstorfer: For beam potential interaction we do not need "boundary" conditions but just these conditions:

```
------------------------------------DESIGN LINE BEAM POTENTIAL CHARGE CONDITIONS
E 1 POTLAW 1 VAL 1.0 FUNCT 1
```

where we define which potential law is active on which design line

meshpy/header_functions.py
```diff
@@ -537,3 +532,181 @@ def set_header_static(
...
+        Input file that the options will be added to.
+    binning_size:
+        Lower bound for bin size. Exact bin size is computed via (Domain edgelength)/BIN_SIZE_LOWER_BOUND.
+    bounding_box:
+        Bounding box for computational domain using binning strategy.
+    """
+    if binning_size is None:
+        binning_size = kwargs.get("binning_size", binning_size)
+    if bounding_box is None:
+        bounding_box = kwargs.get("bounding_box", bounding_box)
```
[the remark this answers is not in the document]
knarfnitram: The problem that I am facing is that `set_beam_contact_section` calls  `set_binning_strategy` .
I don't want to forward all the information of the optional parameters from `set_binning_strategy` to `set_beam_contact_section` within the function, as this causes overhead of documentation, parameters and one would have to introduce additional checks as done for `binning_bounding_box` within the function `set_beam_to_solid_meshtying`. I am therefore using the keyword arguments to tunnel the arguments through `set_beam_contact_section` to `set_binning_strategy`.

knarfnitram:
@isteinbrecher Thank you very much for the comments. 
I have incoperated most of them. 
The boundary factory is now gone, we simply have a function, which estimates and adds the id. I tried to make this a little bit more generic, such that it maybe can be reused for other id-based conditions.

Also there is still one comment open regarding the usage of the keyword arguments. I hope, that this clearifys the usage furhter.

src/meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,157 @@
...
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
```
davidrudlstorfer, line 28: Here you have inserted the license twice

src/meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,157 @@
...
+# -----------------------------------------------------------------------------
+"""This file contains a function to add the beam-to-beam contact conditions for
+4C."""
+
+# MeshPy modules
```
davidrudlstorfer, line 61:
```suggestion
```

This should be removed as well

davidrudlstorfer (commented):
@knarfnitram I just quickly skimmed over the changes

src/meshpy/four_c/beam_to_beam_contact.py
knarfnitram: Thanks for hint!

tests/test_header_functions.py
```diff
@@ -180,3 +182,28 @@ def test_header_functions_stress_output(
...
     # Check the output.
     assert_results_equal(get_corresponding_reference_file_path(), input_file)
+
+
+def test_header_functions_beam_interaction(
+    get_corresponding_reference_file_path, assert_results_equal
+):
+    """Test the beam-to-beam contact header function with default parameter."""
+    # Set default values for global parameters.
+    mpy.set_default_values()
```
isteinbrecher, line 192: This is not needed, as it is done before each test case. (I will open a PR that fixes this in the rest of MeshPy)

isteinbrecher (commented):
Thanks a lot @knarfnitram for the new implementation - a lot cleaner than the old approach.

I left some comments (mainly for the boundary creation part) that should allow to use this proposed approach with all types ob beam interaction BC (also tagging @dharinib98 here).

Could you still split up the PR in the BC creation changes and the header changes? Or am I missing something there?

src/meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,127 @@
...
+
+from meshpy.core.conf import mpy
+from meshpy.core.geometry_set import GeometrySet
+from meshpy.core.mesh import Mesh
+from meshpy.four_c.boundary_condition import BoundaryCondition
+
+
+def get_next_possible_id_for_boundary_condition(
+    mesh: Mesh, bc_type: BoundaryCondition, search_string, group_idx=1
+):
```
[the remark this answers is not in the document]
knarfnitram: So the problem is, that in meshpy exists two classes with the name `BoundaryCondition`. 
One within from `meshpy.core.conf` and one within `meshpy.four_c.`

I agree that i forget to add the correct one from `core.conf`. Can we rename one of these classes ? Maybe the one from `meshpy.core.conf` to `BoundaryConditionEnum` ?

src/meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,127 @@
...
+    Returns:
+        id (int): smallest used id
+    """
+
+    found_conditions = []
+
+    # loop through every possible geometry and find the conditions
+    for geometry_type in mpy.geo:
+        for bc_condition in mesh.boundary_conditions[(bc_type, geometry_type)]:
+            found_conditions.append(bc_condition)
```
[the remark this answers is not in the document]
knarfnitram: I would leave it as it is. The worst thing that can happen is that, the estimated ID is too big, since it also checks  other geometry types as well. However, if you perform the check only for a certain geometry type, the same ID will be returned for different geometry types.

src/meshpy/four_c/beam_to_beam_contact.py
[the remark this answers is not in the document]
isteinbrecher: You are right, there are multiple `BoundaryCondition` classes, this should be resolved in the future. However, for you this is not an issue, as the `BoundaryCondition` in `conf` is never used directly but only via `mpy.bc`, so it should be sufficient to write

```suggestion
def get_next_possible_id_for_boundary_condition(
    mesh: Mesh, bc_type: mpy.bc, search_string: str, group_idx: int=1
):
```

src/meshpy/four_c/beam_to_beam_contact.py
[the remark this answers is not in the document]
isteinbrecher: I think in the standard case, we want continuous IDs for beam-to-xxx coupling conditions. Otherwise this whole implementation could simply be replaced by returning the total number of boundary conditions in the mesh (I really like the current approach of figuring out the actually needed ID).

src/meshpy/four_c/beam_to_beam_contact.py
```diff
@@ -0,0 +1,127 @@
...
+            assert match is not None
+            existing_ids.append(int(match.group(group_idx)))
+
+        # return lowest found id
+        return min(set(range(len(existing_ids) + 1)) - set(existing_ids))
+
+
+def add_contact_boundary_condition_to_mesh(
+    mesh: Mesh, contact_set_1: GeometrySet, contact_set_2=None, id=None
+):
```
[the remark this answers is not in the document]
knarfnitram: Is this renaming of the function here on purpose?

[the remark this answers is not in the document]
isteinbrecher: Yes, this functionality can be used for all types of beam interaction.

isteinbrecher (commented):
Thanks for splitting this up and making it easier to review @knarfnitram !

I added a commit with a few small changes and some renaming. Now your functionality can also be used to create beam-to-solid conditions with automatic numbering of the `COUPLING_ID`s. (I also added a test case for this)

I think this is a very useful new feature.

@dharinib98 you can have a look at the test case `test_four_c_beam_to_solid` maybe that can hep you in the future when you are defining boundary conditions.

@knarfnitram can you have a look at the additional changes?

tests/test_four_c.py
```diff
@@ -469,7 +477,87 @@ def test_meshpy_beam_to_beam_contact(
     )
 
     # Add the beam-to-beam contact condition.
-    add_beam_interaction_condition(mesh, beam_x["line"], beam_y["line"])
+    add_beam_interaction_condition(
+        mesh, beam_x["line"], beam_y["line"], mpy.bc.beam_to_beam_contact
```
knarfnitram, line 481: Should we rename `mpy.bc.beam_to_beam_contact` to '`mpy.bc.beam_to_beam_interaction`?

tests/test_four_c.py
```diff
@@ -376,3 +378,186 @@ def test_linear_time_transformation_flip():
...
+        mat,
+        [0, 0, 0.5],
+        [1, 0, 0.5],
+        n_el=2,
+    )
+
+    # Add the beam-to-beam contact condition.
+    add_beam_interaction_condition(
+        mesh, beam_x["line"], beam_y["line"], mpy.bc.beam_to_beam_contact
+    )
```
knarfnitram, line 482: @isteinbrecher,  I have now understood your vision for the `beam_interaction_condition`. Changes look good to me!

isteinbrecher: Great! Then lets merge this!

PR #166, opened 2025-01-08
Move coupling node replacement functionality out of Abaqus module

isteinbrecher:
In some finite element (FE) solvers, nodes coupled via joints are resolved by assigning a "master" node to represent the joint. This PR introduces a general function that identifies such nodes and creates a mapping to identify the "replaced" and "remaining" nodes.

### DIFF
--- a/meshpy/abaqus/input_file.py
+++ b/meshpy/abaqus/input_file.py
@@ -37,6 +37,7 @@ import numpy as np
 
 from ..conf import mpy
 from ..geometry_set import GeometrySet
+from ..inputfile_utils import get_coupled_nodes_to_master_map
 from ..mesh import Mesh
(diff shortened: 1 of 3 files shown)


davidrudlstorfer (dismissed):
Could we add a test for the `get_coupled_nodes_to_master_map` as there is a lot of logic happening inside this function? If there's no viable way to test this - let's get this merged :rocket:

isteinbrecher:
Good point, I added a test case for the new function. The test case is already suited for pytest, I did not add it to the testing framework, because it should be automatically picked up once #157 is merged.

davidrudlstorfer (approved):
Thanks for adding the test! Looks good! I also really like how the new tests look like :+1:

PR #167, opened 2025-01-09
Check license header with pre-commit hook

davidrudlstorfer:
Also update year to 202~4~5

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,6 +29,7 @@ repos:
         name: Check valid YAML syntax
       - id: end-of-file-fixer
         name: Check for newline at end of file
+        exclude: '^LICENSE$'
       - id: debug-statements
         name: Check for debugger imports
       - id: no-commit-to-branch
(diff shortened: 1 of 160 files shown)


isteinbrecher (approved):
Looks good to me!

I think you still have to update the files added with #166

davidrudlstorfer:
Rebased onto latest main

Would be ready to merge

isteinbrecher:
Would it be possible to keep the lines
```
# -----------------------------------------------------------------------------
```
before and after the license?

davidrudlstorfer:
So I've actually misinterpreted the pre-commit hook and there's actually no option to add a line before and after the license.

We can either add the lines directly into the license (I would not do that) or we drop the lines before and after the license in the source code.

Sorry for my misunderstanding.

Is there actually a reason why you want those lines? I've not seen that in a code base before and in my opinion it's easier to read without those lines

isteinbrecher:
I personally like the lines because I think they help to have a cut between the license and the rest of the text, but if it is not possible, lets go with the proposed approach!

davidrudlstorfer:
@isteinbrecher this would be ready to review

isteinbrecher (approved):

PR #172, opened 2025-01-10
Allow `parametrize` tests

isteinbrecher:
While working on porting the geometric search tests, I saw that we have a small bug in the testing framework that messes up our test selection process when using `@pytest.mark.parametrize`. This should fix it.


davidrudlstorfer:
Nice catch! Thanks for the fix

davidrudlstorfer (approved):

PR #173, opened 2025-01-10
Fix test failure output

isteinbrecher:
Fix the error output for a failed comparison.


davidrudlstorfer (approved):
Thanks for the next fix 😅

Issue #159, opened 2024-12-17
Discuss if MeshPy should move to a src layout


davidrudlstorfer:
Currently MeshPy is structured in a flat layout, i.e. we have the meshpy module at the top as well as "external"

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

https://www.reddit.com/r/learnpython/comments/wpkr4q/using_a_src_directory_for_a_python_package/

https://www.pyopensci.org/python-package-guide/package-structure-code/python-package-structure.html

For simplification of the entire project and its structure I propose to switch to a src based layout

```
meshpy/
├─ src/
│  ├─ pybind11 (as git submodule)/
│  ├─ meshpy/
│  │  ├─ core/
│  │  │  ├─ inputfile.py
│  │  │  ├─ function.py
│  │  │  ├─ ...
│  │  ├─ four_c/
│  │  ├─ geometric_search/
│  │  ├─ .../
│  ├─ utilities/

```

The structure within the `tests/` folder should identically reflect the src structure of meshpy

This would allow us to easily define all modules and also export meshpy as a package

davidrudlstorfer:
Thanks for already looking into this @isteinbrecher! This was the last point on my personal ToDo list for MeshPy which was not yet turned into an issue here on github 😅

For sure this should only happen once the testing is ported 🙏🏻

davidrudlstorfer:
One question or rather discussion because this would be a breaking change for MeshPy

Is there actually a valid reason why we specify so many subpackages within MeshPy (i.e., abaqus, four_c, geometric_search)?

Yesterday I setup a local branch regarding the API documentation for MeshPy - once again I ran into the problem that MeshPy is not a single package but multiple packages - this is not really supported by any documentation. This will also lead to problems down the road if we want to publish the package to PyPi as far as i know

In my opinion we should just have one package, i.e. MeshPy. We can still access all the stuff via, for example, `meshpy.rotation.blablub`

So is there actually a reason behind all of that or can this also be simplified when moving to the src layout?

I would strongly add the move here, because once we change the layout structure of MeshPy - other 3rd party tools need to adjust imports - this could then be done in one sweep

Curious about your opinion @isteinbrecher

isteinbrecher:
[...]

PR #174, opened 2025-01-10
Port geometric search tests

isteinbrecher:
Port all geometric search tests to pytest


isteinbrecher:
@davidrudlstorfer I think the current testing does not work if we install MeshPy in non-editable mode. The reason is somehow related to the installation of the Cython geometric search functionality. I think this might have been already the case before the switch to pytest. Therefore, I suggest to change all pipelines to use editable MeshPy installations (`pip install -e .`) so we can get all the test running again as soon as possible.  We can track this in a separate issue.

tests/test_geometric_search.py
```diff
@@ -0,0 +1,529 @@
...
+    partner_indices_to_point_partners,
+    point_partners_to_partner_indices,
+    point_partners_to_unique_indices,
+)
+from meshpy.mesh_creation_functions.beam_honeycomb import (
+    create_beam_mesh_honeycomb_flat,
+)
+from meshpy.utility import filter_nodes, get_nodal_coordinates
+
+PYTEST_GEOMETRIT_SEARCH_PARAMETRIZE = [
```
davidrudlstorfer, line 52:
```suggestion
PYTEST_GEOMETRIC_SEARCH_PARAMETRIZE = [
```

davidrudlstorfer (dismissed):

davidrudlstorfer:
> @davidrudlstorfer I think the current testing does not work if we install MeshPy in non-editable mode. The reason is somehow related to the installation of the Cython geometric search functionality. I think this might have been already the case before the switch to pytest. Therefore, I suggest to change all pipelines to use editable MeshPy installations (pip install -e .) so we can get all the test running again as soon as possible. We can track this in a separate issue.

@isteinbrecher nice that you already found a solution! For me it's not a problem that we install everything in editable mode. I think once the modularity of MeshPy is cleaned up => #159 everything resolves itself

davidrudlstorfer (approved):

isteinbrecher:
That is the same idea I had, #159 should hopefully automatically solve this.

PR #175, opened 2025-01-10
Fix testname for parameterised tests

isteinbrecher:
Fix the test name. Now we get the base test name and not the one amended with the parameters.

Also improve the error output on failed path compare.


davidrudlstorfer (approved):
Nice change - thanks!

PR #176, opened 2025-01-10
Meshpy tests to pytest

isteinbrecher:
Port the "main" test file to pytest.


tests/testing_meshpy.py
```diff
@@ -1,1974 +0,0 @@
...
-
-    def setUp(self):
-        """This method is called before each test and sets the default meshpy
-        values for each test.
-
-        The values can be changed in the individual tests.
-        """
-
-        # Set default values for global parameters.
-        mpy.set_default_values()
```
isteinbrecher, line 136: @davidrudlstorfer this is the only thing I could not figure out. Some tests change the values in the global object `mpy`. With the old framework this construct ensured that the values are set to the default values again. Is this somehow possible with `pytest` to reset this before every test?

davidrudlstorfer: We can simply create a fixture which is destroyed after each test

For that we can use `@pytest.fixture(scope="function")`

We can either create it in `conftest.py` if more tests use it - or there should also be a way to simply add the pytest.fixture decorator to individual functions and do it there

isteinbrecher: Would that be possible to automatically have this for all tests? This is definitely not the best approach, but as of now, the order of the tests matters, i.e., if I change an option in one test, I have to reset to the default values in the next one that is executed. This can be very error prone, so I would simply automatically reset the `mpy` object for every test.

isteinbrecher: Thanks! That works like a charm!

tests/test_meshpy.py
```diff
@@ -0,0 +1,1938 @@
...
+        else:
+            mesh.reflect(2 * (rot_2 * [1, 0, 0]), flip_beams=flip)
+
+        # Compare the dat files.
+        assert_results_equal(mesh_ref, mesh)
+
+    # Compare all 4 possible variations.
+    for flip in [True, False]:
+        for origin in [True, False]:
+            compare_reflection(origin=origin, flip=flip)
```
davidrudlstorfer, line 254: Could we maybe use the pytest parameterize feature for that?

(Similar to the next test)

tests/test_meshpy.py
```diff
@@ -0,0 +1,1938 @@
...
+):
+    """Test that the input file for all types of Kirchhoff beams is generated
+    correctly."""
+
+    # Create input file.
+    input_file = InputFile()
+
+    with warnings.catch_warnings():
+        # Ignore the warnings for the rotvec beams.
+        warnings.simplefilter("ignore")
```
davidrudlstorfer, line 594: Which warnings are we ignoring here? Should we maybe change the source code so the code runs through and the warnings are simply printed?

tests/test_meshpy.py
```diff
@@ -0,0 +1,1938 @@
...
+            Coupling(*args)
+    else:
+        Coupling(*args, check_overlapping_nodes=False)
+
+
+def test_meshpy_check_overlapping_coupling_nodes():
+    """Perform the test that the coupling nodes can be tested if they are at
+    the same position."""
+    perform_test_check_overlapping_coupling_nodes(True)
+    perform_test_check_overlapping_coupling_nodes(False)
```
davidrudlstorfer, line 1891: Same here - could we remove that function and just use pytest parameterize?

davidrudlstorfer (approved):
Really nice! I have small questions/suggestions otherwise it looks greate - I really like the simplicity of our testing approach across everything :D

tests/test_meshpy.py
isteinbrecher: I don't have a strong opinion on this (also I didn't think of it for a long time). If we want to change this, I would do so in a separate PR, once all the tests are ported.

PR #177, opened 2025-01-10
Port nurbs tests to pytest

davidrudlstorfer:


davidrudlstorfer:
I really really like our new testing approach 😅 porting of this file (there's nothing special happening) took less than 5 minutes

isteinbrecher (approved):

PR #180, opened 2025-01-10
Port mesh creation tests

isteinbrecher:


tests/test_mesh_creation_functions.py
```diff
@@ -0,0 +1,988 @@
...
+            Beam3rHerm2Line3,
+            mat,
+            [1.0, 2.0, 0.0],
+            [3.0, 4.0, 6.0],
+            n_el=1,
+            l_el=l_el,
+        )
+    with pytest.raises(ValueError):
+        mesh = Mesh()
+        return create_beam_mesh_function(mesh, interval=[0.0, 1.0], l_el=2.0)
```
davidrudlstorfer, line 592: Similar to the other test - what are we doing here?

davidrudlstorfer (dismissed):

tests/test_mesh_creation_functions.py
isteinbrecher: These are tests to check that wrong input arguments raise errors (likely `ValueErrror` is not the best choice in this case).

davidrudlstorfer: Could we maybe add why it is wrong? Not even I am sure what's wrong - so maybe it's good for the user to add the problem which is tested here.

On a second look - is it that n_el and l_el is provided?

davidrudlstorfer (approved):
Thanks for adding the information regarding the warnings - it's a lot clearer now

PR #181, opened 2025-01-10
Performance tests

isteinbrecher:


isteinbrecher:
Seems the failure is due to the same reason as for the performance tests. I think this could be related to some relative import  hacks that we have.

davidrudlstorfer (dismissed):
```
tests/test_performance.py::test_performance
  /home_local/github-runner/actions-runner/_work/meshpy/meshpy/tests/test_performance.py:229: UserWarning: Expected time not reached in function cubitpy_create_solid!
    warnings.warn("Expected time not reached in function {}!".format(name))
```

These warnings in the performance testing output are fine?

davidrudlstorfer:
Very interesting that nearly all performance tests "fail" in the sense that they are all slower.

isteinbrecher:
Good catch, the warnings indicate a failed performance test. So we need to actually fail the performance tests in that case. I thought I did that by adding the last line of `test_performance.py`, but does not seem to be the case 🤔.

The more critical part is that the tests actually fail. I already noticed this in the first version of this PR, but was hoping (not realistically) that this was due to the missing Cython geometric search. Turns out it is not. Here is the log from the last run performance tests:
```
Times for cubitpy_create_solid:
    Expected: 8sec
    Actual:   8.69sec
    FAILED
Times for meshpy_load_solid:
    Expected: 1.5sec
    Actual:   1.72sec
    FAILED
Times for meshpy_load_solid_full:
    Expected: 3.5sec
    Actual:   4.08sec
    FAILED
Times for meshpy_create_beams:
    Expected: 9sec
    Actual:   14.4sec
    FAILED
Times for meshpy_rotate:
    Expected: 0.6sec
    Actual:   0.798sec
    FAILED
Times for meshpy_translate:
    Expected: 0.5sec
    Actual:   0.253sec
    OK
Times for meshpy_reflect:
    Expected: 0.7sec
    Actual:   0.901sec
    FAILED
Times for meshpy_wrap_around_cylinder:
    Expected: 2sec
    Actual:   2.82sec
    FAILED
Times for meshpy_wrap_around_cylinder_without_check:
    Expected: 0.7sec
    Actual:   0.811sec
    FAILED
Times for meshpy_find_close_nodes:
    Expected: 0.5sec
    Actual:   0.404sec
    OK
Times for meshpy_write_dat:
    Expected: 9sec
    Actual:   14.6sec
    FAILED
Times for meshpy_write_vtk:
    Expected: 4.5sec
    Actual:   6.47sec
    FAILED
Times for meshpy_write_vtk_smooth:
    Expected: 9sec
    Actual:   15.1sec
    FAILED
Times for geometric_search_find_nodes_brute_force:
    Expected: 0.05sec
    Actual:   0.0233sec
    OK
```
I also re-run the a set of performance tests from before the change to `pytest` https://github.com/imcs-compsim/meshpy/actions/runs/12671070946/job/35515975586 and they passed, so the testing machine should be fine. Is it possible that pytest slows down the execution of the tests? Also we used to not run the performance tests with coverage so maybe that also affects performance. One more possible reason, where is the actual `tmp` directory located? Maybe we now run the tests on a slower drive (don't think that is the case because the slower drive has a very cryptic name and should not be the "default" pytest temp directory).

@davidrudlstorfer May I ask you to have a look at those points?

davidrudlstorfer:
The assert now works as expected

davidrudlstorfer:
Everything works as expected - disabling the coverage for the performance runs works on a single github action workflow

We can reuse this approach in other workflows (cubitpy, ...) once we set a higher minimum coverage

Feel free to restructure my commits. We can now either deactivate coverage or increase test times - I'll leave it up to you

isteinbrecher:
Thanks, I always get the `assert` confused.

Good to see that we identified the issue. I would opt for deactivating the coverage for the performance tests. The fewer "additional overheads" we have in the performance test times, the closer those times should represent the actual performance. Also, coverage should be exactly the same with and without performance tests. Since we don't do any result tests in the performance tests, I would deactivate the coverage here, even if the tests weren't failing. We can think of adding basic result tests at some point, but low priority in my opinion, since all the functionality in the performance tests should be covered elsewhere.

davidrudlstorfer (approved):

isteinbrecher:
I restructured the commits, if the pipeline passes this should be ready to go!

davidrudlstorfer (approved):

PR #183, opened 2025-01-10
Port testing 4C

isteinbrecher:


tests/test_four_c.py
```diff
@@ -0,0 +1,399 @@
...
+    assert_results_equal(get_corresponding_reference_file_path(), input_file)
+
+
+def test_four_c_simulation_beam_potential_helix(
+    assert_results_equal, get_corresponding_reference_file_path
+):
+    """Test the correct creation of input files for simulations including beam
+    to beam potential interactions."""
+
+    def create_model():
```
davidrudlstorfer, line 99: I know its the code I've written - maybe we could directly remove that functino as it is only called once - I can also just do it if you like

davidrudlstorfer (approved):

tests/test_four_c.py
isteinbrecher: I will restrict the commits anyway and can do that on the fly. In the long run, when restructuring the tests, we cad think of providing a few general mesh creation functions that can be used in all test.

davidrudlstorfer: Sorry for the question - what do you mean with "restricting the commits"?

Adding general mesh creation functions sounds like a good plan

isteinbrecher: Sorry, autocorrect - restructure.

davidrudlstorfer (approved):
Thanks for directly fixing this!

PR #187, opened 2025-01-13
Remove old tests

davidrudlstorfer:
Remove header tests => will be replaced by pre-commit hook

Remove old testing main => replaced with pytest


isteinbrecher:
Can we wait with this until all tests are ported and the header pre-commit is merged? Then we can also clean up the dummy tests within this PR.

isteinbrecher (approved):

PR #189, opened 2025-01-13
Port four c simulation tests

isteinbrecher:
Port the 4C simulation functionality.

This proposes the following approach to keep the current behaviour:
We test all test cases in this file twice. Once we only run up to the first call of 4C and compare the created input files, this allows to run some core functionalities of this file even if 4C is not available. To achieve "full" test coverage we also run the test, where we enforce 4C to be run.


tests/test_four_c_simulation.py
```diff
@@ -0,0 +1,965 @@
...
+from meshpy.utility import check_node_by_coordinate
+
+# We test all test cases in this file twice. Once we only run up to the first
+# call of 4C and compare the created input files, this allows to run some core
+# functionalities of this file even if 4C is not available. To achieve "full"
+# test coverage we also run the test, where we enforce 4C to be run.
+PYTEST_4C_SIMULATION_PARAMETRIZE = [
+    "enforce_four_c",
+    [False, pytest.param(True, marks=pytest.mark.fourc)],
+]
```
davidrudlstorfer, line 68: I like this approach 👌 with it we do not have any duplicated code and everything is tested in the regular pipeline

isteinbrecher:
Looks like all the tests pass, this means that with this PR and #176 the pytest transition should be done!

davidrudlstorfer (approved):

PR #194, opened 2025-01-13
Separate actions

davidrudlstorfer:
separate the actions on github to create two separate badges for the readme => one for the test suite and one for the code checks which will be linked

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,5 +1,4 @@
-
-name: MeshPy test suite
+name: Test suite
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


tests/test_performance.py
```diff
@@ -35,7 +35,6 @@
 
 import numpy as np
 import pytest
-from cubitpy import CubitPy, cupy
```
isteinbrecher: I assume this was added by accident.

.github/workflows/testing.yml
```diff
@@ -1,5 +1,5 @@
```
isteinbrecher, line 1: Can we remove the leading empty lines in both yaml files?

isteinbrecher (dismissed):

.github/workflows/testing.yml
davidrudlstorfer: Thanks! Changed it

isteinbrecher (approved):

PR #197, opened 2025-01-16
Website

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher could you briefly change the protection rules to also allow me to push to github pages from this branch?

With this change I could test the website within this PR and I do not need to always merge into main

davidrudlstorfer:
Theoretically everything should work with the current changes - let's test it once the github pages protection rules are updated by you -> then I can change the action so it is only built/deployed on main

davidrudlstorfer:
Screenshot from 2025-01-16 11-03-57

Website now looks like this - the readme is symlinked into the website folder and therefore automatically up-to-date

I've also added the github link on the top right corner of the website

isteinbrecher:
I temporarily remove the restriction to the `github-pages` environment, so this should now work (in theory). I will have a look at the changes later, or tomorrow. I am already really curious how this will look!

davidrudlstorfer:
Unfortunately, as you can see I am still having troubles to deploy the website on github pages

I always get the error:

```
Error: Ensure GITHUB_TOKEN has permission "id-token: write".
```

And in the log it says

```
GITHUB_TOKEN Permissions
  Metadata: read
  Pages: read
```

So I still think that I am not able to push the website due to a wrong github token setting

davidrudlstorfer:
Thanks for giving me the temporary rights! I am just too dumb to figure out what's going wrong.

Currently I think that the problem is that the branch is forked on my repo and that it is not a branch on the main repo

I'll test it out if it works on a meshpy branch

davidrudlstorfer:
Closing in favor of #199

PR #198, opened 2025-01-16
Create and upload coverage badge and coverage report

davidrudlstorfer:


.github/workflows/testing.yml
```diff
@@ -96,6 +96,22 @@ jobs:
...
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-report-${{github.job}}
+          path: htmlcov/
+      - name: Create coverage badge
+        run: |
+          cd ${GITHUB_WORKSPACE}
+          source python-testing-environment/bin/activate
+          pip install coverage-badge
```
isteinbrecher, line 108: Should we add this to the CI-CD dependencies in the `pyproject.toml` file?

isteinbrecher (approved):
Great to see that this (hopefully 🤞) works! I am fine with merging this and then adding the stuff to the readme in a separate PR or doing everything here.

.github/workflows/testing.yml
davidrudlstorfer: Good idea 👍 it was actually already in there so I've removed it from the workflow

I would extend the website in another PR because we need some changes - therefore, this would be ready.

isteinbrecher (approved):

PR #200, opened 2025-01-17
Website fixes

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher so the asset links now work

I still do not know why the workflow triggered although the test suite was not completed. I added everything according to the Github documentation https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

isteinbrecher (approved):
I also don't know why the workflow ordering did not work. Maybe the removal of the quotes did the trick, let's see.

davidrudlstorfer:
I would need another approval - now the centered header also works

isteinbrecher (approved):

PR #201, opened 2025-01-20
Fix bug in assert results equal functionality

isteinbrecher:
This PR fixes a small bug in the `assert_results_equal` function, where when we passed two files with the file ending `dat` no actually comparison took place. The function is now modified in such a way, that the default is a string comparison. If some other case occurs, e.g., `vtu` comparison or `dict` comparison (will be added in a follow up PR), we have to handle it and exit the function before the string comparison.


davidrudlstorfer (approved):
Looks good 👍 Thanks for the fix and changes - especially the good comments describing what we are doing

PR #202, opened 2025-01-20
Add dictionary comparison to testing framework

isteinbrecher:
This PR adds the functionality to compare `dict` to each other within the testing framework. If the value is a numpy object, the comparison will be done using tolerances. In a future PR, this functionality will be used to compare "pure" mesh data with each other without having the overhead of creating a whole input file.

The dictionary case is also added to the `get_string` function in the testing framework (we used a `json` format to store the `dict` in a string).

I adapted the test case `test_cosserat_curve_project_point` to use the new functionality (this one used a similar functionality directly coded in the test) and I am again very happy to see how easy and clean this is with the new testing framework - the work put in already is paying off!

Follows #201


tests/reference-files/test_cosserat_mesh_transformation.json
```diff
@@ -0,0 +1,4378 @@
+{
+    "pos": [
+        [
+            [
+                -1.7999999999999998,
+                0.0,
+                15.000000000000004
```
isteinbrecher, line 7: The only thing I don't like about this new approach is the formatting of the json file. @davidrudlstorfer do you know if there is a formatter that can "save" some lines here?

tests/conftest.py
```diff
@@ -313,6 +327,19 @@ def _assert_results_equal(
...
+            # Do a dictionary comparison
+
+            # First check if we need to load the dictionary from a file for one of the
+            # arguments.
+            if isinstance(reference, Path):
+                reference = json.loads(get_string(reference))
+            if isinstance(result, Path):
+                result = json.loads(get_string(result))
+
+            compare_dicts(reference, result, rtol=rtol, atol=atol)  # type: ignore[arg-type]
```
davidrudlstorfer, line 340: Why do we need the ignore here? Can we fix it?

tests/conftest.py
```diff
@@ -549,3 +576,53 @@ def get_return_values(flag, message):
...
+        if isinstance(value_1, np.ndarray) or isinstance(value_2, np.ndarray):
+            if not np.allclose(value_1, value_2, rtol=rtol, atol=atol):
+                raise AssertionError(
+                    f'Comparison of numpy arrays for the key "{key}" failed.'
+                )
+        elif isinstance(value_1, dict) and isinstance(value_2, dict):
+            # If both values are dictionaries, compare recursively
+            compare_dicts(value_1, value_2, rtol=rtol, atol=atol)
+        elif not value_1 == value_2:
+            raise AssertionError(f'Comparison of values for the key "{key}" failed.')
```
davidrudlstorfer, line 628: Should we also add a float comparison here?

tests/reference-files/test_cosserat_mesh_transformation.json
davidrudlstorfer: I would keep the current approach - the json is formatted as the standard suggests

I do not even know if it's possible to format it in another way

tests/test_cosserat_curve.py
```diff
@@ -176,26 +180,11 @@ def test_cosserat_mesh_transformation(reference_file_directory, current_test_nam
...
-    pos_np = np.array(pos)
-    rot_np = quaternion.as_float_array(rot)
-
-    assert np.allclose(pos_ref, pos_np, rtol=1e-14)
-    assert np.allclose(rot_ref, rot_np, rtol=1e-14)
+    assert_results_equal(
+        get_corresponding_reference_file_path(extension="json"),
+        {"pos": pos, "rot": quaternion.as_float_array(rot)},
+        rtol=1e-14,
+    )
```
davidrudlstorfer, line 187: Really nice!

davidrudlstorfer (commented):
Looks good - I commented a few minor points

tests/conftest.py
isteinbrecher: If if don't have this, I get the following error:
```
Argument 1 to "compare_dicts" has incompatible type "Path | str"; expected "dict[Any, Any]"  [arg-type]
```
in my opinion the type should be clear here (but the logic is maybe a little bit too complicated for `mypy`. We could reformulate the code, but I could not think of a very nice simple solution, so I decided to skip the type error here.

tests/conftest.py
isteinbrecher: Good point, I added it.

isteinbrecher (commented):
Thanks for your comments @davidrudlstorfer . If you have a better idea for the type error, I am open for it, otherwise, I would merge this.

tests/conftest.py
davidrudlstorfer: I've now further analyzed the code and to me its not quite clear what's happening and probably mypy is correct

1. we should add the type anotations to `_assert_results_equal` -> there the dict is missing
2. result can also be of type input file -> if it's an input file the function compare_dicts is called with a wrong type and it's a problem again (we could add better type checks for the dict comparison to only call it if it's a Path or a Dict)
3. the extension of the get_string function is now also not clear to me, if we put a dict into the get_string function we get a dict back - so it's not a string and should not be put into that function

I think the problems should be resolved to restore the correct workings of all functions, especially the _get_string function

isteinbrecher: I uploaded a new commit which should takes a very similar approach to what we do in the string comparison. We now have a function `get_dictionary` that converts all supported types to a dictionary and fails otherwise. This really helped to clean up the code, satisfied `mypy` and catch wrong types. Thanks for suggesting to look into this!

Regarding point 3, are you sure, we return `json.dumps(...)` which, as far as I know, is always a string.

davidrudlstorfer (approved):
@isteinbrecher thanks for going the extra mile and improving this!

Regarding the _get_string function - that was my bad - you are correct and it returns a string. I mixed up json.dumps with other json functions

I am now very happy how this turned out - sometimes mypy is also wrong but in reality it often shows that the code design is not optimal and I think you really improved the code here and now it's much easier to understand.

Thank you and let's get this merged! 🚀

PR #203, opened 2025-01-20
Update README /. website

isteinbrecher:
Update some things in the README /. website
- Add all references
- Add contributors
- Update text
- Add honeycomb structure example


davidrudlstorfer:
@isteinbrecher I do not know if the updates are finished yet but I wanted to leave a comment with a few ideas what we could add

- due to the readme getting quite long I think it would be beneficial to add a brief TOC at the beginning, there are extension for VSCode available which create these automatically

- I think the usage / reference to the tutorials should be after the installation in the readme
- Also the citation could move to the bottom (or more specifically to the papers which use MeshPy
- A good place for the `pre-commit install` would be in the contributing as it is not necessary for a normal install
(- And after you're read we can delete the old website data)

Sorry if it's not ready yet

isteinbrecher:
Yes, this was ready for review.

>- due to the readme getting quite long I think it would be beneficial to add a brief TOC at the beginning, there are extension for VSCode available which create these automatically
>- I think the usage / reference to the tutorials should be after the installation in the readme

That is a good point. The order in the `README.md` can definitely be updated. What to you think about:
```
# MeshPy
## Examples
## How to use MeshPy?
## How to cite MeshPy?
## Work that uses MeshPy
### Peer-reviewed articles
### PhD thesis
## Installation
### Optional dependency on [ArborX](https://github.com/arborx/ArborX)
## Contributing
## Contributors
```

TOC would definitely be a very good addition. Do you have the name of that extension? Otherwise I can also add it by hand, the TOC should not change that often. Side note: how would we do that with the absolute/relative links there?

>- Also the citation could move to the bottom (or more specifically to the papers which use MeshPy

I agree, that that would be the "better" ordering, but for strategic reasons I would like to keep that close to the top. So maybe move up the "Work that uses MeshPy"? (As suggested in the previous list)

>- A good place for the `pre-commit install` would be in the contributing as it is not necessary for a normal install

I changed that

>(- And after you're read we can delete the old website data)

Thanks for the reminder, I forgot about that. Is fixed.

davidrudlstorfer:
I like your suggested approach for the new order in the README 👍

Regarding the TOC - I use Markdown All in One - you should see the option for the TOC at a right click https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one

Regarding the links within the TOC - these work within the Markdown file so as far as I understand/know we need no relative/absolute paths

@isteinbrecher once these changes are added let's merge 🚀

isteinbrecher:
Thanks for the link, the plugin works like a charm. I updated the ordering and this should be ready to merge.

davidrudlstorfer (approved):

Issue #171, opened 2025-01-10
Rethink tutorials for MeshPy


davidrudlstorfer:
Create new tutorials

Some ideas and notes:

- [ ] Use jupyter
- [ ] Maybe host them on binder to automatically interactively host them? (for example see https://github.com/TUM-LNM/FEMTOE)
- [ ] Reimplement/Rethink testing of tutorials

isteinbrecher:
Once #204 is merged, the original points of this issue should be resolved. Let's still use this to discuss the structure and content of the examples we want. My first suggestion:

- [x] Example 1: Large rotations
- [ ] Example 2: Basic geometry functions: Line, Arc, Helix, Parametric curve
- [ ] Example 3: Mesh manipulation functionality: Translate, Rotate, warp around cylinder, reflect
- [ ] Example 4: Advanced geometry creation: start_node, couple nodes (maybe move this to the end of example 3)
- [ ] Example 5: Advanced geometry creation functions: Honeycomb, Stent, Wire, Fibres in rectangle
- [ ] Example 6: MeshPy and 4C with solid elements
- [ ] Example 7: MeshPy and Abaqus
- [ ] Example 8: Cosserat curve?

davidrudlstorfer:
Final ToDo after all examples are ready

- [ ] Delete everything from the old tutorial `/tutorial`
- [ ] Delete mentions of `/tutorial` in the pre-commit hook, specifically for mypy (see #244) 
- [ ] Consider adding explicit links to the README again https://github.com/beamme-py/beamme/pull/682#discussion_r3622694574

PR #204, opened 2025-01-20
Add examples in .ipynb notebooks

isteinbrecher:
This PR adds a first ipython notebook example for MeshPy (how to interact with the finite rotation framework).

I like the way the notebooks like very much compared to the old tutorial.
This is definitely a better way to explain how the code works and I think we should move all the content of the tutorial to similar example notebooks (I had a look at the tutorial and this can likely be split up to show a number of different features in MeshPy).

The only drawback is that the handling of such notebooks in combination with git is somehow (even in the year 2025) very complicated.
I had a look at a number of different tools and formats and nothing is perfect in my opinion.
Good news first, with `testbook`, testing of the examples should be straight forward.

**py:percent**
I had a look at [The percent format](https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-percent-format) which seems very nice at first glance. Here are the pros and cons:
- Pros:
  - Everything is nicely stored in a simple `.py` file
  - Our code formatters and pre-commit hooks should work out of the box
  - Perfect for version control
- Cons:
  - Not as widely used as `ipynb`
  - Editing the file (at least in VSCode) is much less intuitive than with `ipynb`
  - Not rendered at all in GitHub (in `ipynb` files at least the markdown is rendered)
 
There exist ways to convert between `ipynb` and `py:percent` (https://jupytext.readthedocs.io/en/latest/index.html), however, I am not convinced yet that this is the best solution for what we need.

**ipynb**
Here are the pros and cons for `ipynb`:
- Pros:
  - More or less standard for notebooks with python
  - Markdown is always rendered in GitHub
  - Intuitively to handle in VSCode
  - Should work without any issues in combination with `my binder`
- Cons:
  - Store a lot of unnecessary data
  - Even if all the meta data and output is removed, the files are still not as nicely readable as `py:percent`
  - Code checks and pre-commit hooks are not straightforward (should be possible but one has to take a look)
 
I think for the use case we habe, i.e., around 5 examples which are not regularly updated, the way to go is with `ipynb` and use a tool that ensures that all metadata and output is striped before committing.

@davidrudlstorfer I am curios on your opinion, we can move this discussion to #171 or we can try to resolve this here with the first example.

### DIFF
--- a/.binder/Dockerfile
+++ b/.binder/Dockerfile
@@ -1,0 +1,45 @@
+# MeshPy: A beam finite element input generator
+#
+# MIT License
+#
+# Copyright (c) 2018-2025
+#     Ivo Steinbrecher
+#     Institute for Mathematics and Computer-Based Simulation
+#     Universitaet der Bundeswehr Muenchen
+#     https://www.unibw.de/imcs-en
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+ARG BASE_IMAGE_TAG=latest
+FROM ghcr.io/pyvista/pyvista:$BASE_IMAGE_TAG
+
+USER root
+RUN apt-get upgrade
+RUN apt-get update
+RUN apt-get install -y git
+RUN apt-get install -y build-essential
+RUN apt-get install -y ffmpeg libsm6 libxext6
+RUN apt-get install -y libgl1-mesa-glx xvfb
+
+COPY . ${HOME}
+USER root
+RUN chown -R 1000 ${HOME}
+RUN chown -R 1000 ${HOME}/.*
+USER jovyan
+WORKDIR ${HOME}
+RUN pip install -e .
--- a/.binder/start
+++ b/.binder/start
@@ -1,0 +1,32 @@
+#!/bin/bash
+# MeshPy: A beam finite element input generator
+#
+# MIT License
+#
+# Copyright (c) 2018-2025
+#     Ivo Steinbrecher
+#     Institute for Mathematics and Computer-Based Simulation
+#     Universitaet der Bundeswehr Muenchen
+#     https://www.unibw.de/imcs-en
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Copied from https://mybinder.org/v2/gh/pyvista/pyvista-examples/master
+export PYVISTA_TRAME_SERVER_PROXY_PREFIX="$JUPYTERHUB_SERVICE_PREFIX/proxy/"
+exec "$@"
--- a/README.md
+++ b/README.md
@@ -67,7 +67,15 @@ MeshPy is developed at the [Institute for Mathematics and Computer-Based Simulat
 
 ## How to use MeshPy?
 
-Basic tutorials can be found in the directory `tutorial/`.
+MeshPy provides example notebooks to showcase its core features and functionality.
+The examples can be found in the `examples/` directory.
+They can be run locally or directly tested from your browser via the following links:
+
+- Example 1: **Finite rotation framework** [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/imcs-compsim/meshpy/add-examples?labpath=examples%2Fexample_1_finite_rotations.ipynb)
+- Example 2: **Core mesh generation functions** [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/imcs-compsim/meshpy/add-examples?labpath=examples%2Fexample_2_core_mesh_generation_functions.ipynb)
+
+You can also interactively test the entire MeshPy framework directly from your browser here [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/imcs-compsim/meshpy/add-examples)
+
 
 ## How to cite MeshPy?
 
--- a/meshpy/mesh.py
+++ b/meshpy/mesh.py
@@ -57,6 +57,7 @@ from .utility import (
     get_nodal_coordinates,
     get_nodal_quaternions,
     get_nodes_by_function,
+    is_testing_github,
 )
 from .vtk_writer import VTKWriter
 
@@ -769,11 +770,13 @@ class Mesh:
         beam_cross_section_directors=True,
         beam_radius_for_display=None,
         resolution=20,
-        is_testing=False,
         parallel_projection=False,
         **kwargs,
     ):
         """Display the mesh in pyvista.
+
+        If this is called in a GitHub testing run, nothing will be shown, instead
+        the pv.plotter object will be returned.
 
         Args
         ----
@@ -792,9 +795,6 @@ class Mesh:
         resolution: int
             Indicates how many triangulations will be performed to visualize arrows,
             tubes and spheres.
-        is_testing: bool
-            Flag if the function is used for testing. If true, the pv.plotter object
-            will be returned.
         parallel_projection: bool
             Flag to change camera view to parallel projection.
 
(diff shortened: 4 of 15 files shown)


davidrudlstorfer:
@isteinbrecher thanks for looking into the new tutorials/examples!

A big thanks for looking into currently available options and listing their pros and cons. Simply because `.ipynb` is widely used I would go for it.

Could we directly add the Binder integration within this PR (https://mybinder.readthedocs.io/en/latest/index.html)? Then the examples/tutorials can be executed directly from within the browser and we can link it from our readme

Do you have an idea why the tests fail?

isteinbrecher:
> Do you have an idea why the tests fail?

I don't really know. I did a quick google search but did not find anything concrete. For now, I am happy that the test passes on macOS. I will have a look at the `my binder` setup, maybe that will also identify the issue with the testing.

> And one thing I noticed while looking through the example - do want to add the header at the top and then define the auxiliary functions? In my opinion it's a bit odd to scroll down a long way to even find what this example is doing

Very good point, I also did not like that. I think I have a nice solution, we now have the submodule `meshpy.examples` which provides the functionality and can simply be imported at the very top with a single line of code. Note: I also moved the `meshpy_testing` module to be in the main MeshPy namespace `meshpy.testing` and `meshpy_tutorial` will be removed anyway.

@davidrudlstorfer could you maybe have a look at how to do the code checks and pre-commit hooks? If we get all the "standard" code checks working that would be nice, but otherwise I would be fine with having `isort` and a code formatter. I used `nbstripout example_1_finite_rotations.ipynb` to "clean" the notebook, can we somehow enforce this with a pre-commit hook? Similar to a code formatter?

Once this is merged, I think we can finally have a look at the repo structure and then all the infrastructure work should be done!

davidrudlstorfer:
I gave the failing tests a short try and they also locally failed on my Ubuntu 20.04 machine. Installing ipykernel fixed the issue, the pipeline is still failing

Probably it's some weird package that is pre-installed on macOS and not other systems

And I'll look into the pre-commit hook:)

isteinbrecher:
Short update: This is now in a state where github testing (except the code checks) and mybinder works. These badges directly link to the repo and the two examples:

Full repo
[Binder](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples)

Example 1
[Binder](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples?labpath=example_1_finite_rotations.ipynb)

Example 2
[Binder](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples?labpath=example_2.ipynb)

davidrudlstorfer:
@isteinbrecher thanks for your work on the examples and on binder! It looks so nice and I am more than impressed with everything!

Regarding the pre-commit hook - the problem was that the C/C++ formatter clang also formatted the ipynb files - simply deactivating it worked. Ruff now formats everything out of the box. Let's see if everything works

Sorry if i somehow interfered with your commits

examples/utils/utils.py
```diff
@@ -0,0 +1,148 @@
...
+import sys
+
+import numpy as np
+import pyvista as pv
+
+# Store the default system out, so we can reset it after pyvista changes it
+stdout = sys.stdout
+
+# If we are on binder, set to server side rendering
+if "BINDER_LAUNCH_HOST" in os.environ.keys():
```
isteinbrecher, line 40: Should we define globally available functions like
```
is_mybinder
is_testing
is_github_testing
```
in `meshpy.utils`?

examples/utils/utils.py
```diff
@@ -0,0 +1,148 @@
...
+        return self.plotter
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """When exiting the with statement, call this function.
+
+        We show the plotter (except during testing and we reset the
+        console print out).
+        """
+
+        if "GITHUB_ACTION" not in os.environ.keys():
```
isteinbrecher, line 146: See my comment above.

meshpy/mesh.py
```diff
@@ -906,7 +906,7 @@ def display_pyvista(
             solid_grid = pv.UnstructuredGrid(vtk_writer_solid.grid).clean()
             plotter.add_mesh(solid_grid, color="white", show_edges=True, opacity=0.5)
 
-        if not is_testing:
+        if "GITHUB_ACTION" not in os.environ.keys():
```
isteinbrecher, line 909: See above.

pyproject.toml
```diff
@@ -7,16 +7,18 @@ packages = [
   "meshpy",
   "meshpy.abaqus",
   "meshpy.cosserat_curve",
+  "meshpy.examples",
   "meshpy.four_c",
   "meshpy.geometric_search",
   "meshpy.mesh_creation_functions",
-  "meshpy_testing",
+  "meshpy.testing",
```
isteinbrecher, line 14: @davidrudlstorfer do you habe a preferred ay of doing this now? I think this will change anyways with #159

pyproject.toml
```diff
@@ -35,6 +37,8 @@ dependencies = [
   "black==22.12.0",
   "Cython",
   "geomdl",
+  "ipykernel",
+  "notebook",
```
isteinbrecher, line 41: I am currently fine with having this in the standard installation. Alternatively we could add the optional `[examples] installation, but I am not sure if this is necessary at this point.

.binder/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+ARG BASE_IMAGE_TAG=latest
+FROM ghcr.io/pyvista/pyvista:$BASE_IMAGE_TAG
```
davidrudlstorfer, line 29: just a question to better understand the binder stuff - we use the pyvista iamge so we can visualize everything in the browser?

.binder/start
```diff
@@ -0,0 +1,30 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+export PYVISTA_TRAME_SERVER_PROXY_PREFIX="$JUPYTERHUB_SERVICE_PREFIX/proxy/"
```
davidrudlstorfer, line 29: again - just a question to understand it - what is this doing?

examples/utils/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    """Print a matrix to the console."""
+    print(f"{name}:\n{np.round(matrix,decimals=10)}")
+
+
+def print_rotation_matrix(name, rotation):
+    """Print the rotation matrix to the console."""
+    print_matrix(name, rotation.get_rotation_matrix())
+
+
+def add_cube_plot(plotter, row, col, rotation, text, *, plot_outlines=True):
```
davidrudlstorfer, line 59: This function can probably be not reused by other examples - I think it would make sense to create a "general" utils file which contains utils used by all examples and a file for each example. This plotting function could move into "example_1_utils.py" - this way the files "belong together"

tests/test_examples.py
```diff
@@ -0,0 +1,49 @@
...
+    ],
+)
+def test_notebooks(notebook_path):
+    """Parameterized test case for multiple Jupyter notebooks.
+
+    The notebook is run and it is checked that it runs through without
+    any errors/assertions.
+    """
+    with testbook(notebook_path, execute=True) as tb:
+        pass
```
davidrudlstorfer, line 49: Really really nice!

examples/utils/utils.py
davidrudlstorfer: I would keep it this way for now and open an issue with stuff like this. Maybe we can find a more elegant solution to not need this at all.

And if it's only used once I do not see an advantage to define it somewhere else and import everything back again.

examples/utils/utils.py
davidrudlstorfer: Same as above - I would keep it for now

meshpy/mesh.py
davidrudlstorfer: Also from me :D see above

pyproject.toml
davidrudlstorfer: I would also keep this as it probably will be changed anyways in #159

pyproject.toml
davidrudlstorfer: I would also keep this - for me it failed without this package on Ubuntu and in my opinion the examples should work out of the box

davidrudlstorfer (commented):
@isteinbrecher i just have two questions regarding binder and one proposal regarding the example utils

I would leave the remaining stuff in the current state as it is really really nice already!

Thanks for tackling this issue - it could not have turned out better! Regarding the example - I am no expert regarding the finite rotations so I leave the content up to you - I read through everything and have no comments.

Thanks again!

davidrudlstorfer:
And one thing I forgot - if you like we can directly add the badges to the readme

Maybe a new section (something along the lines) "Interactive examples"

And just list all examples with the badge

> ### Interactive examples
> - Finite rotation framework [Binder](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples?labpath=examples%2Fexample_1_finite_rotations.ipynb)
> 
> You can also interactively test the entire MeshPy Framework directly from your browser here [Binder](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples)
>

examples/utils/utils.py
isteinbrecher: I decided to add them to the `utility` file already now, so we can keep a better track where these functions are used.

.binder/Dockerfile
isteinbrecher: Yes, I assume it would also work with the default mybinder image, but I just copied how pyvista suggests to do it.

.binder/start
isteinbrecher: To be honest, I don't know, just copied it from pyvista. I added a note where the code originally comes from.

examples/utils/utils.py
isteinbrecher: Good point, I adapted that.

isteinbrecher:
@davidrudlstorfer very good point, I missed that, I added the links to the `README.md`. Now the examples should be easily accessible. I also added a little more content to the second example. I would now try to get this into a stage where we can merge it and then we can continuously add more content to the examples.

README.md
```diff
@@ -67,7 +67,15 @@ MeshPy is developed at the [Institute for Mathematics and Computer-Based Simulat
...
 
-Basic tutorials can be found in the directory `tutorial/`.
+MeshPy provides example notebooks to showcase its core features and functionality.
+The examples can be found in the `examples/` directory.
+They can be run locally or directly tested from your browser via the following links:
+
+- Example 1: **Finite rotation framework** [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples?labpath=examples%2Fexample_1_finite_rotations.ipynb)
+- Example 2: **Core mesh generation functions** [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples?labpath=examples%2Fexample_2_core_mesh_generation_functions.ipynb)
+
+You can also interactively test the entire MeshPy framework directly from your browser here [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/isteinbrecher/meshpy/add-examples)
```
davidrudlstorfer, line 77: Just a note - at the end of this PR we should update the links to the main repo rather than the fork

davidrudlstorfer (dismissed):

isteinbrecher:
Thanks for the review @davidrudlstorfer . I fixed the links and combined the commits. Once the pipeline passes I will merge this.

PR #205, opened 2025-01-20
Add space-time functionality

isteinbrecher:
Add the possibility to add space-time meshes to MeshPy. For now we support 2D surfaces in 4D space, i.e., 3D beams in space + time. A 4C interface is provided, although not yet supported within 4C.

As it turns out, adding large lists of nodes / elements to a mesh via the `.add` method can be very slow (similar reason as discussed in #138). To avoid this we don't directly use the `.add` method here, but instead directly set the corresponding class variables. I added a performance test which should allow to track that going further. Once this is merged I will open an issue documenting this and providing the next steps.

Depends on #202


meshpy/space_time/beam_to_space_time.py
```diff
@@ -0,0 +1,378 @@
...
+        Either a fixed spatial Mesh object or a function that returns the
+        spatial mesh for a given time. If this is a generator, the topology
+        of the mesh at the initial time is chosen for all times, only the
+        positions and rotations are updated.
+    time_duration:
+        Total time increment to be solved with the space-time mesh
+    number_of_elements_in_time:
+        Number of elements in time direction
+    time_start:
+        Starting time for the space-time mesh. Can be used to create time slaps.
```
davidrudlstorfer, line 159:
```suggestion
        Starting time for the space-time mesh. Can be used to create time steps.
```
?

meshpy/space_time/beam_to_space_time.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+
+    # Get the "reference" spatial mesh
+    if callable(mesh_space_or_generator):
+        mesh_space_reference = mesh_space_or_generator(time_start)
+    else:
+        mesh_space_reference = mesh_space_or_generator
+
+    # Perform some sanity checks
+    valid_element_types = [Beam3rLine2Line2, Beam3rHerm2Line3]
```
davidrudlstorfer, line 178: Would it make sense to define these at the top of the file, like

`VALID_ELEMENT_TYPES_SPACE_TIME = ...`?

meshpy/space_time/beam_to_space_time.py
```diff
@@ -0,0 +1,378 @@
...
+    element_type = element_types[0]
+    if element_type not in valid_element_types:
+        raise ValueError(f"Got unexpected element type {element_type}")
+
+    # Calculate global mesh properties
+    number_of_nodes_in_space = len(mesh_space_reference.nodes)
+    number_of_elements_in_space = len(mesh_space_reference.elements)
+    is_linear = element_type == Beam3rLine2Line2
+    space_time_element: Union[Type[SpaceTimeElementQuad4], Type[SpaceTimeElementQuad9]]
+    if is_linear:
```
davidrudlstorfer, line 195: You could simply use
```suggestion
    if element_type == Beam2rLine2Line2:
```

here. This way the hard to read variable `is_linear` is gone and one directly sees what's happening here

You can also then use the other element in the else branch and raise an error if the element is another type

tests/test_space_time.py
```diff
@@ -0,0 +1,238 @@
...
+    mesh = Mesh()
+    test_performance.time_function(
+        "space_time_create_mesh_in_space",
+        create_beam_mesh_line,
+        args=[mesh, beam_type, mat, [0, 0, 0], [1, 0, 0]],
+        kwargs={"n_el": 100},
+    )
+
+    # Create the beam mesh in time
+    _, _ = test_performance.time_function(
```
davidrudlstorfer, line 231: why do you have return values here which are not saved and above you simply have no return values?

davidrudlstorfer (approved):
Nice addition! I left a few comments regarding code style - if you like you can add some of them or simply merge.

It would also be nice if you could finish the docstrings/typehints. Often the return value is missing or the arguments/return values are not described in the docstring.

meshpy/space_time/beam_to_space_time.py
```diff
@@ -0,0 +1,378 @@
...
+
+class SpaceTimeElement(VolumeElement):
+    """A general beam space-time surface element."""
+
+    def __init__(
+        self,
+        nodes,
+        *,
+        dat_pre_nodes="DUMMY",
+        dat_post_nodes="POST_DUMMY",
```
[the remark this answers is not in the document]
isteinbrecher: No, they will remain, but have to be renamed to the names used in 4C. Similar to `BEAM3R HERM2LINE3` and `TRIADS`.

meshpy/space_time/beam_to_space_time.py
isteinbrecher: Thanks for pointing that out, might have very well been a typo, but in this case it is actually correct. _time slaps_ in space time formulations are space time meshes that are evaluated successively (it is kind of a time stepping scheme, but the used term is slap).

meshpy/space_time/beam_to_space_time.py
isteinbrecher: Good point, I restructured the way types are defined in the space time function, now everything should hopefully be much clearer.

tests/test_space_time.py
isteinbrecher: Good catch!

meshpy/space_time/beam_to_space_time.py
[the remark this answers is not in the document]
davidrudlstorfer: So do these need to be arguments or can they simply be hardcoded?

meshpy/space_time/beam_to_space_time.py
davidrudlstorfer: Thanks for the explanation - again something learned

davidrudlstorfer (approved):

PR #207, opened 2025-01-24
Add transport and thermo element sections for full mesh import

maxfirmbach:
Currently working quite a bit with thermo and scatra elements, thus removing both from the section exclusion and adding the relevant read mechanism for the respective element lines.

### DIFF
--- a/meshpy/inputfile.py
+++ b/meshpy/inputfile.py
@@ -309,9 +309,7 @@ class InputFile(Mesh):
     skip_sections = [
         "ALE ELEMENTS",
         "LUBRICATION ELEMENTS",
-        "TRANSPORT ELEMENTS",
         "TRANSPORT2 ELEMENTS",
-        "THERMO ELEMENTS",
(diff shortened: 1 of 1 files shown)


meshpy/inputfile.py
```diff
@@ -593,6 +591,18 @@ def add_line(self_list, line):
...
+                    if mpy.import_mesh_full:
+                        self.elements.append(Element.from_dat(line))
+                    else:
+                        add_line(self.dat_elements, line)
+            elif section_name == "TRANSPORT ELEMENTS":
+                for line in section_data_comment:
+                    if mpy.import_mesh_full:
+                        self.elements.append(Element.from_dat(line))
+                    else:
+                        add_line(self.dat_elements, line)
```
isteinbrecher, line 605: Are you sure this works? Won't these elements end up in the `STRUCTURE ELEMENTS` section?

isteinbrecher: Maybe a test case would help to see that 😉

PR #210, opened 2025-01-28
Update website deployment

davidrudlstorfer:


davidrudlstorfer:
As a first step I would like to merge this @isteinbrecher 

As i found out yesterday the `workflow_run` rule is not allowed to be combined with a `push` condition

If you are ok with this we could merge this and check if it works on main

That would be a first step for the coverage - as this also only works in upcoming PR's if the correct `workflow_run` rule is on the main branch.

davidrudlstorfer:
One thing I actually cannot wrap my head around is that the website always gets deployed on the main branch once a test suite run is sucessfull within a PR - https://github.com/imcs-compsim/meshpy/deployments

This was the behaviour from the beginning - it is very weird

isteinbrecher (dismissed):

davidrudlstorfer:
@isteinbrecher I've removed the last commit - the reason why the website deploys in each PR is that the old workflow is on main - I'll merge this state and check if it works

PR #212, opened 2025-01-28
Deploy coverage report

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher slowly but surely I get closer to a working version

I would need one thing from you - could you activate `Settings` -> `Actions` -> `General` -> `Allow actions by Marketplace verified creators`

We need the action to download artifacts from other workflows (dawidd6/action-download-artifact@v8). That's the same approach as in 4C - but currently I am not able to use that Github Action

(And i am not able to activate that setting)

Thanks!

davidrudlstorfer:
@isteinbrecher good news - the basic approach works!

see

https://imcs-compsim.github.io/meshpy/coverage-report/

and 

https://imcs-compsim.github.io/meshpy/

Once I can use the other Github actions I can finalize this

isteinbrecher:
Sure, I can do that. Somehow they are taken from the group and can not be set for the individual repositories.

Is it ok to simply specify the action like in the screenshot, or do we have to explicitly set "Allow actions by Marketplace verified creators"?

image

davidrudlstorfer:
@isteinbrecher this would be ready for review and hopefully (knocking 3x times on my desk) this should work as planned on main

I now separated the coverage stuff into a separate action - this way the coverage report/badge can very easily be moved to another testing job

We now deploy our standard website under 
https://imcs-compsim.github.io/meshpy/

Then the coverage report under
https://imcs-compsim.github.io/meshpy/coverage-report

and the coverage badge under
https://imcs-compsim.github.io/meshpy/coverage-badge

For the badge we need a simple html website/index file which just includes the coverage .svg image

Let me know what you think - once you are ok with the changes I squash the commits into one and would love to merge

In a separate PR I'll add the coverage badge to the readme

davidrudlstorfer:
(Note that the link to the coverage badge currently is not working because the artifacts are only pulled from the main branch - and we currently do not create the coverage badge "website" on the main branch)

isteinbrecher (approved):
Looks good to me, I guess there is only one way to find out if it works 🚀

PR #213, opened 2025-01-28
Improve performance of Mesh.add method for lists

isteinbrecher:
The common way to add objects to a `Mesh` is to use the `add` method. This allows to check that no duplicate entries exist in the mesh (if this check is really necessary can be debated). Up until now when we added a list of items, we added each element on its own via the `add` method in a recursive manner. This works fine as long as there are not too many entries in the list.

However, it a list with a large number of entries is added, we run into the problem, that for each element in the list, we check if this element is already in the mesh, then extend the corresponding data structure in the mesh and then do the same thing for the next item. This scales very badly due to the repeated checking for duplicate entries. A better approach is to add the whole list to the mesh and then check in the end, if we have double elements. By doing so we only have to check for duplicate entries once and the performance is much better.

This allows do directly use the `add` method in the creation of the space time meshes, instead of the previous hack. There we can also nicely see the performance impact, as the space time performance test with using the old `add` method did not finish after 5 minutes on my machine, but with the new method, we achieve basically the same performance as with the previous hack.


meshpy/mesh.py
```diff
@@ -76,6 +76,32 @@ def __init__(self):
...
+        elif isinstance(item, Node):
+            return Node
+        elif isinstance(item, Element):
+            return Element
+        elif isinstance(item, GeometrySetBase):
+            return GeometrySetBase
+        elif isinstance(item, GeometryName):
+            return GeometryName
+        else:
+            return type(item)
```
davidrudlstorfer, line 103: Could we simplify this construct?

For example

```suggestion
        for cls in (Mesh, Function, BoundaryConditionBase, Material, Node, Element, GeometrySetBase, GeometryName):
            if isinstance(item, cls):
                return cls
        return type(item)
```

meshpy/mesh.py
```diff
@@ -90,25 +116,25 @@ def add(self, *args, **kwargs):
...
+                elif base_type is GeometryName:
                     self.add_geometry_name(add_item, **kwargs)
-                elif isinstance(add_item, list):
-                    for item in add_item:
-                        self.add(item, **kwargs)
+                elif base_type is list:
+                    self.add_list(add_item, **kwargs)
                 else:
                     raise (
                         TypeError(
```
davidrudlstorfer, line 140: Again maybe directly simplify this, for example

```
method_map = {
    Mesh: self.add_mesh,
    Function: self.add_function,
    BoundaryConditionBase: self.add_bc,
    Material: self.add_material,
    Node: self.add_node,
    Element: self.add_element,
    GeometrySetBase: self.add_geometry_set,
    GeometryName: self.add_geometry_name,
    list: self.add_list,
}

if base_type in method_map:
    method_map[base_type](add_item, **kwargs)
else:
    raise TypeError(f'No Mesh.add case implemented for type: "{type(add_item)}"!')
```

meshpy/mesh.py
```diff
@@ -181,6 +207,46 @@ def add_geometry_name(self, geometry_name):
         for key in keys:
             self.add(geometry_name[key])
 
+    def add_list(self, add_list, **kwargs):
```
davidrudlstorfer, line 210: Optionally you could directly add the type hints if we add/touch functions throughout the code base

meshpy/mesh.py
```diff
@@ -181,6 +207,46 @@ def add_geometry_name(self, geometry_name):
...
+        """Add a list of items to this mesh.
+
+        In the special case of nodes or elements we add the whole list
+        at once. This avoids a check for duplicate entries for every
+        addition, which scales very badly. By doing it this way we only
+        check the final list for duplicate entries which is much more
+        performant.
+
+        For all other types of items, we add each element individually.
```
davidrudlstorfer, line 219: And optionally we could also directly add the args here in the docstring

meshpy/mesh.py
```diff
@@ -181,6 +207,46 @@ def add_geometry_name(self, geometry_name):
...
+        """
+
+        types = {self.get_base_mesh_item_type(item) for item in add_list}
```
davidrudlstorfer, line 222: Why do we create a dictionary here? Would a simple list also suffice?

davidrudlstorfer (commented):
Thanks for the nice changes!

I have a few small suggestions

meshpy/mesh.py
isteinbrecher: This syntax creates a `set` inplace. I think for a dict one would have to write something like `{key: value for key, value in stuff}`

meshpy/mesh.py
isteinbrecher: Also, very nice suggestion!

meshpy/mesh.py
isteinbrecher: Should be done, I hope this aligns with the google style.

davidrudlstorfer (approved):
Nice! Thanks for looking into all comments I left

PR #215, opened 2025-01-28
Update badges

davidrudlstorfer:


davidrudlstorfer:
Everything works as expected :partying_face: 

Last step is to add the different OS badges and the documentation - with the current infrastructure this should be easily doable

isteinbrecher (approved):

PR #217, opened 2025-01-29
Add MeshPy API documentation

davidrudlstorfer:
This PR adds the first version of the MeshPy API documentation which is based on the docstrings. The API documentation is created with https://pdoc.dev/

The structure of the documentation now only includes the core MeshPy module - I intentionally did not add the other submodules as the MeshPy modularity will change with #159 

The API documentation looks like this

[docs.zip](https://github.com/user-attachments/files/18584906/docs.zip)

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  build:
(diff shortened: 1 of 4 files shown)


isteinbrecher (approved):
Thanks a lot @davidrudlstorfer this looks really nice!

PR #220, opened 2025-01-29
Fix documentation deployment 2

davidrudlstorfer:
Problem with the old fix was that the `workflow_run` only works for one workflow. If two prior workflows are added the rule only waits on the successful finish of the first one.

~Therefore, moving back the documentation build to the website creation - which necessitates the creation of the full python venv~

EDIT: as proposed offline - a serial approach makes more sense - now after a successfull test suite run on main the documentation is built. Once this is finished, the website is deployed

### DIFF
--- a/.github/actions/code_check/action.yml
+++ b/.github/actions/code_check/action.yml
@@ -7,7 +7,7 @@ runs:
       shell: bash
       run: |
         cd ${GITHUB_WORKSPACE}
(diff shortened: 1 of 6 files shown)


isteinbrecher (approved):
LGTM. From my point of view we can merge this. The pipeline on main will fail, with the current changes I am OK with that. I will leave that up to you @davidrudlstorfer

PR #221, opened 2025-01-30
Fix import problems

davidrudlstorfer:
When MeshPy is used as a submodule (i.e., specifically not at the toplevel but in a src layout) the imports do not resolve and MeshPy does not work at all

This temporarily fixes these problems

**Will be fixed soon with #159 - this is a quick fix to enable the usage of the current main branch as a submodule**

### DIFF
--- a/meshpy/four_c/dbc_monitor.py
+++ b/meshpy/four_c/dbc_monitor.py
@@ -32,10 +32,12 @@ from typing import Optional
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


isteinbrecher (approved):
LGTM. From my point of view we can merge this. The pipeline on main will fail, with the current changes I am OK with that. I will leave that up to you @davidrudlstorfer

PR #222, opened 2025-02-05
Add custom position for nodes within meshing strategy

knarfnitram:
For the set-up of beam to beam contact problems, it can be beneficial to manipulate the element size of beams.
As an example think of two beams within a plane, which are crossing each other(90° angle). If the number of elements for meshing the beams are chosen even, two nodes of the beams have the same coordinate. 
This MR adds the option to adapt the meshing strategy to a custom distribution of points.
The parameter `custom_node_distribution` provides a list of nodes within the interval `[0,1]`, where the nodes will then be placed.


mayrmt:
I still find it weird, that beam contact cannot handle the case of two nodes contacting each other and that we need to avoid that by fiddling with the mesh.

knarfnitram:
> I still find it weird, that beam contact cannot handle the case of two nodes contacting each other and that we need to avoid that by fiddling with the mesh.

Thank you for your comment, however I think that this MR is the wrong place to discuss warnings of 4C.

mayrmt:
@knarfnitram Sure, you are right! Still, I wonder if we should merge stuff to meshpy, that does not fix, but just to circumvent an issue with the possible need for user interaction.

knarfnitram:
I totally agree that we should only add features to MeshPy that make sense. Since MeshPy consider itself as a _general purpose 3D beam finite element input generator_, I think that the possibility to adapt the element size of a 1D grid arbitrary and not only stick to the same element length is a useful extension - especially when you consider how few changes are necessary for it.

davidrudlstorfer:
@knarfnitram I did not look into the changes in detail but one question regarding the tolerances:

Before your changes the output was "exact" - meaning that the values were `1.0`, `2.0`, ... 
Now the values are no longer "exact", i.e., `1.00..001`, `2.000...001`

Is this change connected to any of your changes? Because on main these rounding errors are not present? And in my opinion it would be nice if the behaviour stays like this for the output of .dat files

knarfnitram:
> Maybe this is connected with this
> 
> https://stackoverflow.com/questions/14716981/numpy-linspace-rounding-error
> 
> Maybe use `np.arange` to precisely set the interval?

I tried to use `np.arange` instead of `np.linspace`, but unfortunately I still have to adapt the tolerances of the same three test cases.

isteinbrecher:
@knarfnitram thanks for proposing this long overdue feature! This is rarely needed, but allows basically total freedom in the element generation. Since this is basically the core functionality in MeshPy, I will have a closer look once I am back at work next week.

A couple of points that come to my mind already:
- The options `n_el`, `l_el` and the new one (we have to find a good naming) should be mutually exclusive. (You can have a look at the `create_helix` function, there it is nicely checked that the provided arguments are sufficient)
- Documentation has to be adapted in a lot ob places.
- Add a check that the positions given are increasing

@mayrmt I agree that we should tackle the problem at its root and figure out a way to allow this contact case in 4C. (But I am glad that this "workaround" adds this missing functionality in 4C.

davidrudlstorfer:
@knarfnitram if you rebase on latest main the tests pass again

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -101,26 +102,48 @@ def create_beam_mesh_function(
...
+                f"Provided values are min={min(custom_node_distribution)} and max={min(custom_node_distribution)}"
+            )
+
+        node_point_of_element = custom_node_distribution
+        n_el = len(custom_node_distribution) - 1
     else:
         raise ValueError("The parameters n_el and l_el are mutually exclusive")
 
+    if node_point_of_element is None:
+        node_point_of_element = np.linspace(interval[0], interval[1], n_el + 1)
```
isteinbrecher, line 145: I like this approach, just a little restructuring suggestion:
```python
# check that the input parameters are correct, i.e., check for mutually exclusive parameters
...

# Go through all the cases (n_el, l_el and the new parameter) and let them create the node_point_of_element array (maybe with a better name). For the existing cases of n_el and l_el we can simply use the same mechanism that is currently used in the loop below.
...

# Let the rest of the function basically unchanged with the exception of the loop parameters at the end
...
```

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -212,8 +235,8 @@ def get_relative_twist(rotation_node, rotation_function):
```
isteinbrecher, line 235: The loop here would then be something like:

```python
for i_el in range(len(node_point_of_element) - 1):
    function = function_generator(node_point_of_element[i_el], node_point_of_element[i_el+1], ...)
```

isteinbrecher (commented):
Thanks @knarfnitram for the provided changes. I left some comments. Can we try to first get the implementation finalised and then adapt possible failing test cases?

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -49,6 +49,7 @@ def create_beam_mesh_function(
     start_node=None,
     end_node=None,
     vtk_cell_data=None,
+    custom_node_distribution=None,
```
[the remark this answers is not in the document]
knarfnitram: ok, i now see what your problem is with that. How about `element_node_positions_for_mesh_generation` ?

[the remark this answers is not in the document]
isteinbrecher: Let’s just go with `node_positions_of_elements`, if we want tonguing it it should be no problem later on.

tests/test_meshpy.py
```diff
@@ -1914,3 +1915,60 @@ def test_meshpy_display_pyvista(reference_file_directory):
...
+    distribution = np.append(distribution, interval[1])
+
+    # add a beam line with three elements
+    create_beam_mesh_line(
+        mesh,
+        Beam3rHerm2Line3,
+        material,
+        [1, 2, 0],
+        [4, 2, 0],
+        custom_node_distribution=distribution,
```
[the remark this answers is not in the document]
knarfnitram: Yes, I think they are not working yet.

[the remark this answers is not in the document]
isteinbrecher: With the approach I outlined here https://github.com/imcs-compsim/meshpy/pull/222#discussion_r1950641939 I think that should work. We just have to make sure to scale the positions given in `node_positions_of_elements` with the provided `interval`.

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -101,26 +102,48 @@ def create_beam_mesh_function(
         With this argument, a vtk cell data can be set for the elements
         created within this function. This can be used to check which
         elements are created by which function.
+    custom_node_distribution: [double]
+        List with nodes within the interval [0,1] that presicribes the position of the nodes for the element.
```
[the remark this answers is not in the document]
knarfnitram: I have updated the description.

tests/test_meshpy.py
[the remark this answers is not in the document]
knarfnitram: Thanks for the hint, it worked. I have provided test cases for 
- `create_beam_mesh_helix`
- `create_beam_mesh_curve`
- `create_beam_mesh_arc_segment_2d`

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -109,17 +115,62 @@ def create_beam_mesh_function(
...
+        pass
+
+    # Calculate the node_positions_of_elements
+    if node_positions_of_elements is None:
+        if len(interval) == 2:
+            node_positions_of_elements = np.linspace(interval[0], interval[1], n_el + 1)
+        else:
+            raise ValueError(
+                f"Please provide an interval with len(interval) == 2, len(interval)={len(interval)}."
+            )
```
isteinbrecher: I think this approach is a little bit confusing, I pushed a commit that should improve that.

tests/test_four_c_simulation.py
```diff
@@ -842,6 +842,7 @@ def test_four_c_simulation_dirichlet_boundary_to_neumann_boundary_with_all_value
     assert_results_equal(
         get_corresponding_reference_file_path(additional_identifier="dirichlet"),
         initial_simulation,
+        atol=1e-10,
```
isteinbrecher, line 845: Should not be needed with the proposed approach.

tests/test_mesh_creation_functions.py
```diff
@@ -211,6 +211,46 @@ def test_mesh_creation_functions_arc_segment_2d(
...
+    create_beam_mesh_arc_segment_2d(
+        input_file,
+        Beam3rHerm2Line3,
+        mat,
+        [1.0, 2.0, 0.0],
+        1.5,
+        np.pi * 0.25,
+        np.pi * (1.0 + 1.0 / 3.0),
+        node_positions_of_elements=[0.0, 0.2, 0.4, 0.6, 0.8, 1.0],
+    )
```
isteinbrecher, line 236: The equally spaced case there is not needed, as all the functionality is used (in a more general way) with the next mesh creation.

tests/test_mesh_creation_functions.py
```diff
@@ -211,6 +211,46 @@ def test_mesh_creation_functions_arc_segment_2d(
     assert_results_equal(get_corresponding_reference_file_path(), input_file)
 
 
+def test_mesh_creation_functions_arc_segment_2d_prescribed_node_positions_of_elements(
```
isteinbrecher, line 214: I would rename this to `test_mesh_creation_functions_node_positions_of_elements_option`. Then we can create a line, an arc and an helix here, and add the checks.

tests/test_mesh_creation_functions.py
```diff
@@ -587,12 +627,47 @@ def test_mesh_creation_functions_element_length_option(
             n_el=1,
             l_el=l_el,
         )
+    with pytest.raises(ValueError):
+        mesh = Mesh()
+        # This should raise an error because node_positions_of_elements can not be used with l_el
+        return create_beam_mesh_function(
+            mesh, l_el=1, node_positions_of_elements=[0.0, 0.5, 1.0]
+        )
```
isteinbrecher, line 635: Really nice that you added the error checks!

Let's move all the checks (also the one that already existed) to an own test `test_mesh_creation_functions_argument_checks` since this test is intended to check the `l_el` option.

We should also add an additional check:
```
    with pytest.raises(ValueError):
        mesh = Mesh()
        return create_beam_mesh_function(mesh, n_el=1, l_el=2.0)
```

tests/test_mesh_creation_functions.py
```diff
@@ -630,7 +705,7 @@ def test_mesh_creation_functions_curve_3d_helix(
         coordinates_mathematica,
         get_nodal_coordinates(input_file.nodes),
         rtol=mpy.eps_pos,
-        atol=1e-14,
+        atol=1e-13,
```
isteinbrecher, line 708: Should not be needed.

tests/test_mesh_creation_functions.py
```diff
@@ -211,6 +211,46 @@ def test_mesh_creation_functions_arc_segment_2d(
...
+    # Create a 2d segments with different element sizes
+    create_beam_mesh_arc_segment_2d(
+        input_file,
+        Beam3rHerm2Line3,
+        mat,
+        [1.0, 2.0, 0.0],
+        2.0,
+        np.pi * 0.25,
+        np.pi * (1.0 + 1.0 / 3.0),
+        node_positions_of_elements=[0, 0.5, 0.75, 0.875, 1.0],
```
isteinbrecher, line 247:
```suggestion
        node_positions_of_elements=[0, 1.0/ 3.0, 1.0],
```
Should be sufficient.

tests/test_mesh_creation_functions.py
```diff
@@ -919,6 +1005,35 @@ def test_mesh_creation_functions_helix_rotation_offset(
...
+    helix_set = create_beam_mesh_helix(
+        input_file,
+        Beam3rHerm2Line3,
+        mat,
+        [0.0, 0.0, 1.0],
+        [0.0, 0.0, 0.0],
+        [2.0, 0.0, 0.0],
+        height_helix=10.0,
+        turns=2.5 / np.pi,
+        node_positions_of_elements=[0, 0.1, 0.5, 0.9, 1],
```
isteinbrecher, line 1029: Same as above, `[0, 1.0/3.0,1]`

tests/test_meshpy.py
```diff
@@ -512,6 +512,7 @@ def test_meshpy_unique_ordering_of_get_all_nodes_for_line_condition(
     assert_results_equal(
         get_corresponding_reference_file_path(),
         input_file,
+        atol=1e-10,
```
isteinbrecher, line 515: Should not be needed.

tests/test_meshpy.py
```diff
@@ -1915,3 +1916,51 @@ def test_meshpy_display_pyvista(reference_file_directory):
...
+    )
+
+    # add a beam line with specified element distribution
+    create_beam_mesh_line(
+        mesh,
+        Beam3rHerm2Line3,
+        material,
+        [1, 2, 0],
+        [4, 2, 0],
+        node_positions_of_elements=[0, 0.1, 0.3, 0.5, 0.7, 0.9, 1.0],
```
isteinbrecher, line 1959: Sam as above.

tests/test_meshpy.py
```diff
@@ -1915,3 +1916,51 @@ def test_meshpy_display_pyvista(reference_file_directory):
...
+    # Set up an equal distribution points.
+    equal_distribution = np.linspace(interval[0], interval[1], n_points)
+
+    # Result is similar to normal mesh line.
+    create_beam_mesh_line(
+        mesh,
+        Beam3rHerm2Line3,
+        material,
+        [1, 0, 0],
+        [4, 0, 0],
```
isteinbrecher, line 1948: This case is not needed.

tests/test_meshpy.py
```diff
@@ -1915,3 +1916,51 @@ def test_meshpy_display_pyvista(reference_file_directory):
     mpy.import_mesh_full = True
     mesh = create_beam_to_solid_conditions_model(reference_file_directory)
     _ = mesh.display_pyvista(resolution=3)
+
+
+def test_create_beam_mesh_line_custom_element_size(
+    assert_results_equal, get_corresponding_reference_file_path
+):
+    """Checks if a custom node distribution can be forwarded to the beam
+    creation."""
```
isteinbrecher, line 1925:
```suggestion
    """Checks if a custom node distribution can be forwarded to the create beam line function."""
```

tests/test_mesh_creation_functions.py
```diff
@@ -752,6 +827,17 @@ def rotation(t):
     )
     input_file.add(sin_set)
 
+    # extend test case with different meshing strategy
+    create_beam_mesh_curve(
+        input_file,
+        Beam3rHerm2Line3,
+        mat,
+        curve,
+        [1.0, 2.0],
```
isteinbrecher, line 836:
```suggestion
        [1.0, 2.5],
```
Then we also test that the inveral can have a length other than 1.

tests/test_mesh_creation_functions.py
knarfnitram: The combination of `n_el=1` and `l_el=2.0` is already checked.

knarfnitram:
@isteinbrecher - Thanks for the comments, I have added them.

tests/test_mesh_creation_functions.py
```diff
@@ -607,18 +678,53 @@ def test_mesh_creation_functions_element_length_option(
         create_beam_mesh_line(
             mesh,
             Beam3rHerm2Line3,
-            mat,
+            MaterialReissner(radius=0.1),
             [1.0, 2.0, 0.0],
             [3.0, 4.0, 6.0],
             n_el=1,
             l_el=l_el,
```
isteinbrecher, line 685:
```suggestion
            l_el=1.5,
```
and let's remove the previous definition of `l_el`.

tests/test_mesh_creation_functions.py
```diff
@@ -656,7 +762,6 @@ def test_mesh_creation_functions_curve_3d_helix(
         coordinates_mathematica,
         get_nodal_coordinates(input_file.nodes),
         rtol=mpy.eps_pos,
-        atol=1e-14,
```
isteinbrecher, line 659: Why do we need to remove this?

tests/test_mesh_creation_functions.py
```diff
@@ -607,18 +678,53 @@ def test_mesh_creation_functions_element_length_option(
...
             [1.0, 2.0, 0.0],
             [3.0, 4.0, 6.0],
             n_el=1,
             l_el=l_el,
         )
+    with pytest.raises(ValueError):
+        mesh = Mesh()
+        # This should raise an error because node_positions_of_elements can not be used with l_el.
+        return create_beam_mesh_function(
+            mesh, l_el=1, node_positions_of_elements=[0.0, 0.5, 1.0]
```
isteinbrecher, line 691: The `return` can be removed here (and later on). One more thing, can we maybe add a special assertion check here? `pytest.raises` supports easy mechanisms to check the error string, to be sure the correct error is raised. Something like
```python
    with pytest.raises(
        ValueError,
        match='The arguments "n_el", "l_el" and "node_positions_of_elements" are mutually exclusive',
    ):
```

tests/test_mesh_creation_functions.py
```diff
@@ -607,18 +678,53 @@ def test_mesh_creation_functions_element_length_option(
         create_beam_mesh_line(
             mesh,
             Beam3rHerm2Line3,
-            mat,
+            MaterialReissner(radius=0.1),
```
isteinbrecher, line 681: I think it should be fine if you also remove the radius argument here.

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -101,25 +102,66 @@ def create_beam_mesh_function(
         With this argument, a vtk cell data can be set for the elements
         created within this function. This can be used to check which
         elements are created by which function.
-
+    node_positions_of_elements: [double]
+        These values specify where elements start and end, additional internal nodes (such as midpoints in higher-order elements) will be placed automatically.
+        The array must be ordered ascending.
```
isteinbrecher, line 107: Can we move this description up, so the ordering of the doc string matches the ordering of the keyword arguments?

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -101,25 +102,66 @@ def create_beam_mesh_function(
...
+            if interval_length is None:
+                raise ValueError(
+                    'The parameter "l_el" requires "interval_length" to be set.'
+                )
+            n_el = max([1, round(interval_length / l_el)])
+        interval_node_positions_of_elements = [
+            interval[0] + i_node * (interval[1] - interval[0]) / n_el
+            for i_node in range(n_el + 1)
+        ]
+    # A list of for the element node positions was provided
```
isteinbrecher, line 142:
```suggestion
    # A list for the element node positions was provided
```

isteinbrecher (commented):
Thanks for the changes @knarfnitram , I left some final comments. Can you please add them and then squash everything into a single or few commits?

tests/test_mesh_creation_functions.py
knarfnitram: That wasn't intentional.

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -101,25 +105,63 @@ def create_beam_mesh_function(
         With this argument, a vtk cell data can be set for the elements
         created within this function. This can be used to check which
         elements are created by which function.
-
```
isteinbrecher, line 104: Don't remove this line

meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -84,6 +85,9 @@ def create_beam_mesh_function(
         created.
     interval_length:
         Total length of the interval. Is required when the option l_el is given.
+    node_positions_of_elements: [double]
+        These values specify where elements start and end, additional internal nodes (such as midpoints in higher-order elements) will be placed automatically.
+        The array must be ordered ascending.
```
isteinbrecher, line 90:
```suggestion
    node_positions_of_elements: [double]
        A list of normalized positions (within [0,1] and in ascending order) that define the
        boundaries of beam elements along the created curve. The given values will be
        mapped to the actual `interval` given as a parameter. These values specify where
        elements start and end, additional internal nodes (such as midpoints in higher-order
        elements) may be placed automatically.
```

tests/test_mesh_creation_functions.py
```diff
@@ -542,7 +609,6 @@ def test_mesh_creation_functions_element_length_option(
     mat = MaterialReissner(radius=0.1)
 
     l_el = 1.5
-
```
isteinbrecher, line 545: Not needed white space change.

tests/test_mesh_creation_functions.py
```diff
@@ -28,6 +28,8 @@
 # SOFTWARE.
 """This script is used to test the mesh creation functions."""
 
+import re
+
```
isteinbrecher, line 32: I think this import is not needed.

tests/test_mesh_creation_functions.py
```diff
@@ -599,25 +665,79 @@ def test_mesh_creation_functions_element_length_option(
...
+        mesh = Mesh()
+        # This should raise an error because the interval [0,1] is violated.
+        create_beam_mesh_function(mesh, node_positions_of_elements=[0.0, 1.0, 2.0])
+
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "The given node_positions_of_elements must be in ascending order. Got [0.0, 0.2, 0.1, 1.0]"
+        ),
+    ):
```
knarfnitram, line 737: @isteinbrecher  - I added the `import re` on purpose, since the string comparison is failing here. I guess the problem for the not matching string expression is arising from the brackets `[]`.

isteinbrecher: Sorry, I looked through all calls to `pytest.raises` but missed that one, then no problem with importing `re`!

isteinbrecher (approved):
Thanks for the effort @knarfnitram, this adds a really usefully extension to the core functionality MeshPy. I added a small fix of a typo that I introduced myself, once the pipeline runs through, we are ready to merge this 🚀

PR #225, opened 2025-02-07
Update output to comply with new 4C standard

davidrudlstorfer:
- Remove Section "introduction", i.e., DSURF 1, DLINE 2, ...
- Remove option to output design_description


davidrudlstorfer:
@isteinbrecher everything works again - it took me some time to figure out where the read-in mechanism failed 😅

While at it I also removed the outdated design-description option and cleaned all reference files.

The unfortunate thin is that cubitpy now fails because as far as I can see this needs to be updated as well.

isteinbrecher (approved):
Thanks for the effort @davidrudlstorfer . I added the changes to CubitPy, the pipeline should pass now - if not, I will have a look next week.

Issue #106, opened 2024-10-02
Make sure that all files in `tests/reference-files` are actually used


isteinbrecher:
As of right now, we don't have a way of checking that all files in `tests/reference-files` are actually used during testing. This sometimes results in old, not used files being there without someone noticing. Let's discuss if there is an easy way to check that all reference files are actually used during testing.

davidrudlstorfer:
I gave this problem some thoughts today and also tested a small strategy - unfortunately I was not succesfull and I think with our current test approach there's no easy way to check that all reference files are used.

1. The optimal way would be to check if all reference files are used by running a pre-commit hook. In our case this is not easily doable because the reference file paths are generated at runtime (with the pytest fixture `get_corresponding_reference_file_path`)

2. Therefore, we definitely must check the usage of all files during the execution of pytest. Again, the optimal solution would be on a file system basis to check if all files were being read - but that also turns out to be not really doable. To solve this problem I simply tracked the files during the pytest run (simply store the path of the used reference file in `get_corresponding_reference_file_path` into a global variable) and compared it with all reference files at the end of the run (a `pytest_sessionfinish` function is executed at the end of each pytest run). And this approach now shows us why it is not easily feasible =>

    2.1 In our testing strategy we also rely on hard coded paths to reference files which are, therefore, not tracked with the function `get_corresponding_reference_file_path` => these files won't be checked

    2.2 A few reference files are only used in specific testing setups - i.e., CubitPy or 4C. This means we would have to track the used reference files across all Github Actions, somehow combine them into a global list in a separate Github workflow and then we are able to check if all files are used.

This means we would have to 1. rewrite all file loads to be handled through one central function, i.e., `get_corresponding_reference_file_path` 2. temporarily write all utilized reference files into a file and upload it as an artifact and 3. create a new Github action which combines all used reference files and checks that all files are used

Summing up, in my opinion it is currently not easily feasible to check if all reference files are used by the testing setup. I propose to just check all of them manually now and remove any old outdated ones. In the future we just have to keep track of changes within the PR's.

@isteinbrecher maybe you have a different opinion or I missed a simpler approach - curious about your opinion

isteinbrecher:
Thanks for the explanations @davidrudlstorfer. I actually had a look at this a while back (unfortunately never pushed the [branch](https://github.com/isteinbrecher/meshpy/tree/check-reference-files) or commented here...). Actually, I think I ended up at similar points as you did.

I think we have two options:
[...]

PR #227, opened 2025-02-11
Add testing for CubitPy created input files

isteinbrecher:
We use a few input files created with CubitPy in MeshPy testing. For some (unfortunately not all of them) we have the scripts checked into the repository that create those files. This PR adds tests to check that the scripts are up to date. These tests require the `--CubitPy` flag.


tests/reference-files/test_four_c_solid_shell_direction_detection_blocks.dat
```diff
@@ -6,6 +6,14 @@
...
 //     https://www.unibw.de/imcs-en
 // -----------------------------------------------------------------------------
+// -----------------------------------------------------------------------------
+// This input file was created with MeshPy.
+// Copyright (c) 2018-2025
+//     Ivo Steinbrecher
+//     Institute for Mathematics and Computer-Based Simulation
+//     Universitaet der Bundeswehr Muenchen
+//     https://www.unibw.de/imcs-en
+// -----------------------------------------------------------------------------
```
davidrudlstorfer, line 16: This file now contains the header twice

tests/test_create_cubit_input.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.cubitpy
+def test_create_cubit_input_tube(
+    tmp_path,
+    current_test_name,
+    get_corresponding_reference_file_path,
+    assert_results_equal,
+):
+    """Test that the solid tube reference file is up to date."""
+
+    file_path = tmp_path / (current_test_name + ".dat")
```
davidrudlstorfer, line 49: can we use the reference file name function here? Or am I missing something here

davidrudlstorfer: Actually throughout the entire file

tests/reference-files/test_four_c_solid_shell_direction_detection_blocks.dat
isteinbrecher: Yes, that is because this dat file is created based on the file `test_create_cubit_input_solid_shell_blocks.dat` - which now contains the MeshPy header since it is created with MeshPy. So when `test_create_cubit_input_solid_shell_blocks.dat` is read, the header is viewed as a comment and printed twice. This is a very special case and I am fine with simply keeping it this way (at least for now).

tests/test_create_cubit_input.py
isteinbrecher: The thing is that `get_corresponding_reference_file_path` returns the full path and not the filename, so the alternative would be:
```python
file_path = tmp_path / get_corresponding_reference_file_path().name
```
I am fine either way.

davidrudlstorfer: I think that's a better approach - because with #106 it is crucial to pipe all reference file stuff through one single function

isteinbrecher: I agree, should be fixed.

davidrudlstorfer (approved):

PR #232, opened 2025-02-14
Improve path finding in testing

isteinbrecher:
Replace all hard coded paths to files in the testing reference file directory with `get_corresponding_reference_file_path`.


tests/create_cubit_input.py
```diff
@@ -302,6 +302,11 @@ def rotate_list(original_list, n):
 if __name__ == "__main__":
     # Execution part of script.
 
+    # Todo:
+    # The paths here are wrong! We have to think anyway if we want to keep this
+    # part of the script or just rely on the testing (which should be more than
+    # sufficient).
+
```
davidrudlstorfer, line 309: Could you open an issue for that so it is documented?

davidrudlstorfer (approved):
Thanks for tackling this problem this fast! Looks very good - I wasn't even aware that the base name feature was already present in our testing approach 😅

PR #236, opened 2025-02-14
Update license to reflect a general MIT license

### LINKED ISSUE #234: Adjust the license so it is recognized by Github
Currently the license reflects a MIT license but the formatting and initial header is not standardized. That's why Github won't recognize it it its tooling.

Therefore, we should adapt the license to be a standard MIT license. 

We should also think about removing the specific authors and institutions from the license as this is already present multiple times across the repository (readme, pyproject, ...)

For example this would be a pretty straight forward MIT license which I found with a quick Google search https://github.com/twbs/bootstrap/blob/main/LICENSE

davidrudlstorfer:
Closes #234

### DIFF
--- a/.binder/Dockerfile
+++ b/.binder/Dockerfile
@@ -1,12 +1,6 @@
-# MeshPy: A beam finite element input generator
+# The MIT License (MIT)
 #
-# MIT License
-#
-# Copyright (c) 2018-2025
-#     Ivo Steinbrecher
-#     Institute for Mathematics and Computer-Based Simulation
-#     Universitaet der Bundeswehr Muenchen
-#     https://www.unibw.de/imcs-en
+# Copyright (c) 2018-2025 MeshPy Authors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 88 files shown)


.binder/Dockerfile
```diff
@@ -1,12 +1,6 @@
-# MeshPy: A beam finite element input generator
```
isteinbrecher, line 1: Should we change this to `MeshPy: A general purpose 3D beam finite element input generator`?

isteinbrecher: Would this work
```
# The MIT License (MIT)
#
# MeshPy: A general purpose 3D beam finite element input generator
# Copyright (c) 2018-2025 MeshPy Authors
#
# Permission is hereby granted, free of charge, to any person obtaining a copy
# of this software and associated documentation files (the "Software"), to deal
# in the Software without restriction, including without limitation the rights
# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
# copies of the Software, and to permit persons to whom the Software is
# furnished to do so, subject to the following conditions:
...
```
Can we test this somehow?

davidrudlstorfer: Sounds good - I'll test it out on my fork and then I'll report back here if it works

davidrudlstorfer:
I am creating a new PR based on the above described improvements - starting from scratch due to too much work regarding the rebase

isteinbrecher:
Works for me. Is there any reason you did not force push to this branch?

Also one thing that I just thought about, if it works with
```
# MeshPy: A general purpose 3D beam finite element input generator
# Copyright (c) 2018-2025 MeshPy Authors
```
should we maybe also add the website or GitHub link there?

PR #240, opened 2025-02-17
Add marker to skip tests in CICD

isteinbrecher:
This PR does two things:
1. It temporarily deactivates `test_four_c_simulation_dirichlet_boundary_to_neumann_boundary_with_all_values` due to a bug in 4C
2. It introduces a new marker in our `pytest` framework to mark temporarily excluded tests.

### DIFF
--- a/tests/test_four_c_simulation.py
+++ b/tests/test_four_c_simulation.py
@@ -774,6 +774,10 @@ def test_four_c_simulation_dbc_monitor_to_input(
     )
 
 
+# TODO remove
+@pytest.mark.skip(
+    reason="Upstream bug in 4C https://github.com/4C-multiphysics/4C/issues/385"
+)
 @pytest.mark.parametrize(*PYTEST_4C_SIMULATION_PARAMETRIZE)
 def test_four_c_simulation_dirichlet_boundary_to_neumann_boundary_with_all_values(
     enforce_four_c,


isteinbrecher:
@davidrudlstorfer what do you think about the additional marker? Does this do what it is supposed to? Where should we document it? I would only to that somewhere "internal" since this is not a marker that is supposed to be used for "normal" testing.

davidrudlstorfer:
that's a really good idea

but I am not yet sold on the name of the new flag - it is not only deactivated in cicd but everywhere

could we do something simpler and more expressive like

`exclude` or `exclude_testing` or `exclude_from_testing` or `tmp_exclude`

isteinbrecher:
I like `tmp_exclude` to suggest that this is not something we want to use in the standard workflow.

davidrudlstorfer:
I'll add the flag - regarding documentation I would not write anything about it or simply a one liner in the Contributing section?

isteinbrecher:
Thanks, then lets go with no documentation.

davidrudlstorfer:
Let's use this approach

https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-test-functions

I'll update the PR

davidrudlstorfer:
Seems to work - now we also see that one test is skipped in the test overview.

Feel free to merge - I am not able to approve

Issue #243, opened 2025-02-17
Follow-ups to improve the new src layout and pyproject.toml setup


davidrudlstorfer:
Following the initial move to the src layout #238 a few minor points should be addressed:

- [x] improve naming /split of `utils/utils.py` to better reflect the contents
- [x] fix the circular import in `utils/utils.py`
- [ ] update the `pyproject.toml` file #280
  - [ ] Identify unused packages #280
  - [ ] Choose sensible packages, e.g., default, `testing`, `arborx`, `cubitpy` ... #280 
- [x] ~~improve setup of pybind - I think we can move everything from the `setup.py` file into the `pyproject.toml`~~ we continue using this setup see #258 
- [x] add pre-commit hook to no longer allow relative imports in `src/meshpy` (ruff supports that and we only need to activate it)
- [x] **Add innit file to meshpy**

isteinbrecher:
Just to be complete, pybind is not related to the stuff in `setup.py` that is only for Cython code.

What do you mean with:
> Add innit file to meshpy

davidrudlstorfer:
> What do you mean with:
> 
>  Add innit file to meshpy

I forgot to add the `__init__.py` file at the top-level folder of MeshPy, i.e., `src/meshpy/`

This way the imports won't work if I import MeshPy somewhere else. I've already added it and the PR will follow soon

isteinbrecher:
I added two points to the description.

PR #245, opened 2025-02-17
Cleanup utils

davidrudlstorfer:
Part of #243


davidrudlstorfer:
I've also deleted because it is unused

```
def is_testing_github():
    """Check if the current environment is a testing action on GitHub."""
    return "GITHUB_ACTION" in os.environ.keys()
```

isteinbrecher (approved):

PR #251, opened 2025-02-18
Move the input file to 4c

isteinbrecher:
Move the `InputFile` class to the `four_c` namespace and out of `core`.

Also, the file is now named `input_file.py` instead of `inputfile.py`.


src/meshpy/core/nurbs_patch.py
```diff
@@ -78,7 +78,7 @@ def get_nurbs_dimension(self):
     def add_element_specific_section(self, sections):
         """Return additional information of the NURBS patch."""
 
-        from .inputfile import InputSectionMultiKey
+        from meshpy.four_c.input_file import InputSectionMultiKey
```
davidrudlstorfer, line 81: Can we move this import to the top of the file directly?

davidrudlstorfer (approved):
Nice - thanks for tackling this - and also thanks for the renaming - this bothered me already yesterday :D

I only have a minor comment regarding one import

src/meshpy/core/nurbs_patch.py
isteinbrecher: I added a todo, to remove the circular import here.

davidrudlstorfer (approved):

PR #258, opened 2025-02-24
Editable and non editable installation with setup tools

isteinbrecher:
Test if editable and non editable installations work with setuptools.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,6 @@ license = {file = "LICENSE"}
 dependencies = [
   "autograd",
   "black==22.12.0",
-  "Cython",
   "geomdl",
   "ipykernel",
   "notebook",
@@ -31,10 +30,6 @@ dependencies = [
 ]
 
 dynamic = ["version"]
-
-[tool.setuptools.packages.find]
-where = ["src"]
-include = ["meshpy"]
 
 [project.optional-dependencies]
 CI-CD = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ CI-CD = [
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
-addopts = "-p pytest_cov --cov-report=term --cov-report=html --cov-fail-under=90 --cov=src/meshpy/ --cov-append"
+addopts = "-p pytest_cov --cov-report=term --cov-report=html --cov-fail-under=90 --cov=src/meshpy/"
 markers = [
   "fourc: tests in combination with 4C",
   "arborx: tests in combination with ArborX",
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,8 +32,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Run the test suite
         uses: ./.github/actions/run_tests
-        with:
-          install-command: "-e .[CI-CD]"
       - name: Upload test results on failure
         if: failure()
         uses: actions/upload-artifact@v4
@@ -58,7 +56,6 @@ jobs:
         uses: ./.github/actions/run_tests
         with:
           source-command: "source python-workflow-venv/bin/activate"
-          install-command: "-e .[CI-CD]"
           additional-pytest-flags: "--CubitPy"
       - name: Upload test results on failure
         if: failure()
@@ -117,7 +114,6 @@ jobs:
         uses: ./.github/actions/run_tests
         with:
           source-command: "source python-workflow-venv/bin/activate"
-          install-command: "-e .[CI-CD]"
           additional-pytest-flags: "--performance-tests --exclude-standard-tests -s --no-cov"
       - name: Upload test results on failure
         if: failure()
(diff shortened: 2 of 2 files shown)


isteinbrecher:
@davidrudlstorfer Looks like everything works, the tests still fail, due to the coverage goal not being reached in non-editable install variants. Could you add the fix you also proposed in #249 here? It should be sufficient to only get the coverage for the 4C+ArborX test case.

I switched all test cases to use a non-editable installation, except the 4C+ArborX test case (there we require an editable installation because ArborX is used). Do you see any issues with that?

pyproject.toml
```diff
@@ -32,10 +31,6 @@ dependencies = [
 
 dynamic = ["version"]
 
-[tool.setuptools.packages.find]
-where = ["src"]
-include = ["meshpy"]
```
davidrudlstorfer, line 37: I am still trying to get a good overview about your raised points in #249 - that might take a while 😅

But one question which directly arises on my side - if we remove this - won't we also treat everything thats inside the src folder as a module? For example, `pybind`, the `build` folder, ...

Is this a behaviour we intend?

davidrudlstorfer: Especially because we do not want to publish pybind11 if we publish everything on pypi

isteinbrecher: Good point! I think we want automatic package discovery with `setuptools` (https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery). The build folder might not be an issue, since it might not be there when installing MeshPy. Otherwise I would be fine to move them to the top level repo again. Also, are these folders detected when they don't contain an `__init__.py` file?

davidrudlstorfer: You're totally right - I've tested everything locally and PyBind11 is not packaged as you suggested because it does not have an __init__.py file

davidrudlstorfer:
@isteinbrecher was this the change you intended to fix the pipeline? This just deactivates the coverage for the test matrix runs (non editable installs)

isteinbrecher:
@davidrudlstorfer yes, that was exactly what I was referring to. I was just not sure, because in #249 you also changed the line
```diff
- addopts = "-p pytest_cov --cov-report=term --cov-report=html --cov-fail-under=90 --cov=src/meshpy/"
+ addopts = "-p pytest_cov --cov-report=term --cov-report=html --cov-fail-under=90 --cov=src/meshpy/ --cov-append"
```
and I have to admit, that these sections in the `pyproject.toml` file are still black magic to me 😅.

I think we can then proceed to merge this PR - only one thing that I thought about, would it make more sense to turn off the coverage by default and only activate it in the relevant jobs? If you agree then I can change that and push to this PR.

davidrudlstorfer:
Good catch - I removed the `--cov-append` so the coverage is "recalculated" in each run.

Until now the "old" coverage was used and appended - this leads to problems if a file get's deleted but the coverage actually never updated - except when you deleted the coverage file

So I would still remove this flag

Regarding the turn-off of the coverage by default - I would not do that. I really like that with a regular run of pytest on a local machine you get the coverage report without fiddling around

davidrudlstorfer:
And regarding the "black magic" its actually really simple.

`-p pytest_cov` adds the coverage plugin to the pytest run
`--cov-report=term` shows the coverage in the terminal
`--cov-report=html` also saves the coverage report to html (we need that for the coverage report we upload)
`--cov-fail-under=90` specifies that we want at least 90% coverage
`--cov=src/meshpy/`specifies the directory from which we want the coverage report

While listing these things actually one thing came up - I think we should also test the coverage for a "standard" install in the pipeline. If we locally run pytest we also enfore 90% coverage

But we only test the coverage for the 4C/ArborX install in the pipeline. So if the coverage drops we will only have problems during local runs because we are really close to 90% in a standard install

isteinbrecher:
Removal of `--cov-append` makes a lot of sense - this also explains some weird behaviour I noticed a while ago with the coverage report.

I agree, it makes sense to keep the coverage for local `pytest` runs. I think we can set the default coverage goal (in `pyproject.toml`) to a stable value for the standard tests (if 90 turns out to be too harsh, lets try 89 or 88) and we can add the option `--cov-fail-under=93` to the 4C-ArborX tests. I tried that locally, this overrules the "default" options. I also pushed a commit, let's see if it works.

The only thing I am not sure, how to extend the matrix tests to include an editable + coverage version. Do you have an idea how to do that? Feel free to simply push to this branch.

davidrudlstorfer:
Sounds good! I'll look into it if we can elegantly extend the matrix testing

davidrudlstorfer:
This works as expected - I think that's a pretty nice solution. Now we test the standard install in editable and non editable mode so the coverage of 90% is also tested in the pipeline.

isteinbrecher:
Very nice! Just a thought - would it make more sense to switch this, i.e., test the non-editable installation with the more recent python version? Because the non-editable installation turned out to cause more issues in the past, and then we would check this one with the newer version of python.

davidrudlstorfer:
I messed somewhere up - now no coverage is anywhere

davidrudlstorfer:
Maybe you could briefly check what's going on here - I do not understand why no coverage is present with Python 3.12 now 😅

isteinbrecher:
Really weird. I pushed a test commit, lets see what happens

davidrudlstorfer:
That's really weird 😅 thanks for debugging it and looking into it

Feel free to merge 🚀

PR #259, opened 2025-02-24
Allow ArborX search for 1D arrays

isteinbrecher:
Until now the ArborX search only worked for points in $\mathbb{R}^3$ and $\mathbb{R}^6$. This PR allows for points from $\mathbb{R}^1$ up to $\mathbb{R}^8$, which should also cover some of the more crazy space-time formulations.


davidrudlstorfer (approved):
Thanks for tackling this directly and providing a solution!

PR #262, opened 2025-02-27
Update MIT license

### LINKED ISSUE #234: Adjust the license so it is recognized by Github

davidrudlstorfer:
Successor of #236 

Closes #234 

@isteinbrecher now this works as expected.

> Works for me. Is there any reason you did not force push to this branch?

Regarding this - I did not want to rebase everything and I took some sort of manual approach to update all licenses because the pre-commit hook always copies in a second license above the first one.

> should we maybe also add the website or GitHub link there?

I would not add the website or the github repo - the license should be tied to the code and not the hosting website or repo

Regarding the additions of the MeshPy description - I tried the following things: 

```
The MIT License (MIT)

Copyright (c) 2018-2025 MeshPy Authors

Permission is hereby granted, free of charge, to any person obtaining a copy
....
```

was successfully recognized by Github on my fork

```
The MIT License (MIT)

MeshPy: A general purpose 3D beam finite element input generator
Copyright (c) 2018-2025 MeshPy Authors

Permission is hereby granted, free of charge, to any person obtaining a copy
....
```

was not recognized at all

```
The MIT License (MIT)

Copyright (c) 2018-2025 MeshPy Authors
MeshPy: A general purpose 3D beam finite element input generator


Permission is hereby granted, free of charge, to any person obtaining a copy
....
```

was also not recognized at all

I would, therefore, stay with the here propposed approach so it is recognized by Github. We could differentiate between a file header and our license - like we do it in 4C. But I think for this repo we could save us the hassle to maintain differing license formats.

If ok for you we could merge

### DIFF
--- a/.binder/Dockerfile
+++ b/.binder/Dockerfile
@@ -1,12 +1,6 @@
-# MeshPy: A beam finite element input generator
+# The MIT License (MIT)
 #
-# MIT License
-#
-# Copyright (c) 2018-2025
-#     Ivo Steinbrecher
-#     Institute for Mathematics and Computer-Based Simulation
-#     Universitaet der Bundeswehr Muenchen
-#     https://www.unibw.de/imcs-en
+# Copyright (c) 2018-2025 MeshPy Authors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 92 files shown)


isteinbrecher (approved):
Looks good to me. We can also think about using just a short License header similar to 4C in the future. I would think about that if MeshPy is renamed at some point.

Regarding the rebase, you can force push any branch to the one used to open a merge request, does not have to be a "rebased" branch.

Edit: I didn't fully read your comment 😅 and agree with the remarks regarding multiple license files.

davidrudlstorfer:
@isteinbrecher thanks for the quick review and the info regarding the force push 🙏🏻

PR #269, opened 2025-02-28
Remove some old file headers

isteinbrecher:


davidrudlstorfer (approved):
Thanks @isteinbrecher 

That also bugged me for a long time 😅

isteinbrecher:
I actually thought we removed them at some point, but seems like some went under the radar 😂

PR #264, opened 2025-02-27
Improve geometric search doc

davidrudlstorfer:
Thanks for the explanation @isteinbrecher in #249!

Also part of #249

PR #270, opened 2025-02-28
Fix warning on escaped string in doctoring of geometric search

isteinbrecher:
With #264 we nicely added documentation for the geometric search module. This introduced a warning
```
=============================== warnings summary ===============================
src/meshpy/geometric_search/__init__.py:22
  /home/runner/work/meshpy/meshpy/src/meshpy/geometric_search/__init__.py:22: DeprecationWarning: invalid escape sequence '\m'
    """This module defines geometric search functionality.
```

This can simply be fixed by using `r"""` instead of `"""`. The other changes in this PR are just added links in the doc.

### DIFF
--- a/src/meshpy/geometric_search/__init__.py
+++ b/src/meshpy/geometric_search/__init__.py
@@ -19,7 +19,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):
Nice fix - I also briefly saw the warning but unfortunately I did not investigate further.

One side note - should we restore the line length here? Unfortunately, this does not work automatically because somehow the formatter does not work once you add the list signs (`-`).

I simply did the line breaks manually

isteinbrecher:
I tried to restore the line length as best as possible, but one of the links itself has over 90 characters 😅. I pushed a small fix, I guess for now this is a sufficient solution.

davidrudlstorfer (approved):

PR #271, opened 2025-02-28
Improve coupling checks

isteinbrecher:
With the previous coupling checks, we had a bug in the evaluation of the distance. This fixes this and adds a test case.

### DIFF
--- a/src/meshpy/core/coupling.py
+++ b/src/meshpy/core/coupling.py
@@ -96,11 +96,11 @@ class Coupling(BoundaryConditionBase):
             return
 
         nodes = self.geometry_set.get_points()
-        pos = np.zeros([len(nodes), 3])
+        diff = np.zeros([len(nodes), 3])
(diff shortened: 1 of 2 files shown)


src/meshpy/core/coupling.py
```diff
@@ -96,11 +96,11 @@ def check(self):
...
 
         nodes = self.geometry_set.get_points()
-        pos = np.zeros([len(nodes), 3])
+        diff = np.zeros([len(nodes), 3])
         for i, node in enumerate(nodes):
             # Get the difference to the first node
-            pos[i, :] = node.coordinates - nodes[0].coordinates
-        if np.linalg.norm(pos) > mpy.eps_pos:
+            diff[i, :] = node.coordinates - nodes[0].coordinates
+        if np.max(np.linalg.norm(diff, axis=1)) > mpy.eps_pos:
```
davidrudlstorfer, line 103: That's a nice check 👌

PR #272, opened 2025-03-03
Update `.gitignore`

isteinbrecher:
Add the default api documentation folder to the `.gitignore. file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,9 @@ _site
 *.lock
 website/public/
 
+# Documentation
+api-documentation/
+
 # Don't ignore directory of Github workflows
 !.github
 !.pre-commit-config.yaml


davidrudlstorfer (approved):
This makes sense 🙏🏻

I think it would also be beneficial if we someday update our `.gitignore` file alltogether.

For all my projects I use gitignore.io - there you simply add your keywords and all important files are added (for example - `Python` `macOS` `Windows` `Linux` etc.

You can see an example in my repo here https://github.com/davidrudlstorfer/pyskel/blob/main/.gitignore

isteinbrecher:
Thanks for the hint, https://gitignore.io seems like a good alternative. I will open an issue on this.

PR #275, opened 2025-03-04
Create beam contact header

knarfnitram:
Add the beam contact header and a small overall test case to meshpy.

Currently blocked by #161 .


isteinbrecher:
@knarfnitram can you rebase this on the current `main` branch? Somehow the diff in GitHub does not work, and there are merge conflicts.

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,245 @@ def set_header_static(
...
+
+
+def set_binning_strategy(
+    input_file,
+    *,
+    binning_bounding_box=None,
+    binning_cutoff_radius=None,
+    option_overwrite=None,
+    **kwargs,
+):
```
isteinbrecher, line 526: Can we add type hints here?

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,245 @@ def set_header_static(
...
+    **kwargs:
+        Optional parameters can also be extracted from the keyword arguments
+    """
+
+    # do nothing if not both arguments are specified
+    if [binning_cutoff_radius, binning_bounding_box].count(None) == 2 and [
+        kwargs.get("binning_cutoff_radius", None),
+        kwargs.get("binning_bounding_box", None),
+    ].count(None) == 2:
+        return
```
isteinbrecher, line 549: I still don't get this construct. What is checked here? Also, I think that "binning_cutoff_radius" will not be a key of `kwargs` since it is its own keyword argument. (Also applies to code below)

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,245 @@ def set_header_static(
...
+
+    # extract values from kwargs or provide default one
+    if binning_cutoff_radius is None:
+        binning_cutoff_radius = kwargs.get("binning_cutoff_radius", 1)
```
isteinbrecher, line 553: Why not simply:
```suggestion
        binning_cutoff_radius = 1
```

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,245 @@ def set_header_static(
...
+        option_overwrite = kwargs.get("option_overwrite", False)
+
+    binning_bounding_box_string = " ".join([str(val) for val in binning_bounding_box])
+
+    input_file.add(
+        InputSection(
+            "BINNING STRATEGY",
+            f"""
+        BIN_SIZE_LOWER_BOUND    {binning_cutoff_radius}
+        DOMAINBOUNDINGBOX                     {binning_bounding_box_string}
```
isteinbrecher, line 568:
```suggestion
        BIN_SIZE_LOWER_BOUND {binning_cutoff_radius}
        DOMAINBOUNDINGBOX {binning_bounding_box_string}
```

src/meshpy/four_c/header_functions.py
```diff
@@ -249,26 +249,12 @@ def set_beam_to_solid_meshtying(
...
-            (
-                f"Binning bounding box ({binning_bounding_box}) and binning cutoff radius "
-                f"({binning_cutoff_radius}) both have to be set or none of them."
-            )
-        )
+    set_binning_strategy(
+        input_file,
+        binning_cutoff_radius=binning_cutoff_radius,
+        binning_bounding_box=binning_bounding_box,
+        option_overwrite=True,
```
isteinbrecher, line 256:
```suggestion
        option_overwrite=False,
```

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,245 @@ def set_header_static(
...
+        """,
+            option_overwrite=option_overwrite,
+        )
+    )
+
+
+def set_beam_contact_section(
+    input_file,
+    *,
+    interaction_strategy="penalty",
```
isteinbrecher, line 664:
```suggestion
    input_file: InputFile,
    *,
    interaction_strategy: str="penalty",
```

tests/test_four_c.py
```diff
@@ -595,3 +601,108 @@ def test_four_c_beam_to_solid(
...
+    )
+
+    # Set the parameters for beam to beam contact.
+    set_beam_contact_section(
+        mesh,
+        btb_penalty=50,
+        penalty_regularization_g0=r_beam * 0.02,
+        binning_cutoff_radius=5,
+        binning_bounding_box=[-3, -3, -3, 3, 3, 3],
+    )
```
knarfnitram, line 699: @isteinbrecher  1/3
For some problems( meshtying, beam to beam interaction), we have to additional add the binning strategy to the input file. 
For the function `set_beam_to_solid_meshtying` this is done by adding the parameters `binning_bounding_box` and `binning_cutoff_radius` to the function as input parameters. Since the parameters are described within the function  `set_beam_to_solid_meshtying` and `set_binning_strategy`, this leads to duplication of the same description of parameters. I want to avoid that but still set them easily from here

My solution for not duplicating this is variables is to add them to the`kwargs` object, which is done here.
The variable `binning_bounding_box`  can be set for `set_beam_contact_section` and it is forwarded to  `set_binning_strategy` with the keyword arguments.

src/meshpy/four_c/header_functions.py
knarfnitram, line 664: @isteinbrecher 2/3 
Within the function declaration, the variables for `set_binning_strategy` are stored within the `kwargs` object therefore parsed to `set_binning_strategy`.

src/meshpy/four_c/header_functions.py
knarfnitram: @isteinbrecher 3/3 Since, the variables may also be within `kwargs`, we need to extract the values for the variables.

src/meshpy/four_c/header_functions.py
knarfnitram: I would actually like these spaces here, because then values within the ´.dat` file are aligned perfectly.

src/meshpy/four_c/header_functions.py
isteinbrecher: I don't think this is how it works. If `"binning_cutoff_radius"`, it will always be stored in the variable `binning_cutoff_radius` and not in the `kwargs` dictionary.

But that does not matter if you chose the approach in my comment below (or above, I am still confused by GitHub sometimes).

isteinbrecher (commented):
Regarding the `kwargs` one thing that is commonly used for the type of problem you face is the following:
```python
def set_beam_to_solid_meshtying(
    input_file,
    interaction_type,
    *,
    contact_discretization=None,
    segmentation=True,
    segmentation_search_points=2,
    couple_restart=False,
    mortar_shape=None,
    n_gauss_points=6,
    n_integration_points_circ=None,
    penalty_parameter=None,
    coupling_type=None,
    binning_bounding_box=None,
    binning_cutoff_radius=None,
    option_overwrite=False,
    binning_strategy_kwargs: dict={}
):
    """Doc"""

    ...

    set_binning_strategy(
        input_file,
        option_overwrite=True,
        **binning_strategy_kwargs
    )

    ...


def set_beam_contact_section(
    input_file,
    *,
    interaction_strategy="penalty",
    btb_penalty: float = 0,
    btb_line_penalty: float = 0,
    per_shift_angle: list[float] = [70, 80],
    par_shift_angle: list[float] = [70, 80],
    b_seg_angle: float = 12,
    num_integration: int = 5,
    option_overwrite: bool = False,
    penalty_law="LinPosQuadPen",
    penalty_regularization_g0: float = 0,
    penalty_regularization_f0: float = 0,
    penalty_regularization_c0: float = 0,
    binning_strategy_kwargs: dict={},
    **kwargs,
):
    """Doc"""

   ...

    # beam contact needs a binning strategy
    set_binning_strategy(input_file, **binning_strategy_kwargs)

    ...


def set_binning_strategy(
    input_file,
    *,
    binning_bounding_box=[-10, -10, -10, 10, 10, 10],
    binning_cutoff_radius=1,
    option_overwrite=None
):
    
    # No need to fiddle with kwargs here
    ...
```
When you call a function that will itself call a (or multiple) function(s) and you don't want to pass the general `kwargs` to all functions, you can define dictionaries `binning_strategy_kwargs={}` that will be passed to `set_binning_strategy` as the keyword arguments. This also has the nice feature, that you can simply define the default values for the `set_binning_strategy` keyword arguments in the signature.

src/meshpy/four_c/header_functions.py
knarfnitram: Because the value can still be in the dict and we need to extract it (get method), if this fails we provide a default value `1`.

src/meshpy/four_c/header_functions.py
knarfnitram: I tried that too, however, by changing it here to `False`, you need than to refactor the test `test_header_functions_static`.

src/meshpy/four_c/header_functions.py
knarfnitram: So the problem is here that the previous assumption is that only if both parameters are specified, something should be specified. 
Since we have now two places where the information can be stored( either by direct command or by the dict `binning_dict`, we need to check both places.

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+    input_file: InputFile,
+    *,
+    interaction_strategy: str = "penalty",
+    btb_penalty: float = 0,
+    btb_line_penalty: float = 0,
+    per_shift_angle: list[float] = [70, 80],
+    par_shift_angle: list[float] = [70, 80],
+    b_seg_angle: float = 12,
+    num_integration: int = 5,
+    option_overwrite: bool = False,
```
isteinbrecher, line 645: Can we move this at the end to indicate that this is not related to the current function?

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+    penalty_law: str = "LinPosQuadPen",
+    penalty_regularization_g0: float = 0,
+    penalty_regularization_f0: float = 0,
+    penalty_regularization_c0: float = 0,
+    binning_parameters: dict = {},
+    beam_contact_parameters: dict = {},
+):
+    """Set default beam contact section, for more and updated details see
+    respective input file within 4C. Parameters for set_binning_strategy and
+    set_beam_interaction may be forwarded as keyword arguments.
```
isteinbrecher, line 655: I still don't get why the doctoring formatter doesn't change this... But that is not in the scope id this PR.

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+    par_shift_angle: list[float] = [70, 80],
+    b_seg_angle: float = 12,
+    num_integration: int = 5,
+    option_overwrite: bool = False,
+    penalty_law: str = "LinPosQuadPen",
+    penalty_regularization_g0: float = 0,
+    penalty_regularization_f0: float = 0,
+    penalty_regularization_c0: float = 0,
+    binning_parameters: dict = {},
+    beam_contact_parameters: dict = {},
```
isteinbrecher, line 651: Maybe rename this to `beam_interaction_parameters`?

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+        If existing options should be overwritten. If this is false and an
+        option is already defined, and error will be thrown.
+    """
+
+    # extract values from beam_contact_dict or provide default one
+    if every_iteration is None:
+        every_iteration = beam_contact_dict.get("every_iteration", False)
+
+    if option_overwrite is None:
+        option_overwrite = beam_contact_dict.get("option_overwrite", False)
```
isteinbrecher, line 618: See previous comment

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+        BEAMS_PENREGPARAM_C0            {penalty_regularization_c0}
+        BEAMS_MAXDISISCALEFAC           -1.0
+        BEAMS_MAXDELTADISSCALEFAC       -1.0
+        """,
+            option_overwrite=option_overwrite,
+        )
+    )
+
+    # beam contact needs a binning strategy
+    set_binning_strategy_section(input_file, **binning_parameters)
```
isteinbrecher, line 734:
```suggestion
    set_binning_strategy_section(input_file, option_overwrite= option_overwrite, **binning_parameters)
```

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +518,220 @@ def set_header_static(
...
+
+    # beam contact needs interaction strategy
+    set_beam_interaction_section(input_file, **beam_contact_parameters)
```
isteinbrecher, line 737:
```suggestion
    set_beam_interaction_section(input_file, option_overwrite= option_overwrite, **beam_contact_parameters)
```

src/meshpy/four_c/header_functions.py
```diff
@@ -249,25 +251,13 @@ def set_beam_to_solid_meshtying(
...
-                f"({binning_cutoff_radius}) both have to be set or none of them."
-            )
+    # Only update section if both parameters are provided
+    if binning_cutoff_radius and binning_bounding_box:
+        set_binning_strategy_section(
+            input_file,
+            binning_cutoff_radius=binning_cutoff_radius,
+            binning_bounding_box=binning_bounding_box,
+            option_overwrite=option_overwrite,
         )
```
isteinbrecher, line 261: Could you actually use the same approach here for binning and the general parameters as in `set_beam_contact_section`. I really like the approach and this would improve this function as well. It will break backwards compatibility, but since I am the main user of this function, I am ok with that.

```python
    # beam interaction needs a binning strategy
    set_binning_strategy_section(input_file, option_overwrite=option_overwrite, **binning_parameters)

    # beam interaction needs interaction strategy
    set_beam_interaction_section(input_file, option_overwrite=option_overwrite, **beam_interaction_parameters)
```
Then we can also remove the duplicate documentation for these parameters for this.

tests/test_header_functions.py
```diff
@@ -173,3 +177,30 @@ def test_header_functions_stress_output(
...
     # Check the output.
     assert_results_equal(get_corresponding_reference_file_path(), input_file)
+
+
+def test_header_functions_beam_interaction(
+    get_corresponding_reference_file_path, assert_results_equal
+):
+    """Test the beam-to-beam contact header function with default parameter."""
+    # Set default values for global parameters.
+    mpy.set_default_values()
```
isteinbrecher, line 187: Not needed, this is done automatically.

tests/test_four_c.py
```diff
@@ -595,3 +601,110 @@ def test_four_c_beam_to_solid(
...
+def test_four_c_beam_to_beam_contact_example(
+    assert_results_equal, get_corresponding_reference_file_path
+):
+    """Small test example to show how a beam contact example with beam penalty
+    contact can be set up.
+
+    The test case consists of two beams: one beam is allocated along the x-axis and
+    the other beam is located along the y-axis.
+    The beam along the y-axis is placed above the other beam by an additional offset in z-Directions.
+    Due to prescribed displacements at the tips of beam in y-axis, the two beams get in contact around the origin.
```
isteinbrecher, line 615: Very nice!

Does this example actually work? If so, we can think of moving it to the four_c_simulation tests. If it works, we can to that in a follow up PR.

tests/test_four_c.py
```diff
@@ -595,3 +601,110 @@ def test_four_c_beam_to_solid(
...
+        [0, -l_beam / 2, h],
+        [0, l_beam / 2, h],
+        n_el=n_ele,
+    )
+
+    # Create a linear interpolation function with the displacement.
+    fun = Function(
+        "SYMBOLIC_FUNCTION_OF_SPACE_TIME -disp\n"
+        f"VARIABLE 0 NAME disp TYPE linearinterpolation NUMPOINTS 3 TIMES 0 1 1000.0 VALUES 0.0 {h+r_beam} {h+r_beam}"
+    )
```
isteinbrecher, line 662: You could `use create_linear_interpolation_function` here.

tests/test_four_c.py
```diff
@@ -595,3 +601,110 @@ def test_four_c_beam_to_solid(
...
+        btb_penalty=50,
+        penalty_regularization_g0=r_beam * 0.02,
+        binning_parameters={
+            "binning_cutoff_radius": 5,
+            "binning_bounding_box": [-3, -3, -3, 3, 3, 3],
+        },
+    )
+
+    # Add normal runtime output.
+    set_runtime_output(input_file=mesh)
```
isteinbrecher, line 704: Is there a reason you proved the mesh as a keyword argument here?

isteinbrecher (commented):
@knarfnitram thanks a lot for gong the extra mile and applying the previous discussions. I left some more comments - please don't get discouraged by the amount. I think this is a really nice addition to ease the creation of beam interaction examples in 4C!

@dharinib98 maybe this could also be relevant for you in the future.

tests/test_four_c.py
knarfnitram: I tried to choose it as reasonable as possible (that's why I contained initially more comments), I would actually prefer move/copying it to the examples as a `.ipynb`

tests/test_four_c.py
knarfnitram: not really, removed it.

tests/test_four_c.py
isteinbrecher: Sounds even better!

src/meshpy/four_c/header_functions.py
```diff
@@ -528,3 +514,227 @@ def set_header_static(
...
+
+    if [binning_bounding_box, binning_cutoff_radius].count(None) == 1:
+        raise ValueError(
+            f"The variables binning_bounding_box {binning_bounding_box} and binning_cutoff_radius {binning_cutoff_radius} must both be set."
+        )
+
+    if binning_bounding_box is not None:
+        binning_bounding_box_string = " ".join(
+            [str(val) for val in binning_bounding_box]
+        )
```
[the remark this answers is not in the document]
isteinbrecher: Can't you do:
```python
if binning_bounding_box is not None and binning_cutoff_radius is not None:
   ...
elif [binning_bounding_box, binning_cutoff_radius].count(None) == 2:
    return
else:
   # raise error
```
then the type check should be fine and you actually check if the input arguments are correct.

isteinbrecher (commented):
One final comment, then this is ready to be merged!

I thought about it, and I think it makes sense to move the test `test_four_c_beam_to_beam_contact_example` to the `four_c_simulation` tests - with some sort of 4C result test. Could you please have a look if this can be done straight forwardly, if so, let's do this in a follow up PR. It will be nice to have this in the examples at some point, but I am still not sure what the best way to do so is (and how to test this then). Also, if we test this now, we can guarantee that the simulation works until we maybe move it to the examples.

PR #276, opened 2025-03-05
Fix node locsys condition

knarfnitram:
### Previous state
The locsys condition in 4c were only written correct for the line geometry, namely:

`DESIGN LINE LOCSYS CONDITIONS`
`E 3 ROTANGLE 0.0 0.0 0.0 FUNCT 0 0 0 USEUPDATEDNODEPOS 0 USECONSISTENTNODENORMAL 0`

However, the option `USEUPDATEDNODEPOS` should be removed for a point and volume geometry resulting in the condition string:
`DESIGN POINT/VOL LOCSYS CONDITIONS`
`E 2 ROTANGLE 0.0 0.0 0.0 FUNCT 0 0 0 USEUPDATEDNODEPOS 0 `

### This MR 

- extracts a small function utility to ensure correct length of the user provided function array 
- checks the provided geometry before initialization of the LocSys condition to ensure that the correct string is added
- adds the possibility to provide a `function array `, `update_node_position` and `use_consistent_node_normal` to the LocSys condition
- adds all changes to the testcase.

### DIFF
--- a/src/meshpy/four_c/function_utility.py
+++ b/src/meshpy/four_c/function_utility.py
@@ -22,6 +22,8 @@
 """This module implements utility functions to create 4C space time
 function."""
 
+from typing import List
+
 import numpy as np
 
 from meshpy.four_c.function import Function
@@ -77,3 +79,26 @@ def create_linear_interpolation_function(
 
     variable_string = create_linear_interpolation_string(t, values, variable_name="var")
     return Function(f"{function_type} var\n" + variable_string)
+
+
+def ensure_length_of_function_array(
+    function_array: List[Function], length: int = 3
+) -> List[Function]:
+    """Performs size check of a function array and appends the function array to the given length, if a list with only one item is provided.
+    Args:
+        function_array: list with functions
+        length: expected length of function array
+
+    Returns:
+        function_array: list with functions with provided length
+    """
+
+    # extend items of function automatically if it is only provided once
+    if len(function_array) == 1:
+        return function_array * length
+
+    if len(function_array) != length:
+        raise ValueError(
+            f"The function array must have length {length} not {len(function_array)}."
+        )
+    return function_array
--- a/src/meshpy/four_c/function_utility.py
+++ b/src/meshpy/four_c/function_utility.py
@@ -81,10 +81,10 @@ def create_linear_interpolation_function(
     return Function(f"{function_type} var\n" + variable_string)
 
 
-def ensure_length_of_function_array(
-    function_array: List[Function], length: int = 3
-) -> List[Function]:
-    """Performs size check of a function array and appends the function array to the given length, if a list with only one item is provided.
+def ensure_length_of_function_array(function_array: List, length: int = 3):
+    """Performs size check of a function array and appends the function array
+    to the given length, if a list with only one item is provided.
+
     Args:
         function_array: list with functions
         length: expected length of function array
@@ -95,7 +95,7 @@ def ensure_length_of_function_array(
 
     # extend items of function automatically if it is only provided once
     if len(function_array) == 1:
-        return function_array * length
+        function_array = function_array * length
 
     if len(function_array) != length:
         raise ValueError(
--- a/src/meshpy/four_c/locsys_condition.py
+++ b/src/meshpy/four_c/locsys_condition.py
@@ -21,8 +21,14 @@
 # THE SOFTWARE.
 """This file contains the wrapper for the LocSys condition for 4c."""
 
+from typing import List, Optional, Union
+
 from meshpy.core.conf import mpy
+from meshpy.core.geometry_set import GeometrySet
+from meshpy.core.rotation import Rotation
 from meshpy.four_c.boundary_condition import BoundaryCondition
+from meshpy.four_c.function import Function
+from meshpy.four_c.function_utility import ensure_length_of_function_array
 
 
 class LocSysCondition(BoundaryCondition):
@@ -34,25 +40,54 @@ class LocSysCondition(BoundaryCondition):
 
     def __init__(
         self,
-        geometry_set,
-        rotation,
+        geometry_set: GeometrySet,
+        rotation: Rotation,
+        *,
+        function_array: Optional[List[Union[Function, int]]] = None,
+        update_node_position: bool = False,
+        use_consistent_node_normal: bool = False,
         **kwargs,
     ):
         """Initialize the object.
 
-        Args
-        ----
-        geometry_set: GeometrySet
-            Geometry that this boundary condition acts on.
-        rotation: Rotation
-            Object that represents the rotation of the coordinate system.
+        Args:
+            geometry_set: Geometry that this boundary condition acts on
+            rotation: Object that represents the rotation of the coordinate system
+            function_array: List containing functions
+            update_node_position: Flag to enable the updated node position
+            use_consistent_node_normal: Flag to use a consistent node normal
         """
+
+        # Validate provided function array.
+        if function_array is None:
+            function_array = [0, 0, 0]
+        else:
+            function_array = ensure_length_of_function_array(function_array, 3)
+
+        condition_string = (
+            "ROTANGLE {} {} {} ".format(*rotation.get_rotation_vector())
+            + "FUNCT {} {} {} "
+            + f"USEUPDATEDNODEPOS {int(update_node_position)}"
+        )
+
+        # Append the condition string with consistent normal type for line and surface geometry
+        if (
+            geometry_set.geometry_type is mpy.geo.line
+            or geometry_set.geometry_type is mpy.geo.surface
+        ):
+            condition_string = (
+                condition_string
+                + f" USECONSISTENTNODENORMAL {int(use_consistent_node_normal)}"
+            )
+        elif use_consistent_node_normal:
+            raise ValueError(
+                "The keyword use_consistent_node_normal only works for line and surface geometries."
+            )
 
         super().__init__(
             geometry_set,
-            "ROTANGLE {} {} {} FUNCT 0 0 0 USEUPDATEDNODEPOS 0 USECONSISTENTNODENORMAL 0".format(
-                *rotation.get_rotation_vector()
-            ),
+            bc_string=condition_string,
             bc_type=mpy.bc.locsys,
+            format_replacement=function_array,
             **kwargs,
         )
--- a/tests/test_four_c.py
+++ b/tests/test_four_c.py
@@ -237,8 +237,9 @@ def test_meshpy_locsys_condition(
 ):
     """Test case for point locsys condition for beams.
 
-    The testcase is similar to beam3r_herm2line3_static_locsys.dat, but
-    with simpler material.
+    The testcase is adapted from to beam3r_herm2line3_static_locsys.dat.
+    However it has a simpler material, and an additional line locsys
+    condition.
     """
 
     # Create the input file with function and material.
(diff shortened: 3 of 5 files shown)


src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +37,52 @@ class LocSysCondition(BoundaryCondition):
...
+        *,
+        function_array=None,
+        update_node_position: bool = False,
+        use_consistent_node_normal: bool = False,
         **kwargs,
     ):
         """Initialize the object.
 
         Args
         ----
```
isteinbrecher, line 51: Thanks for updating the doc style (also in other parts of this PR)!

I think this is slightly off the [google doc style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

```suggestion
        Args:
            geometry_set: ...
            rotation: ...
```

src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +37,52 @@ class LocSysCondition(BoundaryCondition):
...
+            list containing functions
+        update_node_position: Flag to enable the updated node position
+        use_consistent_node_normal: Flag to use a consistent node normal
         """
 
+        # Validate provided function array.
+        if function_array is not None:
+            ensure_length_of_function_array(function_array, 3)
+        else:
+            function_array = [0, 0, 0]
```
isteinbrecher, line 64:
```suggestion
        if function_array is not None:
            function_array = [0]
        ensure_length_of_function_array(function_array, 3)
```

src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +37,52 @@ class LocSysCondition(BoundaryCondition):
...
+        # Append the condition string with consistent normal type for line and surface geometry
+        if (
+            geometry_set.geometry_type is mpy.geo.line
+            or geometry_set.geometry_type is mpy.geo.surface
+        ):
+            condition_string = (
+                condition_string
+                + f" USECONSISTENTNODENORMAL {int(use_consistent_node_normal)}"
+            )
+
```
isteinbrecher, line 81: Maybe we can add this check here
```suggestion
elif use_consistent_node_normal:
    raise ValueError("The keyword use_consistent_node_normal only works for line and surface geometries")
```

tests/test_four_c.py
```diff
@@ -275,6 +276,25 @@ def test_meshpy_locsys_condition(
     # Add locsys condition with rotation
     input_file.add(LocSysCondition(beam_set["end"], Rotation([0, 0, 1], 0.1)))
 
+    # Add line Function with function array
+
+    fun_2 = Function("SYMBOLIC_FUNCTION_OF_SPACE_TIME t")
```
isteinbrecher, line 281: Thanks for being so thorough with the tests!

src/meshpy/four_c/function_utility.py
```diff
@@ -77,3 +77,26 @@ def create_linear_interpolation_function(
...
+
+
+def ensure_length_of_function_array(
+    function_array: list[Function], length: int = 3
+) -> None:
+    """
+    Performs size check of a function array and appends the function array to the given length, if a list with only one item is provided.
+    Args:
+        function_array: list with functions
+        length: expected length of function array
```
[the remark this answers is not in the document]
knarfnitram: I checked the pre-commit, there was a typo. Now it should be working. 
Can I add this here, or do you want a separate MR?

src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +37,52 @@ class LocSysCondition(BoundaryCondition):
 
     def __init__(
         self,
-        geometry_set,
-        rotation,
+        geometry_set: GeometrySet,
+        rotation: Rotation,
+        *,
+        function_array=None,
```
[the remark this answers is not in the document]
knarfnitram: I tried this, and basically this pollutes completely through meshpy. I remove all of them.

.pre-commit-config.yaml
```diff
@@ -54,7 +54,7 @@ repos:
     hooks:
       - id: docformatter
         name: Run docformatter (formatter for docstrings)
-        args: ["--in-place", "--config=./pyproject.toml", "-r", "src/meshpy/", "tests/"]
+        args: ["--in-place", "--config=./pyproject.toml", "--recursive", "src/meshpy/", "tests/"]
```
isteinbrecher, line 57: Are we sure this is needed \ solves any problem? If I look at the documentation of [docformatter](https://docformatter.readthedocs.io/en/latest/usage.html#use-from-the-command-line) "` -r, --recursive       drill down directories recursively`" this should have already been correct.

src/meshpy/four_c/function_utility.py
```diff
@@ -77,3 +77,26 @@ def create_linear_interpolation_function(
 
     variable_string = create_linear_interpolation_string(t, values, variable_name="var")
     return Function(f"{function_type} var\n" + variable_string)
+
+
+def ensure_length_of_function_array(function_array, length: int = 3):
+    """Performs size check of a function array and appends the function array to the given length, if a list with only one item is provided.
+    Args:
+        function_array: list with functions
```
isteinbrecher, line 85:
```suggestion
    """Performs size check of a function array and appends the function array to the given length, if a list with only one item is provided.

    Args:
        function_array: list with functions
```

src/meshpy/four_c/function_utility.py
```diff
@@ -77,3 +77,26 @@ def create_linear_interpolation_function(
...
+        raise ValueError("Provided array is empty.")
+
+    # extend items of function automatically if it is only provided once
+    if len(function_array) == 1:
+        return function_array * length
+
+    if len(function_array) != length:
+        raise ValueError(
+            f"The function array must have length {length} not {len(function_array)}."
+        )
```
isteinbrecher, line 101: I think `function_array is None` gives the wrong result.

```suggestion
    if len(function_array) == 0:
        raise ValueError("Provided array is empty.")

    # extend items of function automatically if it is only provided once
    if len(function_array) == 1:
        return function_array * length

    if len(function_array) != length:
        raise ValueError(
            f"The function array must have length {length} not {len(function_array)}."
        )
```

src/meshpy/four_c/function_utility.py
```diff
@@ -77,3 +77,26 @@ def create_linear_interpolation_function(
 
     variable_string = create_linear_interpolation_string(t, values, variable_name="var")
     return Function(f"{function_type} var\n" + variable_string)
+
+
+def ensure_length_of_function_array(function_array, length: int = 3):
```
isteinbrecher, line 82:
```suggestion
def ensure_length_of_function_array(function_array: List[Union[Int,Function]], length: int = 3):
```
Does this work?

src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +40,57 @@ class LocSysCondition(BoundaryCondition):
...
-        geometry_set: GeometrySet
-            Geometry that this boundary condition acts on.
-        rotation: Rotation
-            Object that represents the rotation of the coordinate system.
+        geometry_set: Geometry that this boundary condition acts on.
+        rotation: Object that represents the rotation of the coordinate system.
+        function_array: (list[Function])
+            list containing functions
+        update_node_position: Flag to enable the updated node position
+        use_consistent_node_normal: Flag to use a consistent node normal
```
isteinbrecher, line 59:
```suggestion
            geometry_set: Geometry that this boundary condition acts on.
            rotation: Object that represents the rotation of the coordinate system.
            function_array: List containing functions
            update_node_position: Flag to enable the updated node position
            use_consistent_node_normal: Flag to use a consistent node normal
```

src/meshpy/four_c/locsys_condition.py
```diff
@@ -34,25 +40,57 @@ class LocSysCondition(BoundaryCondition):
...
         """
 
+        # Validate provided function array.
+        if function_array is None:
+            function_array = [0, 0, 0]
+        elif isinstance(function_array, list) and all(
+            isinstance(item, Function) for item in function_array
```
isteinbrecher, line 66: Why do we need to check that the items are of `Function` type? Shouldn't that also work for `int`?

tests/test_four_c.py
```diff
@@ -237,8 +237,9 @@ def test_meshpy_locsys_condition(
 ):
     """Test case for point locsys condition for beams.
 
-    The testcase is similar to beam3r_herm2line3_static_locsys.dat, but
-    with simpler material.
+    The testcase is adapted from to beam3r_herm2line3_static_locsys.dat.
+    However it has a simpler material, one more rotation(line) with
+    functions.
```
isteinbrecher, line 242:
```suggestion
    However it has a simpler material, and an additional line locsys
    condition.
```

tests/test_four_c.py
isteinbrecher, line 281:
```suggestion
    # Add line Function with function array
    fun_2 = Function("SYMBOLIC_FUNCTION_OF_SPACE_TIME 2.0*t")
```
If you make the functions different, then the test will detect if they are for whatever reason switched.

tests/test_four_c.py
```diff
@@ -275,6 +276,25 @@ def test_meshpy_locsys_condition(
...
+            Rotation(
+                [0, 0, 1],
+                0.1,
+            ),
+            function_array=[fun_2],
+            update_node_position=True,
+            use_consistent_node_normal=True,
+        )
+    )
+
```
isteinbrecher, line 297:
```suggestion
    # Check if the LocSys condition is added correctly for a line with additional options.
    input_file.add(
        LocSysCondition(
            GeometrySet(beam_set["line"]),
            Rotation(
                [0, 0, 1],
                0.1,
            ),
            function_array=[fun_2],
            update_node_position=True,
            use_consistent_node_normal=True,
        )
    )

    # Check that the Locsys condition works with 3 given functions
    input_file.add(
        LocSysCondition(
            GeometrySet(beam_set["start"]),
            Rotation(
                [0, 0, 1],
                0.1,
            ),
            function_array=[fun_1, fun_2, fun_2],
            update_node_position=True,
            use_consistent_node_normal=True,
        )
    )
```

.pre-commit-config.yaml
knarfnitram: I tested it on my machine by adding a docstring containing an additional \n.

src/meshpy/four_c/function_utility.py
knarfnitram: No, by introducing this explict tpyes, I run into the error that we currently overwrite the function array sometimes either with `[fun1 fun2 fun3]` of type `Function` or we have a int array `[0 0 0]`. This leads to a complete refactor of `dbc_monitor.py`. We could avoid using the `Union[Int,Function]` by introducing a class named "ZeroFunction" as a special type of function, providing the`0`, but not does not add itself to the actual `----Funct` Section.

src/meshpy/four_c/function_utility.py
knarfnitram: this check had to be introduced when i tried to add the types `List[Union[Int,Function]]`, I think it makes sense to remove this completly.

src/meshpy/four_c/function_utility.py
isteinbrecher: I see, then lets just use `def ensure_length_of_function_array(function_array: List, length: int = 3):`, that should work.

isteinbrecher: I think you forgot the `function_array: List` change ;)

knarfnitram: Thanks for catching that.

isteinbrecher (approved):

src/meshpy/four_c/function_utility.py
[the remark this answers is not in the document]
davidrudlstorfer: Sorry if I do not have the correct overview yet - have you opened a second PR to fix the typo in the pre-commit hook @knarfnitram?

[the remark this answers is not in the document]
davidrudlstorfer: Oh it's further down below - sorry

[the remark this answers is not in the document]
knarfnitram: No, so Ivo tried it on his machine and the doc sting was not formatted correctly. However, for my machine, changing the `args`  for _docformatter_ from `"-r"` to `"--recursive"` applies it correctly.

PR #284, opened 2025-03-10
Add arc length attribute to space time nodes

isteinbrecher:
In some space-time formulations it is required to store the undeformed arc length coordinate along a beam filament in a node. This PR adds this functionality to the space-time module.


davidrudlstorfer (approved):
Thanks for also adding the test case!

Just one question regarding the general approach - is the arc length needed during the follow-up simulation run in 4C?

I needed the same thing for my beam formulation and the initial arc-length determination for all beam elements across all cores is already included in the code.

davidrudlstorfer:
Thanks for the info:)

Issue #277, opened 2025-03-05
Naming of private members, functions, classes in MeshPy


isteinbrecher:
At the moment we don't really have a consistent handling of private members, functions or classes in MeshPy. Let's use this issue to discuss how we want to go about this in the future and then slowly adapt this in future refactoring of MeshPy.

I would take the common approach of not making anything "truly" private, but rather just append a leading `_` to the respective name.

davidrudlstorfer:
I am also a bit unsure regarding this 

In Python the "privacy" level of variables, classes, ... is not that common as far as I know. Do you have any other repos that do that?

The problem I have is similar to #278 - there's no check that can enforce this - and then everything gets a bit "wilder Westen"

I would stay with the current approach so everything is uniform across the code base

One link regarding that stuff https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

isteinbrecher:
@davidrudlstorfer thanks for your thoughts.

> In Python the "privacy" level of variables, classes, ... is not that common as far as I know. Do you have any other repos that do that?

Just from my memory, I think many major python packages do this in some sort (`pyvista` and `scipy` are the first that come to my mind).

> The problem I have is similar to #278 - there's no check that can enforce this - and then everything gets a bit "wilder Westen"

As you pointed out, there is no way to check these things with a pre-commit hook since this is more of a coding style than an actual strict enforceable guideline (there is no way for the pre-commit hook to know what is supposed to be used from the outside and what is not). So we will have (and actually already do) "wilder Westen" anyway (we have some semiprivate methods (`_get_dat`, `_check_material`, ..., in the code base already). See my comment at the end.

> One link regarding that stuff https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

This is exactly what I was talking about, I like the approach of having class variables and methods that should not be called from outside as _semiprivate_, i.e., they can still be called from everywhere else, but the naming suggests that the user should really know what is happening.

To conclude, I don't think we should rigorously try to add this to the code base, but use it in places where it really makes sense, e.g., consistency checking function (`check_uniqueness`, `check_quaternion_constraint`, ...), functions where one can screw up the behaviour of the code when calling them at the wrong point (`replace_node`, `link_to_nodes`, ...). If we agree to do something like this, I think there is no need for any immediate action, but only consider this for future changes and refactoring.

[...]

PR #285, opened 2025-03-11
Improve README

isteinbrecher:
Follow up of #280

This PR restructures the readme, now the installation part is shorter and more precise. The optional dependencies are moved to an own section. Also a new _Developement_ section is added where information on the testing flags and Cython compile commands is stored - there we could also add coding guidelines in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@
 
 </div>
 
-MeshPy is a general purpose 3D beam finite element input generator written in `python`.
+MeshPy is a general purpose 3D beam finite element input generator written in Python.
 It contains advanced geometry creation and manipulation functions to create complex beam geometries, including a consistent handling of finite rotations.
 It can be used to create input files for the following finite element solvers (adaption to other solvers is easily possibly):
 - [4C](https://www.4c-multiphysics.org/) (academic finite element solver)
@@ -49,7 +49,13 @@ MeshPy is developed at the [Institute for Mathematics and Computer-Based Simulat
 - [How to cite MeshPy?](#how-to-cite-meshpy)
 - [Work that uses MeshPy](#work-that-uses-meshpy)
 - [Installation](#installation)
-  - [Optional dependency on ArborX](#optional-dependency-on-arborx)
+- [Optional dependencies](#optional-dependencies)
+  - [4C](#4c)
+  - [CubitPy](#cubitpy)
+  - [ArborX geometric search](#arborx-geometric-search)
+- [Developing MeshPy](#developing-meshpy)
+  - [Testing](#testing)
+  - [Cython geometric search](#cython-geometric-search)
 - [Contributing](#contributing)
 - [Authors](#authors)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,94 +155,83 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
-pip install .[cubitpy]
-```
->Note: This can also be combined with the development dependencies or an editable install with `pip install -e .[cubitpy,dev]`
+> Note: In some shells (e.g., `zsh`) quotation marks are required `pip install -e ".[dev]"`
 
-To check if everything worked as expected, run the standard tests with
+To check if everything worked as expected, run the standard test suite with (in the root directory)
 ```bash
 pytest
 ```
```
davidrudlstorfer, line 198: Everything in here should probably move to the development section - pytest also only works in the dev install (otherwise we have no pytest)

In the normal install one could execute an example from the example folder to check if it is working (and also for demonstration purposes)

README.md
```diff
@@ -149,94 +155,83 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
-pytest --CubitPy --performance-tests
-```
-executes the standard tests, the CubitPy tests and the performance tests.
->Note: the reference time values for the performance tests might not suite your system.
+## Optional dependencies
+
+### [4C](https://www.4c-multiphysics.org)
 
-We can also run tests in combination with 4C
+MeshPy can run 4C simulations directly from within a Python script, allowing for full control over arbitrarily complex simulation workflows. Fore more information, please have a look at the `meshpy.four_c.run_four_c` module. The MeshPy test suite including the 4C tests can be run with (in the root directory)
```
davidrudlstorfer, line 204:
```suggestion
MeshPy can run 4C simulations directly from within a Python script, allowing for full control over arbitrarily complex simulation workflows. Fore more information, please have a look at the `meshpy.four_c.run_four_c` module. The MeshPy test suite including the 4C tests can be run with
```

I would drop it because everything is being executed from the root directory as stated in the setup process

README.md
```diff
@@ -149,94 +155,83 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
 ```bash
 # 4C Tests require a path to a 4C executable
 export MESHPY_FOUR_C_EXE=<path_to_4C>
 pytest --4C
```
davidrudlstorfer, line 208: Again - the pytest functions will only work in dev mode => I would not mention it here but only in the developer section

README.md
```diff
@@ -149,94 +155,83 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
+```bash
+pip install .[cubitpy]
+```
+The MeshPy test suite including the CubitPy tests can be run with (in the root directory)
 ```bash
-pytest --CubitPy --exclude-standard-tests
+# CubitPy Tests require a path to a Cubit/Coreform installation
+export CUBIT_ROOT=<path_to_4C>
+pytest --CubitPy
 ```
```
davidrudlstorfer, line 224: Same here - only in the dev version

README.md
```diff
@@ -247,10 +242,32 @@ cd <path_to_meshpy>/build/geometric_search
 cmake ../../meshpy/geometric_search/src/
 make -j4
 ```
+The MeshPy test suite including the ArborX tests can be run with (in the root directory)
+```bash
+pytest --ArborX
+```
```
davidrudlstorfer, line 248: Same here

README.md
```diff
@@ -247,10 +242,32 @@ cd <path_to_meshpy>/build/geometric_search
...
 
-If the ArborX extension is working correctly can be checked by running the geometric search tests
+### Testing
+
+MeshPy provides a flexible testing system where additional tests can be enabled using specific flags. The following flags can be used with [pytest](https://pytest-cov.readthedocs.io/en/latest/config.html) to enable specific test sets:
+ - `--exclude-standard-tests`: Disables the default test suite
+ - `--4C`: Runs tests related to 4C integration
+ - `--ArborX`: Enables tests for ArborX-related functionality
+ - `--CubitPy`: Runs tests for CubitPy integration
+ - `--performance-tests`: Includes performance tests
```
davidrudlstorfer, line 259: That's a very nice overview 🙏🏻

README.md
```diff
@@ -247,10 +242,32 @@ cd <path_to_meshpy>/build/geometric_search
...
+
+These flags can be combined arbitrarily; for example, to run the 4C and ArborX tests but exclude the default test suite, use:
 ```bash
-pytest --ArborX --exclude-standard-tests
+pytest --4C --ArborX --exclude-standard-tests
+```
```
davidrudlstorfer, line 264: Maybe also add to this section how we can disable individual tests? We once wanted to add this to the readme

README.md
```diff
@@ -247,10 +242,32 @@ cd <path_to_meshpy>/build/geometric_search
...
+
+### Cython geometric search
+
+Some performance critical geometric search algorithms in MeshPy are written in [Cython](https://cython.readthedocs.io/en/stable/index.html). If Cython code is changed, it has to be recompiled. This can be done by running (in the root directory)
```
davidrudlstorfer, line 268:
```suggestion
Some performance critical geometric search algorithms in MeshPy are written in [Cython](https://cython.readthedocs.io/en/stable/index.html). If Cython code is changed, it has to be recompiled. This can be done by running
```

I would remove this everywhere in the entire file

davidrudlstorfer (commented):
Sorry for the large number of comments 🥲 let's just discuss what makes sense and what doesn't

davidrudlstorfer:
Should we also directly add #277 to the contributing section if we are at it?

> When working on MeshPy, use a leading underscore (_) to indicate functions, classes, and variables that are intended for internal use only. This is a coding convention rather than an enforced rule, so apply it where it improves code clarity, especially for functions that check consistency or modify internal structures.

ChatGPT proposes something like that which would fit

README.md
isteinbrecher: I know I said that, but I know now that a quick search for `pytest skip test` will point me to the correct documentation. So maybe not put it there, since this should not be used for "normal" development anyway.

isteinbrecher:
Thanks a lot for the review @davidrudlstorfer . I applied most of your suggestions. I moved the instal instructions to an own subsection `Install MeshPy from source` and only show how to install meshpy in editable mode (which I think is the only thing that makes sense if you install it from source). I would then simply add a section `Install MeshPy with pip` in the future - and move the `Jupyter` Sentence there.

I added the coding guidelines and moved all the test commands from the optional dependencies to the testing flags section.

Let me know what you think!

README.md
```diff
@@ -149,94 +158,71 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
-If you intend to use CubitPy install the additional dependency with
+Now you are able to use MeshPy. A good way to get started is by going through the examples
 ```bash
-pip install .[cubitpy]
+jupyter notebook examples/
 ```
->Note: This can also be combined with the development dependencies or an editable install with `pip install -e .[cubitpy,dev]`
 
-To check if everything worked as expected, run the standard tests with
+You can also run the MeshPy test suite to check that everything worked as expected (requires MeshPy to be installed with the `[dev]` dependencies)
```
ppraegla, line 197: Thanks for adding this. I just stumbled over this ;)

isteinbrecher: Do we have a new MeshPy user or just for testing purposes? 😉

README.md
davidrudlstorfer: Sounds good

README.md
```diff
@@ -149,94 +158,71 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
+  ```
 
-```bash
-pip install -e .[dev]
-```
->Note: In some shells (e.g., `zsh`) quotation marks are required `pip install -e ".[dev]"`
+### Install MeshPy from source
 
-If `cython` code is changed, it has to be recompiled. This can be done by running (in the root directory)
+You can install MeshPy directly from the source:
```
davidrudlstorfer, line 182: I would still split this up in the "normal" and "editable" install in my opinion. We worked hard to get it there - and for example @ppraegla just wanted to install it in the non-editable mode.

I still think it makes sense that people install it in the non-editable mode because it allows users to "look" at the code. If you install via pip in the future you have no direct chance to see the source code.

I would keep these sections pretty identical but split it up:

> You can either install MeshPy directly from the source in a non-editable and editable fashion like:
> 
> - non-editable:
> 
> ```
> git clone git@github.com:imcs-compsim/meshpy.git
> cd meshpy
> pip install .
> ```
> - editable:
> 
> This allows you to change the source code without reinstalling the module.
> ```
> git clone git@github.com:imcs-compsim/meshpy.git
> cd meshpy
> pip install -e .
> ```
> 
> After the install (either editable or non-editable) you are able to use MeshPy 🎉
> A good way to get started is by going through the examples
> ```bash
> jupyter notebook examples/
> ```
> 
> If you also want to execute the associated test suite check out our development (and link to it here) section.

This way we could split up the entire dev set up to the development section which is a lot cleaner in my opinion.
Also this easily extends with pip in the future as you mentioned (and this is a really good point you raised)

README.md
```diff
@@ -149,94 +158,71 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
...
 
-We can also run tests in combination with 4C
-```bash
-# 4C Tests require a path to a 4C executable
-export MESHPY_FOUR_C_EXE=<path_to_4C>
-pytest --4C
-```
+### [4C](https://www.4c-multiphysics.org)
+
+MeshPy can run 4C simulations directly from within a Python script, allowing for full control over arbitrarily complex simulation workflows. Fore more information, please have a look at the `meshpy.four_c.run_four_c` module.
```
davidrudlstorfer, line 206: I am no expert because I've never used it - do we need the 4C env variable to use the 4C module? Or is this really only necessary for the tests

README.md
```diff
@@ -247,10 +233,38 @@ cd <path_to_meshpy>/build/geometric_search
 cmake ../../meshpy/geometric_search/src/
 make -j4
 ```
+> Note: Currently ArborX only works if MeshPy is installed in _editable_ mode.
+
+## Developing MeshPy
+
```
davidrudlstorfer, line 239: And here I would add a general section regarding our pyproject setup

> If you plan to actively develop MeshPy or run the test suite of MeshPy you must install our optional developer dependencies like
> 
> ```
> pip install -e . [dev]
> ```
> 
> Note: The editable mode is required to generate the coverage report following each local pytest run.

davidrudlstorfer (commented):
@isteinbrecher thanks a lot for your really nice rework of the readme.

I would have another suggestion to more cleanly split up the install and development process. In my opinion it would make sense to be a bit more informative on why things have to be the way they are (like the difference between editable and non-editable mode). I think many developers do not know the exact difference between it.

I have to admit that a few months back I didn't know the real inner workings and only through working on it a lot I understood everything.

So maybe we can be a bit more elaborate here.

Overall I really like how everything looks - it makes a lot of sense!🤝

README.md
isteinbrecher: In testing, we need the variable for `run_four_c`. When using this function in a standard workflow, the default way should be to give the 4C path as an argument.

README.md
isteinbrecher: Good point, I added that!

README.md
isteinbrecher: God point, I adapted that as well!

davidrudlstorfer (approved):
Thanks a lot for going the extra mile and including all proposed changes!

I think now it should be really easy and straightforward to setup. I think now we have a real "roter Faden" throughout the readme and the sections are ordered in a meaningful manner.

Thanks and let's get this merged :rocket:

davidrudlstorfer:
(And this should now also close #277 )

Issue #278, opened 2025-03-05
Renaming all MeshPy internal imports


isteinbrecher:
We should reconsider the naming of our imports in MeshPy. With the new `src` layout imports will look similar to this (taken from `src/meshpy/four_c/input_file.py`):
```python
from meshpy.core.base_mesh_item import BaseMeshItemFull, BaseMeshItemString
from meshpy.core.boundary_condition import (
    BoundaryConditionBase,
    BoundaryConditionContainer,
)
from meshpy.core.conf import mpy
from meshpy.core.element import Element
from meshpy.core.geometry_set import GeometrySetContainer, GeometrySetNodes
from meshpy.core.mesh import Mesh
from meshpy.core.node import Node
from meshpy.core.nurbs_patch import NURBSPatch
from meshpy.utils.environment import cubitpy_is_available
```

This means that I can do the following in every script that uses MeshPy:
```python
from meshpy.four_c.input_file import Mesh, Node, NURBSPatch
```
There is nothing fundamentally wrong with that, however, I think it makes sense to encourage users to import the variables, functions, objects from the place where they are actually defined. We could help this by changing the all imports in MeshPy to something like:
```python
from meshpy.core.base_mesh_item import BaseMeshItemFull as _BaseMeshItemFull
from meshpy.core.base_mesh_item import BaseMeshItemString as _BaseMeshItemString
from meshpy.core.boundary_condition import BoundaryConditionBase as _BoundaryConditionBase
from meshpy.core.boundary_condition import BoundaryConditionContainer as _BoundaryConditionContainer
from meshpy.core.conf import mpy as _mpy
from meshpy.core.element import Element as _Element
from meshpy.core.geometry_set import GeometrySetContainer as _GeometrySetContaine
from meshpy.core.geometry_set import GeometrySetNodes as _GeometrySetNodes
from meshpy.core.mesh import Mesh as _Mesh
from meshpy.core.node import Node as _Node
from meshpy.core.nurbs_patch import NURBSPatch as _NURBSPatch
from meshpy.utils.environment import cubitpy_is_available as _cubitpy_is_available
```
I have seen other packages do it this way. This change should be easily scriptable and would help to reduce the number of accidential "wrong" imports.

Next steps:
- [ ] Do we want this?
  - [ ] Do we also want this for standard packages, e.g., `import numpy as _np`?
  - [ ] Create a script and adapt all imports

davidrudlstorfer:
[...]

PR #287, opened 2025-03-13
Rework MeshPy internal imports to force alias imports

isteinbrecher:
As proposed in #278 this PR shows how it would look if we renamed all MeshPy internal imports to have a leading underscore.

Turns out the simplest way to check this is to actually parse the code, which woks very well and is added as a pre-commit hook. We now enforce the imports to look like:
```python
from meshpy.core.mesh import Mesh as _Mesh
```
it is checked that the names match and only the leading underscore is added.

A few other things:
- For now only MeshPy imports are changed, we can think about changing standard library imports as well, although I am not sure if this really needed.
- I decided to keep the names in the documentation as they are intended.
- As far as I can see, the API documentation always displays the "true" name of a type, so this should also be OK.
- I rebased this on #285 to directly add to the new coding guidelines section

~~I am marking this as draft, as~~ I would like to think about actually doing this step. I like the safer handling of imports - but I also don't think that was a big issue to beginn with. I am not so sure about the fell of the new code, but maybe that just takes a while to get used to. What I don't like, is that we have to maintain our own pre-commit hook (although that script is not too complicated).

I would very much appreciate the opinion of the MeshPy developers and users.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,14 @@ repos:
         name: Run bandit (security linter for python)
         args: ["--configfile", "pyproject.toml", "--exclude", "tests/*"]
         additional_dependencies: ["bandit[toml]"]
+  - repo: local
+    hooks:
+      - id: check-imports
+        name: Check Python Imports
+        entry: python3 check_imports.py
+        files: ^src/meshpy.*\.py$
+        language: system
+        types: [python]
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.5
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -117,12 +117,12 @@ repos:
         args: ["--fix", "--select", "I,TID252"] # "I => sort imports, TID252 => ban relative imports
       - id: ruff-format # formatter
         name: Run ruff (formatter for Python)
-  - repo: https://github.com/crate-ci/typos
-    rev: v1.28.4
-    hooks:
-      - id: typos
-        name: Run typo checker
-        exclude: ".*\\.(nb|wls)$"
+  # - repo: https://github.com/crate-ci/typos
+  #   rev: v1.28.4
+  #   hooks:
+  #     - id: typos
+  #       name: Run typo checker
+  #       exclude: ".*\\.(nb|wls)$"
   - repo: https://github.com/pamoller/xmlformatter
     rev: v0.2.8
     hooks:
--- a/README.md
+++ b/README.md
@@ -257,6 +257,27 @@ pytest
 ### Coding guidelines
 
 - When working on MeshPy, use a leading underscore (`_`) to indicate functions, classes, and variables that are intended for internal use only. This is a coding convention rather than an enforced rule, so apply it where it improves code clarity, especially for functions that check consistency or modify internal states.
+- To avoid ambiguous or incorrect imports when using MeshPy as a library, internal imports must follow a strict aliasing convention as illustrated below:
+  ```python
+  # Not OK
+  import numpy  # No alias
+  import numpy as np  # Missing leading underscore
+  from numpy.linalg import norm  # No alias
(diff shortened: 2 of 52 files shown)


davidrudlstorfer:
Thanks for tackling this issue this fast!

I really like how it turned out and how easy it was doable. I have a few general questions/discussions:

- If we add this check I think we should also check for the standard library. It will be easier if all imports are handled the same
- What do you mean with `I decided to keep the names in the documentation as they are intended.`?
- Regarding the pre-commit hook I think it would make sense to slightly restructure the script. Most of the times pre-commit hooks just go over all of the code, and save everything to a some sort of data structure. In our case I think it would make sense to just store a dict with every filename and a list all things that are wrong. For example

```
dict = {"meshpy/file1": ["`from abc import def` must be changed to `from abc import def as _def`"]}
```

This way we could simply run through the check and print a nice summary at the end. With the current status the code is a bit hard to comprehend due to all the print statements on different levels on the one hand. On the other hand we could easily format everything nicely if we just print a dict at the end.
- I also only noticed at the end that the proposed pre-commit hook simply checks if everything is correct but does not do the actual correction of the imports. The first version of the code I generated with deepseek just fixed all the errors. I think from a user perspective this would be a lot better. Actually, how did you fix all imports? Did you replace all of them manually?

All in all, I think this is a very welcome change to MeshPy and I think it makes sense to go for this step.

isteinbrecher:
> - If we add this check I think we should also check for the standard library. It will be easier if all imports are handled the same

Makes sense, then all imports have the leading underscore. Then one exception can occur, if the imported name itself already starts with an underscore, in this case I would simply leave the existing name.

> - What do you mean with `I decided to keep the names in the documentation as they are intended.`?

When we refer to a name in the documentation (not type hints), e.g., `Mesh`, `InputFile`, `mpy`, ... I would still use the "true" names and not `_Mesh`, `_InputFile`, `_mpy`. Because the user reading the documentation will know these objects by their true name and not the alias.

> - Regarding the pre-commit hook I think it would make sense to slightly restructure the script. Most of the times pre-commit hooks just go over all of the code, and save everything to a some sort of data structure. In our case I think it would make sense to just store a dict with every filename and a list all things that are wrong. For example
> ```
> dict = {"meshpy/file1": ["`from abc import def` must be changed to `from abc import def as _def`"]}
> ```
> This way we could simply run through the check and print a nice summary at the end. With the current status the code is a bit hard to comprehend due to all the print statements on different levels on the one hand. On the other hand we could easily format everything nicely if we just print a dict at the end.

Makes perfect sense! I never wrote a pre-commit hook myself, so I am really happy for the input.

> - I also only noticed at the end that the proposed pre-commit hook simply checks if everything is correct but does not do the actual correction of the imports. The first version of the code I generated with deepseek just fixed all the errors. I think from a user perspective this would be a lot better. Actually, how did you fix all imports? Did you replace all of them manually?


I used the following script (directly copied from ChatGPT) to adapt the current code base [fix_imports.txt](https://github.com/user-attachments/files/19246436/fix_imports.txt). This works OK, but one still has to manually check the results (sometimes some imports are formatted wrong, double line imports cause issues during replacement, some names in the code themselves cause issues).

I agree that it would be nice to automatically fix the errors, however, I am very hesitant to do so. If we add a pre-commit hook that actually changes the code, then we have to make sure that we know that all corner cases are covered and that nothing unexpected happens. Otherwise someone's work might be overwritten. This means that we also have to test the pre-commit hooks to ensure that. To conclude, while it would be the best option to directly replace the wrong imports, I think we should definitely start of with a pure "check" (a nice error output definitely will help this). We can still add the replacement check in the future, if we feel comfortable with an implementation (that we then should also test).

I like the way this conversation is going! I add some open points to the issue #278 and then we can tackle them one by one. What would be the best option for this PR? Just merge the modified code base?

isteinbrecher:
The actual change to the code base is done with #293 . This PR is kept open so we can easily access the used pre-commit hook for the new import guideline, when actually creating the pre-commit-hook.

davidrudlstorfer:
This can be closed - it is linked in #294

PR #290, opened 2025-03-24
Fix bandit pre-commit hook

isteinbrecher:
This fixes a wrong configuration of the pre-commit setup that resulted in bandit not being run at all. Now it is run for the actual source code, but not the tests.

As far as I could figure out, the `# nosec` are required and kind of confirm that the developers checked that this is save, as bandit cannot ensure that. Theoretically one could modify the expected behaviour by injecting something into `MESHPY_FOUR_C_EXE` but that means that the system is already compromised, so I see no issue with skipping that warning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
     hooks:
       - id: bandit
         name: Run bandit (security linter for python)
-        args: ["-c", "pyproject.toml", "-r", "-ll", "src/meshpy/", "tests/"]
+        args: ["--configfile", "pyproject.toml", "--exclude", "tests/*"]
         additional_dependencies: ["bandit[toml]"]
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.5
--- a/src/meshpy/four_c/input_file.py
+++ b/src/meshpy/four_c/input_file.py
@@ -25,7 +25,8 @@
 import datetime
 import os
 import re
-import subprocess
+import shutil
+import subprocess  # nosec B404
 import sys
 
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -57,3 +57,6 @@ markers = [
   "cubitpy: tests in combination with CubitPy",
   "performance: performance tests of MeshPy"
 ]
+
+[tool.bandit]
+exclude_dirs = ["tests"]
```
davidrudlstorfer, line 62: Hmm I am a bit hesitant with this change.

My thought-process of the current setup was the following - I wanted to put all the tool settings into one central place - which is currently the `.pre-commit-config.yaml`. Therefore, if I am looking for a tool setting I will find it in there.

With the proposed changes - a few tool settings "live" in the `.pre-commit-config.yaml` while others are in the `pyproject.toml`. In my opinion we should either move all settings into one file to not spread it. In the end this will lead to duplicated settings and other problems.

If we want to put everything into one file we could also move everything into a `.bandit` file - but this would clutter up everything even more.

https://github.com/PyCQA/bandit/issues/499 here they wrote the correct syntax on how to include it into the pre-commit config file.

Maybe this works as well.

davidrudlstorfer (commented):
Thanks for fixing yet another bug in the pre-commit hook :D 

I am actually impressed on how many I introduced without knowing 🙃

I left one comment regarding a design decision I want to talk about

pyproject.toml
isteinbrecher: Good point, I had a look again and we can directly do this over the command line argument in the pre-commit file. I think that is the cleanest solution.

davidrudlstorfer (approved):
nice - thanks for further improving this!

PR #293, opened 2025-03-26
Replace MeshPy imports

isteinbrecher:
This PR tackles most of #278. We add the coding guideline to the readme and replace all occurrences within `/src`.

Two things to check:
- The typo pre-commit hook replaces `NDArray` with `ANDrray`
- The test `test_is_cubitpy_available` fails due to some wrong patching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,28 @@ pytest
 ### Coding guidelines
 
 - When working on MeshPy, use a leading underscore (`_`) to indicate functions, classes, and variables that are intended for internal use only. This is a coding convention rather than an enforced rule, so apply it where it improves code clarity, especially for functions that check consistency or modify internal states.
+- To avoid ambiguous or incorrect imports when using MeshPy as a library, internal imports must follow a strict aliasing convention as illustrated below:
+  <details>
+
+  <summary>Import guidelines</summary>
+
+  ```python
+  # Not OK
+  import numpy  # No alias
+  import numpy as np  # Missing leading underscore
+  from numpy.linalg import norm  # No alias
+  from meshpy.core.mesh import Mesh as _BeamMesh  # MeshPy imports have to be aliased with the same name — should be `_Mesh`
+
+  # OK
+  import numpy as _np
+  import sys as _sys
+  from pathlib import Path as _Path
+
+  import meshpy.core.conf as _conf
+  from meshpy.core.mesh import Mesh as _Mesh
+  from meshpy.core.node import Node as _Node, NodeCosserat as _NodeCosserat
+  ```
+  </details>
 
 ### Testing
 
@@ -293,6 +315,7 @@ If you contribute actual code, fork the repository and make the changes in a fea
 Depending on the topic and amount of changes you also might want to open an [issue](https://github.com/imcs-compsim/meshpy/issues).
 To merge your changes into the MeshPy repository, create a pull request to the `main` branch.
 A few things to keep in mind:
+- Read our [coding guidelines](#coding-guidelines).
 - It is highly encouraged to add tests covering the functionality of your changes, see the test suite in `tests/`.
 - To maintain high code quality, MeshPy uses a number of different pre-commit hooks to check committed code. Make sure to set up the pre-commit hooks before committing your changes
   ```bash
--- a/src/meshpy/abaqus/input_file.py
+++ b/src/meshpy/abaqus/input_file.py
@@ -22,15 +22,18 @@
 """This module defines the class that is used to create an input file for
 Abaqus."""
 
-from enum import Enum, auto
+from enum import Enum as _Enum
+from enum import auto as _auto
 
-import numpy as np
+import numpy as _np
 
-from meshpy.core.conf import mpy
-from meshpy.core.geometry_set import GeometrySet
-from meshpy.core.mesh import Mesh
-from meshpy.core.mesh_utils import get_coupled_nodes_to_master_map
-from meshpy.core.rotation import smallest_rotation
+from meshpy.core.conf import mpy as _mpy
+from meshpy.core.geometry_set import GeometrySet as _GeometrySet
+from meshpy.core.mesh import Mesh as _Mesh
+from meshpy.core.mesh_utils import (
+    get_coupled_nodes_to_master_map as _get_coupled_nodes_to_master_map,
+)
+from meshpy.core.rotation import smallest_rotation as _smallest_rotation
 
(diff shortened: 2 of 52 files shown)


isteinbrecher:
@davidrudlstorfer could you have a look at the failing test `test_is_cubitpy_available`?

davidrudlstorfer:
@isteinbrecher thanks for the update

Sure - I'll have a look

Regarding the typo hook - we can add a file with things to ignore

davidrudlstorfer:
Test is fixed - you can squash the commits if you like

README.md
```diff
@@ -257,6 +257,27 @@ pytest
...
+  from numpy.linalg import norm as _norm
+
+  from meshpy.core.conf import mpy as _mpy
+  import meshpy.core.conf as _conf
+  from meshpy.core.mesh import Mesh as _Mesh
+  from meshpy.core.node import Node as _Node, NodeCosserat as _NodeCosserat
+
+  import _module_with_leading_underscore  # Structures that already start with `_` are fine as-is
+  from module import _structure_with_leading_underscore  # Structures that already start with `_` are fine as-is
+  ```
```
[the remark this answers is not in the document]
isteinbrecher: Good point, I added it.

davidrudlstorfer:
And this would fix the typo hook

Unfortunately it is not possible to redirect the correct words directly via the args into that hook. Similar problem to #290 

but i would keep this approach

isteinbrecher:
Thanks, I found a similar solution for typos, but yours is better! I added the combined commits, should be ready for a final review.

Additionally I checked that there are no unused symbols (for some reasons mypy did not detect some) and I checked that only code is modified, not comments.

isteinbrecher:
Thanks for the fixes, nice to see this large change being done so fast!

davidrudlstorfer:
https://github.com/crate-ci/typos/discussions/1262 I've opened this discussion regarding the typo file - maybe we get an answer.

src/meshpy/abaqus/input_file.py
```diff
@@ -22,15 +22,18 @@
 """This module defines the class that is used to create an input file for
 Abaqus."""
 
-from enum import Enum, auto
+from enum import Enum as _Enum
+from enum import auto as _auto
```
davidrudlstorfer, line 26: Do want to leave it like this?

```suggestion
from enum import Enum as _Enum, auto as _auto
```

this would be more concise

src/meshpy/core/boundary_condition.py
```diff
@@ -22,12 +22,13 @@
 """This module implements a class to represent boundary conditions in
 MeshPy."""
 
-from meshpy.core.base_mesh_item import BaseMeshItemFull, BaseMeshItemString
-from meshpy.core.conf import mpy
-from meshpy.core.container import ContainerBase
+from meshpy.core.base_mesh_item import BaseMeshItemFull as _BaseMeshItemFull
+from meshpy.core.base_mesh_item import BaseMeshItemString as _BaseMeshItemString
```
davidrudlstorfer, line 26: Same here

src/meshpy/core/mesh.py
```diff
@@ -22,41 +22,48 @@
...
+from meshpy.core.geometry_set import GeometrySetBase as _GeometrySetBase
+from meshpy.core.geometry_set import GeometrySetContainer as _GeometrySetContainer
+from meshpy.core.material import Material as _Material
+from meshpy.core.node import Node as _Node
+from meshpy.core.node import NodeCosserat as _NodeCosserat
+from meshpy.core.rotation import Rotation as _Rotation
+from meshpy.core.rotation import add_rotations as _add_rotations
+from meshpy.core.rotation import rotate_coordinates as _rotate_coordinates
+from meshpy.core.vtk_writer import VTKWriter as _VTKWriter
+from meshpy.four_c.function import Function as _Function
```
davidrudlstorfer, line 53: Here the approach with an individual line for each import really elongates everything

davidrudlstorfer (commented):
Overall it looks good - I have one question/comment which affects almost all changes

README.md
```diff
@@ -257,6 +257,32 @@ pytest
...
+  from numpy.linalg import norm  # No alias
+  from meshpy.core.mesh import Mesh as _BeamMesh  # MeshPy imports have to be aliased with the same name — should be `_Mesh`
+
+  # OK
+  import numpy as _np
+  import sys as _sys
+  from numpy.linalg import norm as _norm
+
+  from meshpy.core.conf import mpy as _mpy
+  import meshpy.core.conf as _conf
```
davidrudlstorfer, line 278: while looking at the other comments - this is also a sub optimal example? If i import conf i do not need to import parts of conf again

README.md
```diff
@@ -257,6 +257,32 @@ pytest
...
+  from meshpy.core.mesh import Mesh as _Mesh
+  from meshpy.core.node import Node as _Node, NodeCosserat as _NodeCosserat
+
+  import _module_with_leading_underscore  # Structures that already start with `_` are fine as-is
+  from module import _structure_with_leading_underscore  # Structures that already start with `_` are fine as-is
```
davidrudlstorfer, line 283: and here - are there actually any packages starting with `_` - this would indicate that they should not be imported?

src/meshpy/abaqus/input_file.py
isteinbrecher: I would also prefer your solution (that's actually how I changed it), but the formatter apparently does not like it.

README.md
isteinbrecher: That is a very good point!

README.md
isteinbrecher: You are right, let's not give people bad suggestions.

src/meshpy/abaqus/input_file.py
davidrudlstorfer: You're right - somehow ruff pulls apart everything...

I looked into it if there is a magical setting - but I was not able to find anything. Also chatgpt did not "know" it

So if this is the standard - let's keep it this way so the magical linter and formatter is happy :D

davidrudlstorfer (approved):
Nice! Thanks for the discussions along this PR - looks good and let's get this merged :rocket:

isteinbrecher:
Thanks for the comments. Regarding the imports, I had a quick search and I could not find an easy solution to this, so I am fine with this less than optimal formatting (but since it is done automatically it should not really matter much.

PR #296, opened 2025-03-27
Add pre commit import style

### LINKED ISSUE #294: Add pre-commit hook to check import guidelines
We should add a pre-commit hook to check our input guidelines.

See #278 for the initial discussion.

In #287 a working version of the pre-commit hook exists, the main thing that is missing is providing a nice interface to the pre-commit framework and error output.

davidrudlstorfer:
Closes #294

Follows #278 #287 #293

The major part of the script is generated with deepseek AI

I've reworked the error detection for better error messages and a more reliable detection.

The error of the current code base looks like this

```
Check Python Imports.................................................................Failed
- hook id: check-imports
- exit code: 1

Found imports which do not align with our coding style:

    * src/meshpy/mesh_creation_functions/nurbs_geometries.py:26:0: Import 'compatibility' from 'geomdl' has incorrect alias '_compat'; should be '_compatibility'!

For more information on how to fix these issues, please refer to the contributing guidelines in our README.md.
```

The nice thing is that we now have line and column numbering. By clicking on the file this directly brings you to the related part in your IDE/Editor

The last question remaining is if we also want to check if the initial import already starts with an `_` and throw an error? This almost always means that a private function is imported. Currently this throws an error and suggests two under scores.

### DIFF
--- a/utils/check_python_imports.py
+++ b/utils/check_python_imports.py
@@ -1,0 +1,214 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2018-2025 MeshPy Authors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+"""This script checks if all Python imports within provided files align with
+our coding style, as described in the section "Coding guidelines" in the
+repository README.md."""
+
+import ast as _ast
+import sys as _sys
+from dataclasses import dataclass as _dataclass
+from pathlib import Path as _Path
+from typing import Optional as _Optional
+
+
+@_dataclass
+class Error:
+    """Class to store import errors."""
+
+    filename: str
+    message: str
+    line_nr: _Optional[int] = None
+    column_nr: _Optional[int] = None
+
+
+class ImportChecker(_ast.NodeVisitor):
+    """Class to check if Python imports align with our coding style."""
+
+    def __init__(self, filename: str) -> None:
+        """Initialize the ImportChecker.
+
+        Args:
+            filename: The filename of the current file (to store for errors)
+        """
+
+        self.errors: list[Error] = []
+        self.filename = filename
+
+    def visit_Import(self, node: _ast.Import) -> None:
+        """Visit an import statement and check if the import aligns with our
+        coding style.
+
+        Args:
+            node: The import node to check.
+        """
+
+        for alias in node.names:
+            if alias.asname is None:
+                self.errors.append(
+                    Error(
+                        self.filename,
+                        f"Import '{alias.name}' has no alias!",
+                        node.lineno,
+                        node.col_offset,
+                    )
+                )
+            elif not alias.asname.startswith("_"):
+                self.errors.append(
+                    Error(
+                        self.filename,
+                        f"Import '{alias.name}' has alias '{alias.asname}' that doesn't start with '_'!",
+                        node.lineno,
+                        node.col_offset,
+                    )
+                )
+        self.generic_visit(node)
+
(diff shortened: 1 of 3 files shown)


src/meshpy/mesh_creation_functions/nurbs_geometries.py
```diff
@@ -23,7 +23,7 @@
 
 import numpy as _np
 from geomdl import NURBS as _NURBS
-from geomdl import compatibility as _compat
+from geomdl import compatibility as _compatibility
```
isteinbrecher, line 26: I am not sure if we want to check this for imports from a standard library. Theoretically, it could be that we have two libraries that have objects with the same name, then we could not directly import them. I thought about that and think for standard imports it makes sense to just check the leading underscore. Also, we allow the import `import numpy as _np`, so I would only enforce the same names for MeshPy imports (maybe I should have chosen the examples better in the section I added to the read me).

utils/check_python_imports.py
```diff
@@ -0,0 +1,182 @@
...
+        """Initialize the ImportChecker.
+
+        Args:
+            filename: The filename of the current file (to store for errors)
+        """
+
+        self.errors: list[Error] = []
+        self.filename = filename
+
+    def visit_Import(self, node: _ast.Import) -> None:
```
isteinbrecher, line 55:
```suggestion
    def visit_import(self, node: _ast.Import) -> None:
```

Or is this name required since we are overloading a method?

utils/check_python_imports.py
```diff
@@ -0,0 +1,182 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+"""This script checks all Python imports within of provided files to align with
+our coding style, i.e., all internal imports are aliased with an underscore."""
```
isteinbrecher, line 23:
```suggestion
"""This script checks if all Python imports within provided files align with
our coding style, as described in the section "Coding guidelines" in the
repository readme."""
```

utils/check_python_imports.py
```diff
@@ -0,0 +1,182 @@
...
+                        Error(
+                            self.filename,
+                            f"Import '{alias.name}' has incorrect alias '{alias.asname}'; should start with '_'!",
+                            node.lineno,
+                            node.col_offset,
+                        )
+                    )
+        self.generic_visit(node)
+
+    def visit_ImportFrom(self, node: _ast.ImportFrom) -> None:
```
isteinbrecher, line 85:
```suggestion
    def visit_import_from(self, node: _ast.ImportFrom) -> None:
```

Or is this name required since we are overloading a method?

utils/check_python_imports.py
```diff
@@ -0,0 +1,182 @@
...
+                    Error(
+                        self.filename,
+                        f"Import '{alias.name}' from '{from_module}' has no alias!",
+                        node.lineno,
+                        node.col_offset,
+                    )
+                )
+            else:
+                expected_asname = "_" + alias.name
+                if alias.asname != expected_asname:
```
isteinbrecher, line 117: If we decide to allow the renaming of external imports, we have to add an if here to check if the module starts with `meshpy`.

isteinbrecher (commented):
Thanks a lot for the effort @davidrudlstorfer ! This is a lot better than my initial approach, and I really like the ability to click on the error and directly get to the line in the IDE.

I think the location in `/utils` makes sense for now, can always be changed if we want to restructure the utilities folder.

utils/check_python_imports.py
davidrudlstorfer: This overloads the regular method - therefore, keeping the current name

utils/check_python_imports.py
davidrudlstorfer: This overloads the regular method - therefore, keeping the current name

davidrudlstorfer:
@isteinbrecher thanks for the quick review! I've incorporated everything accordingly

The examples from the readme now produce this pre-commit output

```
Check Python Imports.................................................................Failed
- hook id: check-imports
- exit code: 1

Found imports which do not align with our coding style:

    * src/meshpy/mesh_creation_functions/beam_curve.py:24:0: Import 'numpy' has no alias!

    * src/meshpy/mesh_creation_functions/beam_curve.py:25:0: Import 'numpy' has alias 'np' that doesn't start with '_'!

    * src/meshpy/mesh_creation_functions/beam_curve.py:26:0: Import 'norm' from 'numpy.linalg' has no alias!

    * src/meshpy/mesh_creation_functions/beam_curve.py:28:0: Internal import 'Mesh' from 'meshpy.core.mesh' has incorrect alias '_BeamMesh'; should be '_Mesh'!

For more information on how to fix these issues, please refer to the contributing guidelines in our README.md.
```

The last question remaining is what I've written in the PR description - do we want to throw an error if someone imports something that already starts with an underscore? This currently results in a failing pre-commit hook suggesting to use another underscore (i.e. if I import `_test` the pre-commit hook suggests `__test`

isteinbrecher:
Looks very good!

Should we maybe add one example to the documentation?
```python
# OK
from math import sin as _sin_math
from numpy import sin as _sin_np
```

Sorry, I missed the point on the import of packages with a leading underscore. I would not chose the approach with the two leading underscores. Otherwise I am not really sure - maybe the cleanest way would be to just disallow it and if this becomes a problem with some external library we can reconsider this (my intuition is to never allow it for MeshPy because the requirement to do so would likely indicate a design flaw).

Maybe also add this to the examples:
```python
from numpy import _core # We don't allow the import of private items
```

PR #297, opened 2025-03-28
Improve interface and type hints of beam generic function

isteinbrecher:
This PR removes a not needed argument to `create_beam_mesh_function` (`add_sets`). Also type hints are added, some small bugs are fixed (for `node_positions_of_elements` `n_el` was not set).

### DIFF
--- a/src/meshpy/mesh_creation_functions/beam_generic.py
+++ b/src/meshpy/mesh_creation_functions/beam_generic.py
@@ -21,29 +21,42 @@
 # THE SOFTWARE.
 """Generic function used to create all beams within meshpy."""
 
+from typing import Callable as _Callable
+from typing import Dict as _Dict
+from typing import List as _List
+from typing import Optional as _Optional
+from typing import Tuple as _Tuple
+from typing import Type as _Type
+from typing import Union as _Union
+
 import numpy as _np
 
 from meshpy.core.conf import mpy as _mpy
+from meshpy.core.element_beam import Beam as _Beam
 from meshpy.core.geometry_set import GeometryName as _GeometryName
 from meshpy.core.geometry_set import GeometrySet as _GeometrySet
+from meshpy.core.material import MaterialBeam as _MaterialBeam
+from meshpy.core.mesh import Mesh as _Mesh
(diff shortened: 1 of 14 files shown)


src/meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -21,29 +21,41 @@
...
-    n_el=None,
-    l_el=None,
-    interval_length=None,
-    node_positions_of_elements=None,
-    add_sets=False,
-    start_node=None,
-    end_node=None,
-    vtk_cell_data=None,
+    beam_object: _Type[_Beam],
+    material: _MaterialBeam,
```
davidrudlstorfer, line 48: Should this be consistent, i.e., `_Type[_MaterialBeam]`? (As well as for others)

src/meshpy/utils/nodes.py
```diff
@@ -101,16 +103,12 @@ def get_min_max_coordinates(nodes):
...
+    a Node object as well as a GeometrySet object.
 
-    Args
-    ----
-    item:
-        This can be a GeometrySet with exactly one node or a single node object.
-    check_cosserat: bool
-        If a check should be performed, that the given node is a CosseratNode.
+    Args:
+        item: This can be a GeometrySet with exactly one node or a single node object.
```
davidrudlstorfer, line 111: Return statement is missing here

davidrudlstorfer (commented):
Nice! I really like our coding style within the code base - looks very good

src/meshpy/mesh_creation_functions/beam_generic.py
isteinbrecher: For material `_MaterialBeam` is fine as we expect a variable of that type. In the `beam_object` case we expect a beam class - so the name `beam_object` is very wrong. I replaced that to `beam_class` in the whole code base in an extra commit.

isteinbrecher: Good catch, I don't know how such a wrong name wasn't detected before.

davidrudlstorfer (approved):

PR #298, opened 2025-03-28
Add general functionality to compute arc lengths for created beam filaments

isteinbrecher:
This PR adds the functionality to allow for every beam creation function to also add the arc length along the created filament as nodal data. This is required for certain beam formulations, but might also be useful for pure mesh generation applications. 

This is added in a general and generic form to `create_beam_mesh_function`.


isteinbrecher:
Marking as draft, as some code checks and tests are still failing. Also, some additional new tests are required. But I still appreciate input / suggestions to the current changes, as they are more or less in a working state.

src/meshpy/core/element_beam.py
```diff
@@ -51,34 +51,49 @@ def __init__(self, material=None, nodes=None):
...
-            Returns the position and rotation of the beam along the local
-            coordinate xi.
-        start_node: Node
-            If this argument is given, this is the node of the beam at xi=-1.
-        end_node: Node
-            If this argument is given, this is the node of the beam at xi=1.
-        relative_twist: Rotation
-            Apply this relative rotation to all created nodes. This can be used to
-            "twist" the created beam to match the rotation of given nodes.
+        Args:
```
davidrudlstorfer, line 68: Thanks for updating the docstrings!

src/meshpy/core/element_beam.py
```diff
@@ -51,34 +51,49 @@ def __init__(self, material=None, nodes=None):
...
 
-        def check_node(node, pos, rot, name):
-            """Check if the given node matches with the position and
-            rotation."""
+        if arc_length_function is None:
+
+            def arc_length_function(xi):
+                """If no arc length function is provided, we define this dummy
+                function to return the default parameter None."""
+                return None
```
davidrudlstorfer, line 92: This still seems like a weird construct to me that we hand over the function and if no function is defined we create this dummy function?

src/meshpy/mesh_creation_functions/beam_generic.py
```diff
@@ -285,6 +329,28 @@ def get_relative_twist(rotation_node, rotation_function):
...
+                interval_b = node_positions_of_elements[i_el + 1]
+                element_start_in_space = interval_a * interval_length
+                element_end_in_space = interval_b * interval_length
+                return (
+                    0.5
+                    * (
+                        (1.0 - xi) * element_start_in_space
+                        + (1.0 + xi) * element_end_in_space
+                    )
+                ) + nodal_arc_length_start
```
davidrudlstorfer, line 349: Is it theoretically possible that the arc length function to this one?

If there should be multiple arc length functions maybe just define them like

```
def arc_length_function_1():
    blabla
    
def arc_length_function_2():
    blabla
```
    
and then just import them in the other file?

For me it's weird that we use the function as an argument

davidrudlstorfer (commented):
Nice changes - I only have a comment regarding the code structure which could be improved in my opinion

src/meshpy/core/element_beam.py
isteinbrecher: Yes, that was very weird. I already changed that. Now either the function or None is passed.

davidrudlstorfer: Is this included in the latest commits?

src/meshpy/mesh_creation_functions/beam_generic.py
isteinbrecher: I am not sure I fully understand your comment. Maybe it helps to clarify what is done here. The beam creation phase can be split into two levels. First calculate the positions of the boundaries between elements along the beam _filament_. Then for each _element_ we pass a functions that describes the beam centreline position and rotation (and now also arc length) along that element in the parameter space range [-1,1] - these functions have to be transformed to the element parameter space for each element, which is done here. By doing this split in _filament_ and _elements_, we don't care in the "filament" stage how many nodes a beam has, or how the beam element evaluates the data it requires to represent itself. We simply pass the functions to the beam element class and the rest is done there.

Does that make sense and or answers your question?

isteinbrecher: An additional note: I am not sure if this is the best approach to do all of that, but it has proven to be useful in the past.

davidrudlstorfer: I'll reformulate the question:

The calculation of the arc length from the parameter xi will always be identical. Why not make it a free function but inline it here again?

You could just create an entirely free function which takes the necessary arguments. This way we would also be able to simply test the function.

I just do not understand why such a function is an argument for the create beam functionality - could this not be handled internally?

src/meshpy/core/element_beam.py
isteinbrecher: No, it wasn't - should be now.

src/meshpy/mesh_creation_functions/beam_generic.py
davidrudlstorfer: Thanks for looking into it:)

isteinbrecher:
Can be closed in favour of #300

PR #300, opened 2025-04-02
Add general functionality to compute arc lengths for created beam filaments (alternative to #298)

isteinbrecher:
This provides an alternative implementation to #298. If the arc length should be evaluated, we require beam functions to provide the arc length along their centreline. This reduces the changes compared to #298.


isteinbrecher:
@davidrudlstorfer I think this is cleaner than #298 and I would prefer this approach. I am still not 100% happy, but that is maybe just because of the whole beam creation workflow that has a lot of functionality - and became quite loaded over time.

davidrudlstorfer (dismissed):
I think this implementation is a lot cleaner than #298

In my opinion you can merge this:)

Thanks for going the extra mile and improving the implementation

isteinbrecher:
Thanks for your suggestion, that really helped. I did come slight documentation changes, will merge once the pipeline passes.

davidrudlstorfer (approved):

PR #302, opened 2025-04-03
Convert dat files to yaml

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher this would be ready for review and marks the first PR towards .yaml based input :rocket:

Some explanations:

 - The first commit fixes all .dat files which are not readable by 4C. Note, that in the future our test files must not be compatible with 4C. When we write out our .yaml input file with fourcipp we have the option to either validate the input file (checks if all the input is compliant with 4C) or not validate it. Therefore, we can still write non valid input files. But the major of these checks is then useless because we have a working 4C input parser
 
 - The changes to the beam potential are based on my changes in the input. This will be updated once the move to .yaml is finished.
 
 - The second commit converted all .dat files to .yaml with the help of this https://github.com/4C-multiphysics/4C/issues/598#issuecomment-2774776846 script
 
 - The third commit deletes all .dat files
 
 - The final commit temporarily disables all failing pytest tests and lowers the necessary coverage. This way we still have a working pipeline

tests/reference-files/test_meshpy_userdefined_boundary_condition.dat
```diff
@@ -8,8 +8,6 @@
 // -----------------------------------------------------------------------------
 -----------------------------------------------------------------------MATERIALS
 MAT 1 MAT_BeamReissnerElastHyper YOUNG -1.0 POISSONRATIO 0.0 DENS 0.0 CROSSAREA 3.141592653589793 SHEARCORR 1 MOMINPOL 1.5707963267948966 MOMIN2 0.7853981633974483 MOMIN3 0.7853981633974483
--------------------------------------------------------------USER SECTION FOR BC
-E 1 test
```
isteinbrecher, line 12: This should not be removed as it tests the functionality that we can define any boundary condition and not just the ones defined in `core.conf.BoundaryCondition`. So let's switch this here to a BC that is valid in 4C but not implemented in MeshPy. For example:
```
--------------------------------------------DESIGN VOL ALE DIRICH CONDITIONS
E 1 NUMDOF 3 ONOFF 0 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
```

isteinbrecher (commented):
Thanks @davidrudlstorfer for this PR. I especially like how you structured the commits, makes the review very easy! I just have one small comment.

tests/reference-files/test_meshpy_userdefined_boundary_condition.dat
davidrudlstorfer: Thanks for the info regarding this test - prior to your explanation I was not really sure what this actually tested.

Changed in latest push.

isteinbrecher (approved):

PR #306, opened 2025-04-07
Compare dictionaries in testing framework

isteinbrecher:
When switching to yaml, we need to do a lot of dictionary comparisons in the testing framework, this we need to improve the one we currently have.

Some basic requirements:
- We want to be able to compare `list()` with `np.array`.
- Sometimes we get numbers as "numpy" numbers and sometimes as `float`. Maybe this is not ideal and should changed on the implementation side, but for now we need to compare them.
- String comparison is a little bit tricky. We still have legacy strings at the moment, so they should be compared with the `compare_strings` function to account for numeric differences - maybe if we have two strings, try if they are equal, if not, try with `compare_strings`, if both fail raise an error.

Nice to have:
- If one of the dictionaries originates from a yaml file, open a diff in VSCode

One solution could be to have 3 functions
- `compare_dict` Check that keys match and then compare the values with `compare_value`
- `compare_list` Check that length matches and then compare each value with `compare_value`
- `compare_value` Check what types we have and choose appropriate comparison function, e.g., string, np.array, ... . For `dict` and `list` call `compare_dict` and `compare_list`, respectively.

I think we already have most of the functionality to do this ourselves, but alternatively we could look at [deepdiff ](https://zepworks.com/deepdiff/current/).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,6 +32,7 @@ from typing import Callable, Optional, Union
 import numpy as np
 import pytest
 import vtk
+import yaml
 from _pytest.config import Config
 from _pytest.config.argparsing import Parser
 from vtk_utils.compare_grids import compare_grids
@@ -224,7 +225,7 @@ def get_corresponding_reference_file_path(
     def _get_corresponding_reference_file_path(
         reference_file_base_name: Optional[str] = None,
         additional_identifier: Optional[str] = None,
-        extension: str = "dat",
+        extension: str = "4C.yaml",
     ) -> Path:
         """Get path to corresponding reference file for each test. Also check
         if this file exists. Basename, additional identifier and extension can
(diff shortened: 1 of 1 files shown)


isteinbrecher:
@davidrudlstorfer the initial state is my hacky solution to get this to work, feel free to push over it.

davidrudlstorfer:
@isteinbrecher sorry for not looking into this earlier as discussed. I will look into this tomorrow so we complete the initial switch to yaml as soon as possible.

Regarding deepdiff I am a bit hesitant if it makes so much sense. In a first search I was not able to find anything regarding the comparison with tolerances, i.e., relative and absolute.

And although our new input is a standard format (yaml) there will probably still be a few edge cases (with legacy sections) or something else.

If we fully rely on a third party package we might not be able to find solutions to some of these things.

In my opinion our comparison function in the test suite is still very nicely maintained and not too much. Therefore, I would keep the current approach.

isteinbrecher:
I also prefer the “own” implementation here. We already have all the base comparison functions so we just need to plug them together in an organized manner.

A couple of things that came to my mind while working on the yaml files last week:
- When the key comparison of two dict fails, it would be nice to  also print the keys that are unique in dict1 and dict2 respectively. Makes identifying what goes wrong a lot easier.
- I really like that for dat files a simple diff in vs code opens to visually check what went wrong. This could also be easily doable for yaml files, we just have to write them to a file in “ordered” fashion - with #308 one could then even directly edit the reference files to adapt them.

These points are not a must right now, but I think we should aim for them in the future to make fixing failing tests easier.

davidrudlstorfer:
> * I really like that for dat files a simple diff in vs code opens to visually check what went wrong. This could also be easily doable for yaml files, we just have to write them to a file in “ordered” fashion - with [Store yaml reference files in ordered fashion #308](https://github.com/imcs-compsim/meshpy/issues/308) one could then even directly edit the reference files to adapt them.


I do not fully understand why comparing in an ordered manner improves anything here (also #308)? Maybe you could ellaborate here

isteinbrecher:
For pure comparison functionality, the order of keys doesn’t matter — two YAML files are considered equal regardless of key order. However, when viewing diffs (e.g. in VS Code, Git, or GitHub), maintaining a consistent key order should significantly improve readability. If the content remains the same but the key order changes, the diff may misleadingly suggest large changes, making it harder to see what’s actually different. So this would be a pure "make development easier" feature and nothing related to the actual code or testing functionality.

davidrudlstorfer:
I've discussed this matter with @gilrrei and we came to the conclusion that it makes sense that this functionality moves to fourcipp.

A general function will allow us to simply compare dicts (and input files, and fourcipp input objects) with numerical tolerances for numbers (relative and absolute).

This way we can simply reuse this functionality here and do not have to reimplement it again. Probably this will be useful for CubitPy as well.

davidrudlstorfer:
@isteinbrecher I think we can then close this PR. The identical commits are in #310 and in the very near future we will use the dict comparison functionality from fourcipp within meshpy

PR #307, opened 2025-04-07
Remove 4C capabilities from space-time functionality

isteinbrecher:
Remove the 4C input file capabilities of the space time functionality, as the space time functionality is fully covered by comparing the results from `mesh_to_data_arrays`. When the space-time functionality will eventually make its way to 4C, the exact format for the nodes and elements will have to be changes anyway.


davidrudlstorfer (approved):
Sounds good 🚀

PR #313, opened 2025-04-10
Fix yaml beam potential

davidrudlstorfer:


src/meshpy/four_c/beam_potential.py
```diff
@@ -141,23 +141,23 @@ def add_header(
...
-            "NUM_GAUSSPOINTS": gauss_points,
+            "N_INTEGRATION_SEGMENTS": integration_segments,
+            "N_GAUSS_POINTS": gauss_points,
             "POTENTIAL_REDUCTION_LENGTH": potential_reduction_length,
             "AUTOMATIC_DIFFERENTIATION": automatic_differentiation,
         }
 
         if regularization_type is not None:
-            settings | {
+            settings = settings | {
```
davidrudlstorfer, line 154: This was actually an error

@isteinbrecher do you know if this is more often in the repo? I tried my best to find other appearances but this was the only one as far as I can tell

isteinbrecher: I think this was something I tried without testing it 😬.

davidrudlstorfer:
Nice - thanks for the fast review - I am already at fixing the header tests:)

PR #314, opened 2025-04-11
Fix header tests

davidrudlstorfer:


isteinbrecher:
Yes, this `None` thing has also caught me off guard. Maybe in the future we can use a custom dumper for `None` to align with what 4C expects - likely that would be on the 4CIPP side, and is not relevant for this PR anyways.

I think this construct was to check that the `overwrite_option` was passed through correctly and overwrites the default values (default `set_beam_to_solid_meshtying` values, not default 4C values). It it is easy I would leave it like this, if it makes trouble we can also remove this - if we keep it could you add a short comment like
```python
# This is to test that the overwrite_option argument is handled correctly
```

davidrudlstorfer:
So this would be ready now @isteinbrecher 

The underlying problem was that with the conversion of the .dat to .yaml files 4C dismisses some default variables.

I've now readded them to the reference files and everything works

In the entire test suite now only 6 cubitpy tests are disabled with 

```python
@pytest.mark.skip(
    reason="Temporarily disabled due to switch to .yaml based input files - check if test is necessary and fix"
)
```

All other tests are now activated and running

isteinbrecher (approved):
Thanks for taking care of this @davidrudlstorfer

davidrudlstorfer:
Thanks for the quick look - next PR is already finished - purging dat.

I'll push in a minute

PR #315, opened 2025-04-11
Purge all mentionings of dat

davidrudlstorfer:


src/meshpy/core/base_mesh_item.py
```diff
@@ -47,36 +47,12 @@ def __init__(self, data=None, comments=None):
...
-            return_list.extend(self.comments)
-
-        # Return final list.
-        return_list.extend(data)
-        return return_list
-
-    def _get_dat(self):
-        """Return the content of this object as either a list or a str."""
-        return self.data
-
```
davidrudlstorfer, line 73: I was able to completely remove this from the BaseMeshItem

src/meshpy/abaqus/beam.py
```diff
@@ -69,19 +68,3 @@ def generate_abaqus_beam(beam_type: str):
             "n_dim": n_dim,
         },
     )
-
-
-class AbaqusBeamMaterial(_MaterialBeam):
```
isteinbrecher, line 74: Nice cleanup!

src/meshpy/core/base_mesh_item.py
isteinbrecher: I thought we should be able to do so - great that it actually works!

src/meshpy/core/base_mesh_item.py
```diff
@@ -47,36 +47,12 @@ def __init__(self, data=None, comments=None):
...
 class BaseMeshItemFull(BaseMeshItem):
     """Base class for all objects that are related to a mesh and are fully
     created in MeshPy."""
 
 
 class BaseMeshItemString(BaseMeshItem):
-    """Base class for all objects that are imported from a dat file as a plain
-    string."""
+    """Base class for all objects that are imported from an input file as a
+    plain string."""
```
isteinbrecher, line 58: Nice side note: We should now be able to remove the `BaseMeshItemFull` and `BaseMeshItemString` completely! But let's do that in a different PR.

src/meshpy/core/boundary_condition.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, geometry_set, bc_type=None, **kwargs):
         ----
         geometry_set: GeometrySet, int
             Geometry that this boundary condition acts on. An integer can be
-            given, in the case a dat file is imported. This integer is only
+            given, in the case an input file is imported. This integer is only
             temporary and will be replaced with the GeometrySet object.
```
isteinbrecher, line 43: With the restructuring in the `InputFile`, we now directly pass the correct geometry set here, so no need for integers any more.
```
        geometry_set: GeometrySet
            Geometry that this boundary condition acts on.
```

src/meshpy/core/coupling.py
```diff
@@ -60,12 +60,12 @@ def __init__(
             physical position.
         check_at_init: bool
```
isteinbrecher, line 61: Similar to the previous comment, this might be obsolete, but requires some changes in the code. I will create an issue, for this PR we can leave it like this.

src/meshpy/core/rotation.py
```diff
@@ -324,15 +324,15 @@ def __eq__(self, other):
             return object.__eq__(self, other)
 
     def get_dat(self):
```
isteinbrecher, line 326: Is this the only remaining `*_dat` function? We can keep this one, and clean that up once we move to full `double` output, should be very easy then.

src/meshpy/four_c/input_file.py
```diff
@@ -395,11 +395,9 @@ def get_dict_to_dump(
         Args:
             header:
                 If the header should be exported to the input file files.
-            dat_header:
-                If header lines from the imported dat file should be exported.
-            append_script_to_header:
+            add_script_to_header:
                 If true, a copy of the executing script will be added to the input
-                file. This is only in affect when dat_header is True.
+                file. This is only in affect when header is True.
```
isteinbrecher, line 400:
```suggestion
                file. This is only in affect when `add_script_to_header` is True.
```

src/meshpy/four_c/input_file.py
```diff
@@ -615,7 +611,7 @@ def get_number_of_coupling_conditions(key):
 
         # Add additional element sections (e.g. STRUCTURE KNOTVECTORS)
         # We only need to to this on the "real" elements as the imported ones already have their
-        # dat sections.
+        # input sections.
```
isteinbrecher, line 614:
```
        # Add additional element sections, e.g., for NURBS knot vectors.
```
The rest of the comment is outdated.

isteinbrecher (commented):
Thanks for the quick changes, I left some comments.

isteinbrecher:
I am actually surprised that we did not have more mentions of `dat`!

src/meshpy/core/rotation.py
davidrudlstorfer: Yes, that's the only one as far as my search revealed

src/meshpy/four_c/input_file.py
davidrudlstorfer: I would quickly leave it like this because I am already at updating this in its entirety.

The PR will also follow soon

Then we can again dump all the MeshPy/CubitPy information into the header/title/future metadata section and the script is added as a comment at the bottom

davidrudlstorfer:
I think the two updated comments are the only necessary change - as I've written I'll update the entire header thing in a separate PR which is already in work

src/meshpy/four_c/input_file.py
isteinbrecher: Works for me.

isteinbrecher (approved):

PR #316, opened 2025-04-11
Increase coverage again after yaml switch

davidrudlstorfer:


isteinbrecher (approved):
@davidrudlstorfer what do you think, should we increase this again after all tests are running?

davidrudlstorfer:
I think we have to adjust either way soon. It will probably again change once we use fourcipp in the input

Issue #304, opened 2025-04-03
Discuss new `InputFile` interface for 4C YAML


isteinbrecher:
With #135 we will switch to YAML. This will mostly result in a complete rework of the current `InputFile` class. Let's discuss the interface of the new `InputFile` (or whatever name it will have) class here, as this will be one of the main points that will change for users.

Currently the interface looks something like this:
```python
input_file = InputFile()

material = MaterialReissner()
input_file.add(material)

create_beam_mesh_line(input_file, beam_type, material, p1, p2)

mesh = Mesh()
create_beam_mesh_line(mesh, beam_type, material, p3, p4)

input_file.add(mesh)

input_file.add("""
--------------------------------------------------------------STRUCTURAL DYNAMIC
DYNAMICTYPE                           Statics
TIMESTEP                              0.05
NUMSTEP                               20
""")

input_file.write_input_file(path)
```

I think this general approach still makes sense, but I am open to any changes that can improve the code design behind everything and/or the usability.

Some thoughts about a possible new `InputFile`:

- I think we only need the following class variables in the input file:
  ```
  class InputFile:
      def __init__(self, ...):
          self.yaml = {} # either a pure dictionary or a FourCIPP object
          self.mesh = Mesh()
          self.nox_xml = None
          self._nox_xml_file = None
  ```
- If the `InputFile` does not derive from `Mesh` any more, we would have to change the first `create_beam_mesh_line` call in the example to:
  ```diff
  - create_beam_mesh_line(input_file, beam_type, material, p1, p2)
  + create_beam_mesh_line(input_file.mesh, beam_type, material, p1, p2)
  ```
  I am very much in favour of this change.
- What do we want to do about the `.add` methods? We at least need the options to add parameters and meshes directly to the `InputFile`:
  ```python
  input_file.add(mesh)
  input_file.add({
      "STRUCTURAL DYNAMIC": {
          "DYNAMICTYPE": "Statics",
          "TIMESTEP": 0.05,
          "NUMSTEP": 20
      }
  }
  ```
  Do we want to also be able to add other objects, e.g., boundary conditions, functions, ...? I would be fine to pipe all additions of non-dictionaries to `input_file.mesh.add`, but my opinion on this is not strong.

isteinbrecher:
I am not sure what exactly you mean with "treat the input file like a "discretization". The `Mesh` object can be considered as a discretisation. In my understanding the InputFile combines the information of the discretisation with the parameters not directly related to the mesh.

[...]

PR #324, opened 2025-04-16
Allow import of geometry sets with non-consecutive numbering

isteinbrecher:
This PR adds the functionality to read in non-consecutive numbered geometry sets from an existing input file.

Also, the way we construct boundary conditions is improved, by simply passing the correct geometry set instead of relying on the boundary condition to find "it's" geometry set.


isteinbrecher:
@davidrudlstorfer if this introduces too many merge conflicts with your current work, we can also wait with this PR.

src/meshpy/four_c/input_file.py
```diff
@@ -219,6 +217,10 @@ def sections_to_mesh(self):
         """Convert mesh items, e.g., nodes, elements, element sets, node sets,
         boundary conditions, materials, ...
 
+        Note: In the current implementation we cannibalize the mesh sections in
+            the self.sections dictionary. This should be reconsidered and be done
```
davidrudlstorfer, line 221: What do you mean with cannibalize?😅

isteinbrecher: That we delete the sections when creating a mesh. This comment should be resolved with the proposed way to read meshes in #304.

davidrudlstorfer (approved):

PR #326, opened 2025-04-16
Refactor boundary condition

isteinbrecher:
Follows #325 #324 (this PR already includes the changes made in those PRs).

This PR refactors the boundary condition framework. Previously, we had both a general boundary condition base class and a separate boundary condition in the 4C folder. However, the 4C implementation included general-purpose functionality that wasn’t truly specific to 4C.

This PR proposes using the general boundary condition class for 4C meshes as well, which seems reasonable since all necessary data can be stored in the `BaseMeshItem.data` attribute.

Once this is merged, this should serve as a starting point for refactoring the `Coupling`.

Other points:
- I decided to rename the argument `bc_dict` to `data`, this should not break any interfaces, but internally is more consistent and general in my opinion.
- The "create a `BoundaryCondtion` from an existing yaml file functionality was moved to `input_file.py` which makes much more sense in my opinion.
- With the new structure, a few places where we had to do imports inside a function to avoid cyclic dependencies was resolved.


src/meshpy/four_c/input_file.py
```diff
@@ -575,7 +609,14 @@ def _dump_mesh_items(yaml_dict, section_name, data_list):
...
             for item in data_list:
-                item_dict_list.extend(item.dump_to_list())
+                if hasattr(item, "dump_to_list"):
+                    item_dict_list.extend(item.dump_to_list())
+                elif isinstance(item, _BoundaryCondition):
+                    item_dict_list.append(
+                        {"E": item.geometry_set.i_global, **item.data}
+                    )
+                else:
+                    raise TypeError(f"Could not dump {item}")
```
isteinbrecher, line 619: @davidrudlstorfer I think something like this could be a good way to improve modularisation. For general cases that require almost no 4C specific logic, we can implement the conversion to a dict right here. For special cases, we can use special classes that define complex functionality in the `.dump_to_list` method.

For example, we could define a general way to convert solid elements to 4C, but still have specialised classes that take care of beam elements in 4C.

src/meshpy/four_c/input_file.py
```diff
@@ -57,6 +60,31 @@
...
+        _mpy.bc.beam_to_solid_surface_contact,
+        _mpy.bc.beam_to_solid_volume_meshtying,
+    ) or isinstance(bc_key, str):
+        return _BoundaryCondition(geometry_set, data, bc_type=bc_key)
+    elif bc_key is _mpy.bc.point_coupling:
+        return _Coupling(geometry_set, bc_key, data, check_overlapping_nodes=False)
+    else:
+        raise ValueError("Got unexpected boundary condition!")
+
+
```
davidrudlstorfer, line 87: Would it make sense that this moves again in `four_c/boundary_condition.py`?

src/meshpy/four_c/input_file.py
```diff
@@ -581,7 +609,14 @@ def _dump_mesh_items(yaml_dict, section_name, data_list):
 
             item_dict_list = yaml_dict[section_name]
             for item in data_list:
-                item_dict_list.extend(item.dump_to_list())
+                if hasattr(item, "dump_to_list"):
+                    item_dict_list.extend(item.dump_to_list())
+                elif isinstance(item, _BoundaryCondition):
+                    item_dict_list.append(
+                        {"E": item.geometry_set.i_global, **item.data}
+                    )
```
davidrudlstorfer, line 617: I must confess that I do not really like this construct. This check is more or less a "runtime" check if the item has the attribute or not. In my opinion we should aim for a general way to dump stuff. Maybe I do not fully understand your intention here.

davidrudlstorfer (commented):
Nice cleanup!

src/meshpy/four_c/input_file.py
isteinbrecher: I don't think so, as this is more related to the reading of input files. Maybe we can make an input_file_utils.py file? I am not sure if I would want to have a `boundary_condition.py` file that does not contain a boundary condition.

src/meshpy/four_c/input_file.py
isteinbrecher: I’m not sure it’s feasible (or even possible) to have a general dump method across the board. In MeshPy, we can roughly distinguish between two types of mesh objects:

First, there are the ones defined in `core`, e.g., Cosserat nodes, control points, and boundary conditions. These objects are general-purpose and can be reused across different solvers like 4C, Abaqus, or AceFEM. For these, a universal dump function doesn’t really make sense, since the output format depends on the solver in use. For example, we currently have a `dump_to_list` method in `core.Node` that’s tied specifically to 4C input files, which really shouldn’t be there. The same goes for a lot of other objects in `core`.

On the other hand, there are solver-specific objects that live in their respective folders, such as `four_c` or `abaqus`. These objects include logic that is only relevant to that specific solver, e.g., the torsion-free beam element in 4C needs to check that it is straight in its initial configuration. Because these objects are directly related to a specific solver, defining a `dump` method for them makes sense.

So I think we have a few options going forward:
- Keep this proposed split, where only solver-specific objects have a dump method, and core objects do not
- Have no dump methods at all and implement all the dump logic in `four_c.InputFile` and `abaqus.InputFile` (actually for Abaqus this is the way we currently do it)
- Find a way that we can attach a `dump` function to `core` objects at runtime, depending on the active solver

davidrudlstorfer: @isteinbrecher thanks a lot for the detailed explanation - now I finally understood it 🙏🏻

> Have no dump methods at all and implement all the dump logic in four_c.InputFile and abaqus.InputFile (actually for Abaqus this is the way we currently do it)

But then this would make the most sense wouldn't it? We then would have our "mesh to dict" function and this function automatically converts all the MeshPy-4C objects into a dict. I somehow like it more if the functions are in one central place and not scattered around MeshPy

isteinbrecher: I agree, after thinking about my comment, this makes the most sense. I would then propose to still go ahead with this PR, as it removes the `dump` function from the `BoundaryCondition`.

davidrudlstorfer (approved):

Issue #308, opened 2025-04-09
Store yaml reference files in ordered fashion


isteinbrecher:
The testing reference files should be stored with the dict keys ordered. This makes viewing changes on them (locally or in GitHub) easier.

isteinbrecher:
Let's rewrite all reference files again once we have `fourcipp`, then hopefully all formatting issues should be resolved and allow for a comparison of failed tests.

PR #327, opened 2025-04-16
Full precision double output in input and testing reference files

isteinbrecher:
This PR changes the way we write out floats to the input file. Now they are written in full double precision and not some arbitrary user-specified precision.

Most reference files are recreated with the full precision - and are sorted by keys for now, we can think about how to handle that in the future (also depends on how we want to sort the output kets in the future).

The following reference files are not yet updated:
- 4C_input_solid_cuboid.4C.yaml
- test_cosserat_curve_mesh.4C.yaml
- test_create_cubit_input_block.4C.yaml
- test_create_cubit_input_solid_shell_blocks.4C.yaml
- test_create_cubit_input_solid_shell_dome.4C.yaml
- test_create_cubit_input_tube.4C.yaml
- test_four_c_beam_to_solid.4C.yaml
- test_four_c_import_non_consecutive_geometry_sets_input.4C.yaml
- test_four_c_simulation_dbc_monitor_to_input_restart.4C.yaml
- test_four_c_simulation_dirichlet_boundary_to_neumann_boundary_with_all_values_neumann.4C.yaml
- test_four_c_simulation_honeycomb_sphere_import.4C.yaml
- test_meshpy_comments_in_solid_initial.4C.yaml
- test_meshpy_fluid_element_section_import.4C.yaml
- test_meshpy_nurbs_import_solid_mesh.4C.yaml
- test_meshpy_surface_to_surface_contact_import_solid_mesh.4C.yaml
- test_meshpy_vtk_writer_solid_elements_import.4C.yaml
- test_tutorial.4C.yaml


tests/reference-files/test_create_cubit_input_block.4C.yaml
```diff
@@ -1,13 +1,16 @@
...
+  - COUPLING_ID: 1
+    E: 1
 DESIGN POINT COUPLING CONDITIONS:
   - E: 1
     NUMDOF: 3
-    ONOFF: [1, 2, 3]
+    ONOFF:
+      - 1
+      - 2
+      - 3
```
davidrudlstorfer, line 13: Thanks a lot for the nice change!

I appreciate it very much that we now no longer write out `0` for the coords, triads, etc... but `0.0`

It took me some time to figure out why we now have 17k more lines - this can be attributed to this change. Is it anticipated to now output a list in this way? I actually liked the old output with the list in one line a lot.

With this change the functions change from

```yaml
FUNCT107:
  - SYMBOLIC_FUNCTION_OF_SPACE_TIME: "var"
  - VARIABLE: 0
    NAME: "var"
    TYPE: "linearinterpolation"
    NUMPOINTS: 6
    TIMES: [-1000, 0, 0.3333333333333333, 0.6666666666666666, 1, 1001]
    VALUES: [-8.881784197001252e-16, -8.881784197001252e-16, 0.21795397931885296, 0.8772042896298684, 1.5674410542562862, 1.5674410542562862]
```

to 

```yaml
FUNCT107:
  - SYMBOLIC_FUNCTION_OF_SPACE_TIME: var
  - NAME: var
    NUMPOINTS: 6
    TIMES:
      - -1000
      - 0
      - 0.3333333333333333
      - 0.6666666666666666
      - 1
      - 1001
    TYPE: linearinterpolation
    VALUES:
      - -8.881784197001252e-16
      - -8.881784197001252e-16
      - 0.21795397931885296
      - 0.8772042896298684
      - 1.5674410542562862
      - 1.5674410542562862
    VARIABLE: 0
```

So what's the reason behind this change?

isteinbrecher: These changes are due to reading, sorting, and then re-writing the YAML - apparently with a different formatting style applied during the write step.

I agree, I also prefer the old formatting. That said, I’d suggest we leave it as is for now and regenerate all reference files once we’ve integrated `fourcipp` (that should also close #308).

davidrudlstorfer: Sounds good 👌

davidrudlstorfer (approved):
Thanks for tackling this  - and one final question - what's up with the not converted files?

Maybe let's add the two ToDo's to #308 

- Files which are not already stored in "full precision"
- Inline lists (rather than expanded across multiple lines)

isteinbrecher:
The not converted files are because those test were not run on my machine (CubitPy and 4C). Once CubitPy is working again, I will convert the rest of them.

PR #331, opened 2025-04-29
Allow to control twist along a Cosserat curve

### LINKED ISSUE #328: Improve interface of Cosserat Curve
_There is no good way of prescribing the twist angle along a cosserat curve._ @isteinbrecher 

Within the function `get_quaternions_along_curve` (Line 106) the rotation of the cosserat curve is calculated.
Depending on this rotation, the initial curve may experience a twist rotation along it's axis.

isteinbrecher:
Aș described in #328 it is currently not possible to change / modify the twist along a Cosserat curve, although from a geometrical point of view, this twist is arbitrary. This causes issues for some warping simulations. This PR adds the functionality to define the twist, wither by providing an initial guess triad to the constructor, or by providing the twist angle manually.

Closes #328


isteinbrecher:
@knarfnitram can you please try if this allows to solve your problem with the warping?

davidrudlstorfer (approved):
Thanks for also updating the docstrings @isteinbrecher :)

One question out of curiosity - this twists the Cosserat curve along its axis? How does this influence the simulation if we use axisymmetric cross sections?

isteinbrecher:
@davidrudlstorfer the `CosseratCurve` in MeshPy is not directly related to any beam theory, in the current form, it is simply a geometrical tool that can perform some functionality like unwrapping for smooth transformations. So a `CosseratCurve` does not have a cross-section.

If it were a beam with an axisymmetric cross section (statics, no torsional loads or BCs), then the twist angle is a zero energy mode of the beam and has no unique solution (but there is a unique solution for all other fields along the beam).

knarfnitram:
@isteinbrecher thanks for implementing the rotation. I have taken a look at the changes and they look good to me. 
I tried to test it, but I failed due to mismatch of cubitpy and meshpy. My plan is to update the test case at the end of the week, but until then feel free to merge this.

PR #334, opened 2025-05-05
Update latest supported python release to 3.13

isteinbrecher:
Set the latest supported python version to 3.13.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os-version: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10", "3.12"]
+        python-version: ["3.10", "3.13"]
     runs-on: ${{ matrix.os-version }}
     steps:
       - name: Checkout repository
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: '3.13'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/README.md
+++ b/README.md
@@ -167,11 +167,11 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
 
 ### Python environment
 
-MeshPy is tested with, and supports Python versions 3.9-3.12. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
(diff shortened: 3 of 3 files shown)


isteinbrecher:
Pipelines passed. The only weird thing is that the all of the python3.13 pipelines took significantly longer than the python3.10 ones.

isteinbrecher:
There are also some deprecation warnings. I am marking this as draft for now.

isteinbrecher:
Good, then let't see that we can merge this soon! I had a look and the newly introduced warnings are the following (non are related to actual MeshPy code):
- [ ] Should be fixed in pyvista utils 
  ```
    /opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/vtk_utils/compare_grids.py:126: DeprecationWarning: Call to deprecated method GetFaces. (Use GetPolyhedronFaces instead.) -- Deprecated since version 9.4.0.
      grid_1.GetFaces(), grid_2.GetFaces(), "face_connectivity"
  ```
- [ ] If possible, should be fixed in splinepy
  ```
  tests/test_mesh_creation_functions.py::test_mesh_creation_functions_nurbs_unit
    /opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/splinepy/spline.py:731: FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior
      self._logd(
  ```
  ```
  tests/test_mesh_creation_functions.py::test_mesh_creation_functions_nurbs_unit
    /opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/splinepy/spline.py:1149: FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior
      self._logd("Evaluating spline")
  ```
  ```
  tests/test_mesh_creation_functions.py::test_mesh_creation_functions_nurbs
  tests/test_mesh_creation_functions.py::test_mesh_creation_functions_nurbs_unit
    /opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/splinepy/spline.py:1202: FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior
      self._logd("Evaluating derivatives of the spline")
  ```

I will take care of the first one and will talk to @clemens-fricke (for full reference also tagging @j042) about the relevance of the others.

The only reason why I would wait for now is that splinepy only provides pre-build wheels for python<3.12, so pip is building the underlying C++ code in each testing job, which causes the jobs to take between 8 and 20 minutes, compared to the 1,5-3 minutes for older python versions (I am actually amazed that this works, but would like to keep the low testing time if possible). I have already contacted to the splinepy developers and they will see if they can publish pre-build wheels for python3.13.

clemens-fricke:
Ahh yeah, that was the reason why we postponed the integration of the 3.13 version. If I have time, I will look into it...

PR #338, opened 2025-05-05
Add converters of MeshPy rotation with numpy quaternion

isteinbrecher:
Add controlled converters between the MeshPy quaternion implementation and the numpy quaternion package.

In the future we can think if we want to completely get rid of the quaternion dependency or make it the core data structure of the `Rotation` class.

### DIFF
--- a/src/meshpy/core/rotation.py
+++ b/src/meshpy/core/rotation.py
@@ -24,6 +24,7 @@
 import copy as _copy
 
 import numpy as _np
+import quaternion as _quaternion
 
 from meshpy.core.conf import mpy as _mpy
 
@@ -88,6 +89,9 @@ class Rotation:
             where we can be sure that the input quaternion is normalized to avoid
             error accumulation.
         """
+        if isinstance(q, _quaternion.quaternion):
+            q = _quaternion.as_float_array(q)
+
         rotation = object.__new__(cls)
         if normalized:
             rotation.q = _np.array(q)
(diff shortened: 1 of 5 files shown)


src/meshpy/core/rotation.py
```diff
@@ -195,9 +199,13 @@ def get_rotation_matrix(self):
 
     def get_quaternion(self):
         """Return the quaternion for this rotation, as numpy array (copy)."""
-
         return _np.array(self.q)
 
+    def get_numpy_quaternion(self):
+        """Return a numpy quaternion object representing this rotation
+        (copy)."""
+        return _quaternion.from_float_array(_np.array(self.q))
```
davidrudlstorfer, line 207: Sorry for the dumb question - do we need this double conversion here?
We convert the quaternion into an np array and then create a quaternion from the array? Can we directly return the quaternion?

```suggestion
        return self.q
```

davidrudlstorfer (approved):
Nice! Just one question because I do not understand it

src/meshpy/core/rotation.py
isteinbrecher: No and no 😉

`_quaternion.from_float_array` returns a numpy array with the datatype `quaternion.quaternion`, so we can't just return `self.q`. Furthermore, `_quaternion.from_float_array` will return a view if given a numpy array, so we have to add the `_np.array` copy conversion.

In theory, we should be fine if we return a view here since at the moment we never modify the return value from this function. I think returning a copy here is the safe way. If we need a view, we can always implement it, but from my experience, dealing with numpy views in such cases can lead to some nasty bugs.

Also, we should maybe add type hints to `rotation.py` at some point since this is one of the modules that every user interacts with.

davidrudlstorfer: Thanks for the explanation 🙏🏻😊 Unfortunately, I never understand these things from the start 😅

davidrudlstorfer (approved):
Let's get this merged 🚀

src/meshpy/core/rotation.py
isteinbrecher: I don't blame you, referencing in python is a mess...

PR #330, opened 2025-04-28
Move input file & testing to fourcipp

davidrudlstorfer:
This evening I finally found some time to really tackle this larger move (following the initial discussion and plan in #321) to fourcipp

The present changes are a very rough WIP and are not ready yet (I hope that this is clear :D)

The first draft now proposes the completely overhauled FouCInputFile structure which is based on fourcipp.

Some points:

- **The basic functionality works as expected** and the code is simplified by a lot
- I now provided a general utility function which converts a MeshPy mesh into a dictionary (without depending on 4C stuff) => we can reuse this in testing to compare two Meshes

The missing points

- [x] Finalize coupling, boundary conditions, geometry sets, Nurbs to mesh->dict conversion
- [x] Finalize initial conversion from an input file to a mesh for boundary conditions and geometry sets
- [x] remove last `dump_to_list` remnants **(except for Nurbs -> I have no overview on how to approach this the best way)**
- [x] remove last `_if_fourcipp_is_available` occurences **(except for Nurbs -> I have no overview on how to approach this the best way)**
- [ ] use fourcipp dict comparison in testing
- [ ] update tests

PR #341, opened 2025-05-07
Remove mesh from input file and rework input file read in mechanism

davidrudlstorfer:
This PR now removes the mesh entirely from the input file.

Additionally this PR introduces our new API to read in an existing input file as discussed. In my opinion it makes sense that this function is not a free function - but we overwrite the existing function from fourcipp (hopefully this is soon included).

With this change the next step would be to introduce fourcipp into MeshPy. This should be rather straight forward.

### DIFF
--- a/src/meshpy/four_c/input_file.py
+++ b/src/meshpy/four_c/input_file.py
@@ -33,63 +33,28 @@ from typing import Dict as _Dict
 from typing import List as _List
 from typing import Optional as _Optional
 from typing import Tuple as _Tuple
-from typing import TypeVar as _TypeVar
-from typing import Union as _Union
 
 import yaml as _yaml
 
-import meshpy.core.conf as _conf
 from meshpy.core.boundary_condition import BoundaryCondition as _BoundaryCondition
-from meshpy.core.boundary_condition import (
-    BoundaryConditionBase as _BoundaryConditionBase,
-)
 from meshpy.core.conf import mpy as _mpy
 from meshpy.core.container import ContainerBase as _ContainerBase
-from meshpy.core.coupling import Coupling as _Coupling
-from meshpy.core.element import Element as _Element
 from meshpy.core.geometry_set import GeometryName as _GeometryName
-from meshpy.core.geometry_set import GeometrySetNodes as _GeometrySetNodes
 from meshpy.core.mesh import Mesh as _Mesh
-from meshpy.core.node import Node as _Node
 from meshpy.core.nurbs_patch import NURBSPatch as _NURBSPatch
+from meshpy.four_c.input_file_mappings import (
+    boundary_condition_names as _boundary_condition_names,
+)
+from meshpy.four_c.input_file_mappings import geometry_set_names as _geometry_set_names
 from meshpy.four_c.yaml_dumper import MeshPyDumper as _MeshPyDumper
 from meshpy.utils.environment import cubitpy_is_available as _cubitpy_is_available
 from meshpy.utils.environment import fourcipp_is_available as _fourcipp_is_available
-
-# necessary to allow for type hint of from_4C_yaml for python version <3.11
-# can be replaced with _Self in python 3.11
-T = _TypeVar("T", bound="InputFile")
 
 if _cubitpy_is_available():
     import cubitpy as _cubitpy
 
 
-def _boundary_condition_from_dict(
-    geometry_set: _GeometrySetNodes,
-    bc_key: _Union[_conf.BoundaryCondition, str],
-    data: _Dict,
-) -> _BoundaryConditionBase:
-    """This function acts as a factory and creates the correct boundary
-    condition object from a dictionary parsed from an input file."""
-
-    del data["E"]
-
-    if bc_key in (
-        _mpy.bc.dirichlet,
-        _mpy.bc.neumann,
-        _mpy.bc.locsys,
-        _mpy.bc.beam_to_solid_surface_meshtying,
-        _mpy.bc.beam_to_solid_surface_contact,
-        _mpy.bc.beam_to_solid_volume_meshtying,
-    ) or isinstance(bc_key, str):
-        return _BoundaryCondition(geometry_set, data, bc_type=bc_key)
-    elif bc_key is _mpy.bc.point_coupling:
-        return _Coupling(geometry_set, bc_key, data, check_overlapping_nodes=False)
-    else:
-        raise ValueError("Got unexpected boundary condition!")
-
-
-def _get_geometry_set_indices_from_section(
+def get_geometry_set_indices_from_section(
     section_list: _List, *, append_node_ids: bool = True
 ) -> _Dict:
     """Return a dictionary with the geometry set ID as keys and the node IDs as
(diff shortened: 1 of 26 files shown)


davidrudlstorfer:
Nice that we check older python versions as well - turns out `Self` is only available from 3.11 onwards

isteinbrecher (commented):
Thanks a lot for the changes @davidrudlstorfer . I think we can merge this soon, I left some small comments.

The only thing I would like to discuss is regarding the interface and name of `InputFile.from_4C_yaml`: This function returns an input file and a mesh, thus I am not 100% happy that this is inside of InputFile, I see this more as a free function. Also, there is one aspect we (and especially me) didn't think about yet, in 99% of the cases we don't read an existing input file but we convert a CubitPy object to a MeshPy mesh and input file. Thus I am also not sure we want to name this `from_4C_yaml`.

Maybe we can have something like (naming is just a suggestion):
```python
def load_existing_model(yaml_path=None, cubit=None, full_import=False):
    """yaml_path and cubit are mutually exclusive"""
    
    if yaml_path:
        input_file = InputFile.from_4C_yaml(yaml_path)
    else:
        input_file = InputFile()
        input_file.add(cubit.get_dict())
    
    if full_import:
        mesh = input_file.pop_mesh() # This also deletes the mesh sections in input_file
    else:
        mesh = Mesh()

    return input_file, mesh
```

I am curious about your opinion.

src/meshpy/four_c/input_file.py
```diff
@@ -329,7 +338,7 @@ def _get_section_items(section_name):
                     _boundary_condition_from_dict(geometry_set, bc_key, item),
                 )
 
-        self.add(mesh)
+        return mesh
 
     def add(self, *args, **kwargs):
```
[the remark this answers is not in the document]
davidrudlstorfer: This makes total sense and it's more or less exactly what I've done in all the tests.

Unfortunately, I do not really understand what "TODO" I should add? To rework this function? This will be done once fourcipp is introduced in the next step

src/meshpy/four_c/input_file.py
```diff
@@ -442,8 +452,9 @@ def write_input_file(
                 )
                 input_file.writelines(application_script_lines)
 
-    def get_dict_to_dump(
+    def add_mesh_to_dict(
         self,
+        mesh: _Mesh,
         *,
         add_header_information: bool = True,
         check_nox: bool = True,
```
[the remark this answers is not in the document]
davidrudlstorfer: Your're 100% right - I actually had the changes prepared to add these only during writing the file. To keep the changes smaller within this PR I reverted them. I will add it back in

tests/test_meshpy.py
```diff
@@ -1170,73 +1167,74 @@ def test_meshpy_replace_nodes_geometry_set(
...
 
-    # Add the beam to the solid mesh.
-    input_file.mesh.add(beam_mesh)
-
-    return input_file
+    return input_file, mesh
 
 
-@pytest.mark.parametrize("test_type", [None, "full"])
+# TODO change additional identifier
```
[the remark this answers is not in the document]
davidrudlstorfer: I know that I need to improve my ToDo's :D

Across the tests we have like 5 different approaches on how to run these tests where we do a "full" import or only an input file import. These approaches should be standardized

Once we use 

```
@pytest.mark.parametrize(
    ("full_import", "additional_identifier"),
    [(False, "dict_import"), (True, "full_import")],
)
```

another time

```
@pytest.mark.parametrize("full_import", [False, True])
```

as well as 

```
@pytest.mark.parametrize(
    ("full_import", "additional_identifier"),
    [(False, None), (True, "full")],
)
```

Maybe it makes sense to provide a fixture for that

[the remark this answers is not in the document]
davidrudlstorfer: And similarly - the reference files are also named differently for each case. This was very confusing

src/meshpy/four_c/input_file.py
```diff
@@ -216,30 +215,40 @@ def __init__(self, *, yaml_file: _Optional[_Path] = None, cubit=None):
...
-                A file path to an existing input file that will be read into this
-                object.
+            input_file_path: A file path to an existing input file that will be read
+                into this object.
+            convert_input_to_mesh: If True, the input file will be converted to a
+                MeshPy mesh.
+
+        Returns:
+            A tuple with the input file and the mesh. If convert_input_to_mesh is
+            False, the mesh will be empty.
```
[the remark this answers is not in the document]
davidrudlstorfer: Done

src/meshpy/four_c/input_file.py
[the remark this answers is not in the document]
davidrudlstorfer: Done

tests/test_meshpy.py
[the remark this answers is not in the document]
davidrudlstorfer: Added a better comment

davidrudlstorfer:
@isteinbrecher thanks for your review - I've added a few commits to address your comments

1. I also was very divided if this function makes sense as a class function of input file. My motive for putting it into the InputFile was that we overwrite the fourcipp function in the future. But after giving it some thought the original base function still works and this makes more sense.

I've now created a new file called `model_importer.py` which contains the new `import_4C_model` function. This way the future refactoring is more contained and it also is not directly connected to the 4C input file. You also noted that most of the times a cubit input is loaded. In my opinion it would make sense to also provide a `import_cubit_model` function. This way the functionality is better separated as these two do not have anything in common. But this can be discussed in the future.

I've now also extracted the input file mappings into a separate file as proposed and already discussed in #330

2. I've reworked and already simplified the add function

3. I've again extracted the add_header_information flag to the input file writing mechanism which makes a lot more sense.

4. I've improved the comment regarding the additional identifiers in testing

This would be ready for review again and I am curious about your opinion now :D

davidrudlstorfer:
I now also fixed the 3 4C tests which were failing. This was really random because originally something wrong was compared. I needed to adjust the reference file with a definitely missing key. Otherwise the test (correctly) failed.

So prior to this the tests were not working at all.

src/meshpy/four_c/input_file.py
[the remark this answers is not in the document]
isteinbrecher: Thanks, that is exactly what I was referring to.

tests/test_meshpy.py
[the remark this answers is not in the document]
isteinbrecher: Yes that totally makes sense, thanks for adding the TODOs (and sorry for introducing so many different ways to do this parametrise).

src/meshpy/four_c/model_importer.py
```diff
@@ -0,0 +1,221 @@
...
+    return geometry_sets_in_this_section
+
+
+def sections_to_mesh(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
+    """Convert an existing input file to a MeshPy mesh with mesh items, e.g.,
+    nodes, elements, element sets, node sets, boundary conditions, materials.
+
+    Note: In the current implementation we cannibalize the mesh sections in
+        the self.sections dictionary. This should be reconsidered and be done
+        in a better way when this function is generalized.
```
isteinbrecher, line 132: This comment is outdated and can be removed (since we can now copy the `input_file` before passing to this function, we can already control this behaviour).

src/meshpy/four_c/model_importer.py
```diff
@@ -0,0 +1,221 @@
...
+        raise ValueError("Got unexpected boundary condition!")
+
+
+def get_yaml_geometry_sets(
+    nodes: _List[_Node], geometry_key: _conf.Geometry, section_list: _List
+) -> _Dict[int, _GeometrySetNodes]:
+    """Add sets of points, lines, surfaces or volumes to the object."""
+
+    # Create the individual geometry sets. The nodes are still integers at this
+    # point. They have to be converted to links to the actual nodes later on.
```
isteinbrecher, line 116: This comment is outdated and can be removed.

src/meshpy/four_c/model_importer.py
```diff
@@ -0,0 +1,221 @@
...
+    get_geometry_set_indices_from_section as _get_geometry_set_indices_from_section,
+)
+from meshpy.four_c.input_file_mappings import (
+    boundary_condition_names as _boundary_condition_names,
+)
+from meshpy.four_c.input_file_mappings import geometry_set_names as _geometry_set_names
+from meshpy.utils.environment import fourcipp_is_available as _fourcipp_is_available
+
+
+def import_4C_model(
```
isteinbrecher, line 54: I like this, also having this in this file. What about the name? Currently we name python functionally related to 4C like: `run_four_c`. Should we name this `import_four_c_model` to be consistent? Instinctively, I don't like the `4C` in the name (numerals and upper case).

I understand your point about having two functions for CubitPy and YAML input and thinking about it, let's do it in your proposed way. We can always merge the functionality if we feel we need to.

src/meshpy/four_c/model_importer.py
```diff
@@ -0,0 +1,221 @@
...
+        input_file = _InputFile()
+        input_file.sections = _yaml.safe_load(stream)
+
+    if convert_input_to_mesh:
+        return sections_to_mesh(input_file)
+    else:
+        return input_file, _Mesh()
+
+
+def boundary_condition_from_dict(
```
isteinbrecher, line 85: Should we mark this (and the following) function private?

src/meshpy/four_c/model_importer.py
```diff
@@ -0,0 +1,221 @@
...
+    geometry_set_dict = _get_geometry_set_indices_from_section(section_list)
+    geometry_sets_in_this_section = {}
+    for geometry_set_id, node_ids in geometry_set_dict.items():
+        geometry_sets_in_this_section[geometry_set_id] = _GeometrySetNodes(
+            geometry_key, nodes=[nodes[node_id] for node_id in node_ids]
+        )
+    return geometry_sets_in_this_section
+
+
+def sections_to_mesh(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
```
isteinbrecher, line 126: Suggestion: `pop_mesh_sections_to_mesh` or `extract_mesh_sections`, since his is a quite impactful function, I would like to indicate this in the name - but maybe I am overthinking this.

src/meshpy/four_c/input_file_mappings.py
isteinbrecher, line 1: Nice to have this in a file, this improves the readability from `InputFile` a lot!

src/meshpy/four_c/input_file.py
```diff
@@ -60,32 +52,7 @@
...
-    ) or isinstance(bc_key, str):
-        return _BoundaryCondition(geometry_set, data, bc_type=bc_key)
-    elif bc_key is _mpy.bc.point_coupling:
-        return _Coupling(geometry_set, bc_key, data, check_overlapping_nodes=False)
-    else:
-        raise ValueError("Got unexpected boundary condition!")
-
-
-def _get_geometry_set_indices_from_section(
+def get_geometry_set_indices_from_section(
```
isteinbrecher, line 55: This function seems a little out of place there. Suggestion: we could move this to `input_file_utils.py`.

src/meshpy/four_c/input_file.py
```diff
@@ -384,41 +152,51 @@ def write_input_file(
         self,
         file_path: _Path,
         *,
-        nox_xml_file: _Optional[str] = None,
+        nox_xml_file: bool | str = False,
```
isteinbrecher, line 155: Nice that this type hint works - should we use `Type | None` in favour of `_Optional` in the future?

src/meshpy/four_c/input_file.py
```diff
@@ -384,41 +152,51 @@ def write_input_file(
...
-                    _os.path.splitext(_os.path.basename(file_path))[0] + ".xml"
-                )
-            else:
-                self._nox_xml_file = nox_xml_file
+        if self.nox_xml_contents:
+            if nox_xml_file is False:
+                xml_file_name = "NOT_DEFINED"
+            elif nox_xml_file is True:
+                xml_file_name = _os.path.splitext(file_path)[0] + ".xml"
+            elif isinstance(nox_xml_file, str):
```
isteinbrecher, line 186: Does an final `else` which throws an error make sense?

src/meshpy/four_c/input_file.py
```diff
@@ -657,32 +403,30 @@ def get_number_of_coupling_conditions(key):
...
-                _dump_mesh_items(yaml_dict, self.geometry_set_names[geom_key], item)
+                _dump_mesh_items(self.sections, _geometry_set_names[geom_key], item)
 
         # Add the nodes and elements.
-        _dump_mesh_items(yaml_dict, "NODE COORDS", self.mesh.nodes)
-        _dump_mesh_items(yaml_dict, "STRUCTURE ELEMENTS", self.mesh.elements)
-
-        return yaml_dict
+        _dump_mesh_items(self.sections, "NODE COORDS", mesh.nodes)
+        _dump_mesh_items(self.sections, "STRUCTURE ELEMENTS", mesh.elements)
```
isteinbrecher, line 429: Could you add one TODO here:
```python
# TODO: reset all links and counters set in this method.
```

isteinbrecher (commented):
Thanks for the changes @davidrudlstorfer , I really like the way this looks now and think this is a good way to go to in the future!

> So prior to this the tests were not working at all.

Thanks a lot for fixing the tests! Do you know why the issue was not detected previously and the pipeline passed?

src/meshpy/four_c/model_importer.py
davidrudlstorfer: Removed it

src/meshpy/four_c/model_importer.py
davidrudlstorfer: Removed

src/meshpy/four_c/model_importer.py
davidrudlstorfer: Good catch! Makes sense, renamed it to `import_four_c_model`

src/meshpy/four_c/model_importer.py
davidrudlstorfer: Good idea - changed that

src/meshpy/four_c/model_importer.py
davidrudlstorfer: I've now renamed this to `_extract_mesh_sections`

I'll add an additional comment regarding all of this at the bottom

src/meshpy/four_c/input_file.py
davidrudlstorfer: I think we should stay with the Optional as it is a bit clearer in my opinion.

Could we convert this option to something using Optional at all? Because Optional automatically implies that it can be None but in this case we have no None?

davidrudlstorfer: Oh that was wrong

We could use `str | None`

I still prefer the Optional

src/meshpy/four_c/input_file.py
davidrudlstorfer: Good idea. I've added a simple TypeError

src/meshpy/four_c/input_file.py
davidrudlstorfer: Added

davidrudlstorfer:
And regarding the failing tests - I did not take the time to look deeper into it why it did not work before. I suspect that somewhere the wrong mesh was utilized, and, therefore, the wrong simulation/data was used.

isteinbrecher (approved):
Thanks for going the extra mile and tackling all the points @davidrudlstorfer (I actually feel sorry about a few of them, because a lot of the points were originally my mistake...).

Let's this in the current state!

> With the upcoming change I want to simplify the mesh->dict conversion and the section->mesh conversion a lot.
> In a next PR I would clean up a few things across the changed files of this PR. I just want to split it up to keep the better overview.

Cleaning up some small things makes sense! Regarding the next "larger" PRs, I would suggest to introduce fourcipp as a dependency - with simply adapting the testing comparison and removing all legacy string. This should be a relatively nice to review PR. After that, let's have a look at the import and export functionality.

One nice aspect about splitting this up in a few PRs is that I think (or hope) that the upcoming changes will require no changes to the testing (except the comparison with fourcipp).

isteinbrecher:
After this is merged, I will try to have a look at all the remaining 4C stuff in `core`, so we know how to handle that before we rework the import and export stuff.

PR #342, opened 2025-05-09
Small input file cleanups

davidrudlstorfer:


davidrudlstorfer:
I think with this PR we are in a really good state to use fourcipp in a next step.

This PR now moves the `get_git_data` function out of the input file and adds unit tests as proposed in #330.
Additionally, the input file mappings are converted into a dict for easier import/access as proposed in #330. Into this dict we can simply add the name of the sections as well (i.e., material="MATERIALS", element="STRUCTURAL ELEMENTS", metadata_section="TITLE")

With this the 4C input file currently has the following structure

```python
class InputFile:

    def __init__(self):
    
    def add():
    
    def add_section():
    
    def delete_section():

    def write_input_file():
    
    def add_mesh_to_input_file():
    
    def _get_header():
    
    def _get_application_script():
```

With fourcipp we should be able to entirely delete `add_section` and `delete_section` as well as use the super methods for the `init` (with the extension of the nox content or some other solution for that) and `add` function (with our mesh addition). The `write_input_file` should probably be renamed to dump to overwrite the fourcipp function.

Once this is done the "output side" (`add_mesh_to_input_file`) as well as "input side" (`import_four_c_model`) can be reworked. But this is some sort of "bonus" for cleaner code and not actually necessary.

src/meshpy/four_c/input_file_mappings.py
```diff
@@ -24,65 +24,66 @@
...
-    ): "BEAM INTERACTION/BEAM TO BEAM CONTACT CONDITIONS",
-    (
-        _mpy.bc.point_coupling_penalty,
-        _mpy.geo.point,
-    ): "DESIGN POINT PENALTY COUPLING CONDITIONS",
-    (
-        "DESIGN SURF MORTAR CONTACT CONDITIONS 3D",
-        _mpy.geo.surface,
-    ): "DESIGN SURF MORTAR CONTACT CONDITIONS 3D",
+INPUT_FILE_MAPPINGS = {
```
isteinbrecher, line 27: Do we ever plan to use an alternative mapping that this one? I don't see a case where we actually want to provide an alternative mapping than this one. So I am not sure if it makes sense to add this possibility if we don't even know if we will ever use it. But my opinion on that is not super strong.

isteinbrecher (approved):
@davidrudlstorfer thanks for the changes, only one general comment but otherwise this looks good to me!

src/meshpy/four_c/input_file_mappings.py
davidrudlstorfer: My rough idea would be that we could provide a "general" mapping which does not depend on 4C. Naturally the general syntax would be very close to the 4C syntax. BUT we could convert a mesh into dictionary somehow independent of 4C (in a code related sense). If the "general" mappings are situated in core and the mesh to dict conversion as well, the modules and testing would be completely uncoupled.

Independent of this idea - what would you do different if we only ever use this one - rename it? Or put it in a different place?

isteinbrecher: I think there are simpler approaches to fully uncouple the general MeshPy testing from 4C, but we will see, for now this does not matter anyways.

> Independent of this idea - what would you do different if we only ever use this one - rename it? Or put it in a different place?

Good point, initially I thought, that we should just leave it as it is, but once `"ELEMENTS"`, `"NODES"` and so on are also in this dict (I assume that this is the intention) it makes a lot of sense to have them bundled together. So let's merge this! :rocket:

PR #343, opened 2025-05-09
Modularize coupling conditions

isteinbrecher:
This PR moves all the 4C related coupling functionality out of `core` and into `input_file`. The functionality of `_dump_coupling` is also refactored. The function `_dump_coupling` will likely be moved to a better place in the future - when the mesh export stuff is refactored.

### DIFF
--- a/src/meshpy/core/conf.py
+++ b/src/meshpy/core/conf.py
@@ -49,6 +49,21 @@ class BoundaryCondition(_Enum):
     point_coupling = _auto()
     point_coupling_penalty = _auto()
 
+    def is_point_coupling_pairwise(self) -> bool:
+        """Check whether the point coupling condition should be applied
+        pairwise.
+
+        Returns:
+            bool: True if the coupling should be applied individually between pairs of nodes,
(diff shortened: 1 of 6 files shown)


src/meshpy/core/conf.py
```diff
@@ -49,6 +49,21 @@ class BoundaryCondition(_Enum):
...
+        Returns:
+            bool: True if the coupling should be applied individually between pairs of nodes,
+                rather than to the entire geometry set as a whole.
+        """
+        if self == self.point_coupling:
+            return False
+        elif self == self.point_coupling_penalty:
+            return True
+        else:
+            raise TypeError(f"Gor unexpected coupling type {self}")
```
davidrudlstorfer, line 65:
```suggestion
            raise TypeError(f"Got unexpected coupling type: {self}")
```

davidrudlstorfer (approved):
Thanks for the cleanup!

src/meshpy/core/conf.py
isteinbrecher: Nice catch!

PR #344, opened 2025-05-09
Improve import of elements from yaml file

isteinbrecher:
This PR moves the import of (mainly) solid elements from a yaml file from `core.element` to `model_importer` which make a slot more sense from a modularisation point of view. The function is now named `_element_condition_from_dict` (at the moment it still converts a legacy string, but that will hopefully change soon). 

This also allows us to get rid of a lot of "in function" includes.

@davidrudlstorfer I hope this does not interfere with your existing work too much. If so, let's talk and we will find a solution.

### DIFF
--- a/src/meshpy/four_c/model_importer.py
+++ b/src/meshpy/four_c/model_importer.py
@@ -36,10 +36,16 @@ from meshpy.core.boundary_condition import (
 )
 from meshpy.core.conf import mpy as _mpy
 from meshpy.core.coupling import Coupling as _Coupling
-from meshpy.core.element import Element as _Element
+from meshpy.core.element_volume import VolumeHEX8 as _VolumeHEX8
+from meshpy.core.element_volume import VolumeHEX20 as _VolumeHEX20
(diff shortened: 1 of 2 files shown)


src/meshpy/four_c/model_importer.py
```diff
@@ -82,6 +89,62 @@ def import_four_c_model(
         return input_file, _Mesh()
 
 
+def _element_condition_from_dict(nodes: _List[_Node], input_line: str):
```
davidrudlstorfer, line 92: I do not really understand the naming here, why _element_condition_from_dict? Just `_element_from_dict`?

davidrudlstorfer (approved):
Looks good - thanks!

I am currently working on fourcipp so no overlaps here. Let's get this merged

src/meshpy/four_c/model_importer.py
isteinbrecher: Nice catch, this is simply a copy paste mistake.

PR #345, opened 2025-05-09
Add base solid material

isteinbrecher:
Add a base solid material, this allows to remove the `four_c` import in `nurbs_patch.py`.

### DIFF
--- a/src/meshpy/core/material.py
+++ b/src/meshpy/core/material.py
@@ -44,7 +44,7 @@ class Material(_BaseMeshItem):
         return self
 
 
-class MaterialBeam(Material):
+class MaterialBeamBase(Material):
     """Base class for all beam materials."""
 
     def __init__(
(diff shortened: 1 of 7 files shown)


src/meshpy/core/material.py
```diff
@@ -44,7 +44,7 @@ def __deepcopy__(self, memo):
         return self
 
 
-class MaterialBeam(Material):
+class MaterialBeamBase(Material):
```
isteinbrecher, line 47: @davidrudlstorfer what do you think about this, I am deciding between naming the classes in this file
- `MaterialBeamBase` <-> `MaterialBeam`
- `MaterialSolidBase` <-> `MaterialSolid`
I am not sure if we need/want the `Base` identifier. (Not an important point, just something that I thought about).

davidrudlstorfer (approved):

PR #349, opened 2025-05-14
Activate CubitPy support again

isteinbrecher:
This PR activates the CubitPy support again. Currently this is in _draft_ mode, as it requires #348 to be merged and https://github.com/imcs-compsim/cubitpy/pull/70 to be merged in CubitPy.

Currently this PR contains some hacks in the conversion that can be omitted once the previous mentioned PRs are merged.


isteinbrecher:
Nice to see that the CubitPy tests pass (all others fail due to some import issue - not relevant here). Performance tests look pretty bad right now, but I think that is because in the current state we use the default `yaml` library and not `fourcipp` - we will see, I am confident that this should resolve itself.

knarfnitram:
@isteinbrecher  Since #354 and https://github.com/imcs-compsim/cubitpy/pull/70 is merged, I rebased the branch. Feel free to cherry pick the commits from `martin-activates-cubitpy-support` from my repo or I can create a pull request and we close this. 

You mentioned the hacks/ conversions which you would like to get rid of. Can you elaborate a little bit, what you would like to change?

davidrudlstorfer (commented):
@isteinbrecher thanks a lot! Really nice that everything works (except for the speed of the performance tests)

Regarding the performance, currently we only use rapidyaml during dumping of the .dat file. The same approach with rapidyaml could also be applied for reading in .yaml files. 

But I haven't looked much deeper into the failing tests to determine what the problem could be.

davidrudlstorfer:
And one more thing.

We could mark the failing performance tests with XFAIL

https://docs.pytest.org/en/stable/how-to/skipping.html

This way they are executed but they do not have to suceed. This way they still run in each pipeline

src/meshpy/four_c/model_importer.py
```diff
@@ -51,6 +51,37 @@
...
 )
+from meshpy.utils.environment import cubitpy_is_available as _cubitpy_is_available
+
+if _cubitpy_is_available():
+    from cubitpy.cubit_to_fourc_input import (
+        get_input_file_with_mesh as _get_input_file_with_mesh,
+    )
+
+
+def import_cubitpy_model(cubit, convert_input_to_mesh: bool = False):
```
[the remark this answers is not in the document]
isteinbrecher: Me too, I had to create some examples last week, and I really like the current approach.

tests/test_performance.py
```diff
@@ -240,7 +257,8 @@ def test_performance(tmp_path):
         "meshpy_wrap_around_cylinder": 2.0,
         "meshpy_wrap_around_cylinder_without_check": 0.7,
         "meshpy_find_close_nodes": 0.5,
-        "meshpy_write_dat": 9.0,
+        "meshpy_add_mesh": 5,
```
davidrudlstorfer, line 260: Is this time from a state prior to moving to yaml or just arbitrary?

isteinbrecher: I think it is just adapted to the time it takes on my local machine. But this is a very good point, it would be nice to see how much that changed.

isteinbrecher:
@davidrudlstorfer I think this has something to do with how we write the file to disk. I am trying out a minimal example to see that is going on.

Regarding the failing tests, the performance tests are configured such that we can merge even with a failing job in GitHub. I would leave this for now. Having just looked at those tests, a refactoring surely makes sense. Then we can have a look if XFAIL is a better solution that the current one. But l would do that in a follow up PR, and after the performance bottleneck is fixed (which I would also do in a follow up PR).

davidrudlstorfer:
> @davidrudlstorfer I think this has something to do with how we write the file to disk. I am trying out a minimal example to see what is going on.
> 
> Regarding the failing tests, the performance tests are configured such that we can merge even with a failing job in GitHub. I would leave this for now. Having just looked at those tests, a refactoring surely makes sense. Then we can have a look if XFAIL is a better solution that the current one. But l would do that in a follow up PR, and after the performance bottleneck is fixed (which I would also do in a follow up PR).

I also just had a look and i had it wrong in my mind.

We are currently not using rapidyaml to dump the file (we only use it to read it in). According to the official rapidyaml docs we should gain a speedup of around 10-25x faster than with the standard yaml library we currently use.

I'll have a look on the fourcipp side and open a PR (hopefully today). Therefore, let's first fix this and then you can have a deeper look into it (you can save your time for now)

knarfnitram (approved):

PR #355, opened 2025-05-20
Improve handling of `data` in BaseMeshItem

isteinbrecher:
We have the member variable `self.data` in `BaseMeshItem`, which is very well suited to hold solver specific data. This PR fixes some weird initialisations of this variable in `Node`, `Material` and `Element`.

@davidrudlstorfer at the moment the default value of `BaseMeshItem.data` is `None`. I think that is the most natural choice, but if it helps us for 4C, we can also change this to an empty dict.

### DIFF
--- a/src/meshpy/core/element.py
+++ b/src/meshpy/core/element.py
@@ -28,7 +28,7 @@ class Element(_BaseMeshItem):
     """A base class for an FEM element in the mesh."""
 
     def __init__(self, nodes=None, material=None, **kwargs):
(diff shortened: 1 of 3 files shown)


davidrudlstorfer (approved):
Thanks for the change

I think an empty dict could make sense - have you checked if something would break if we change it to an empty dict?

isteinbrecher:
I am pretty sure that all tests should pass with `{}` as default.

PR #357, opened 2025-05-28
Fix docu pipeline following #354

davidrudlstorfer:


isteinbrecher (commented):
Thanks for fixing this!

One thought: Should we maybe make fourcipp also a dependency of the `dev` dependencies instead of requiring `[dev,fourc]` to „develop“ MeshPy. Very few tests will pass if fourcipp is not available, even if one does not want to develop for 4C.

davidrudlstorfer:
So this took more time than anticipated. Unfortunately, my 1st proposed way of solving this did not actually work.

When we create the API docu with pdoc, pdoc actually imports all files to get more information on the type hints etc.

Because our InputFile inherits from the FourCInput from FourCIPP we need FourCIPP to create the documentation.

Regarding your proposal to merge the fourc dependency with the dev dependency: I would keep them separated for now. I have an idea to enable the tests without the FourCIPP dependency. Then we would really only need it for writing 4C files.

Therefore, this would be ready (and the docu creation works again) :)

isteinbrecher:
Thanks for the explanation. My proposal was actually not to merge `[dev]` and `[fourc]`, but to additionally add `fourcipp` to the `[dev]` optional dependencies, then one would simply have to install MeshPy with `pip install -e .[dev]` if one wants to develop it.

davidrudlstorfer:
> Thanks for the explanation. My proposal was actually not to merge [dev] and [fourc], but to additionally add fourcipp to the [dev] optional dependencies, then one would simply have to install MeshPy with pip install -e .[dev] if one wants to develop it.

I would still keep it as it is right now. The command on how to install the development environment is in the readme. I would not add the FourCIPP dependency twice into the pyproject.

isteinbrecher (approved):
Sounds good to me! Lets get this merged!

Issue #356, opened 2025-05-28
General improvements and fixes following the switch to .yaml based input


davidrudlstorfer:
From #354 and #135 (and other issues) some open ToDo's emerged which are summarized in the following:

### High Priority

- [x] Readd cubitpy support #349 

- [x] Fix documentation pipeline #357 

- [x] Fix testsuite (somehow the last nightly pipeline failed https://github.com/imcs-compsim/meshpy/actions/runs/15291400842/job/43051162413 I have no clue why because it pulls an old version of MeshPy. Maybe some cache is invalid. Let's see if it works tomorrow) => solved itself

- [x] Fix xml writeout #360 

### Testing infrastructure

- [x] improve testing infrastructure (test abaqus input files also with compare_nested_list_or_dicts, remove entire string comparison from our side, also open a diff in VSCode for the yaml files) #378 #379 #381

- [x] Check & increase minimum testing coverage #362 

### Code quality

- [x] Replace `self.element_description` with `self.data`. #372 

- [x] Pass the connectivity array of elements into the input file and hand over the conversion for a MeshPy Node to the FourCIPP converter #370 

- [x] move all dump_to_list logic to the input file (similar to boundary_condition, function, etc…)

- [x] move element_type_to_four_c_string to a more central place, i.e., out of core. This should be easily possible once the dump_to_list functions are in the input file (currently a circular import restricts me from moving this)
- [x] then we can also fix the current hackery in the solidrigidsphere for free (here also a circular import hinders me from working further on this)

- [x] Finally we can improve and simplify (by a lot) the model importer #473 

- [x] And finally, finally we can improve and simplify the mesh to dict conversion (by a lot)

### Miscellaneous

- [x] Improve performance again #358, #363, #364, https://github.com/4C-multiphysics/fourcipp/pull/59
- [x] discuss if the validation should be turned on during dump (note: this could lead to failing tests, i.e., some form of input changes in 4C - we pull the latest FourCIPP version nightly - then the validation would fail, note that this could also be a good thing) #361 

- [x] Update fourcipp to be pulled from pip once it is published to pip #448 

### Reference files

- [x] Convert remaining input files to full precision (from #309) #411 

    <details>
    <summary>List of files</summary>

    - [ ] ~4C_input_solid_cuboid.4C.yaml~
    - [ ] ~test_create_cubit_input_block.4C.yaml~
    - [ ] ~test_create_cubit_input_solid_shell_blocks.4C.yaml~
[...]

Issue #358, opened 2025-05-30
Compare performance dat vs yaml


isteinbrecher:
The switch to yaml seems to have introduced a performance issue when writing files to disk (see https://github.com/4C-multiphysics/fourcipp/issues/58). @davidrudlstorfer I didn't see your comment https://github.com/imcs-compsim/meshpy/pull/349#issuecomment-2921457691 until now😅.

To allow for quick comparisons, I have pushed a branch https://github.com/isteinbrecher/meshpy/tree/final_dat_version where MeshPy is still working with full dat support (some links to repos had to be updated).

At the end of this description is a script that shows the time needed to write a beam filament with a given number of elements to disk (working for the current `main` and the `final_dat_version` branch). The following table shows the time comparisons:

| `n_el`      | YAML Time (s) | DAT Time (s) | YAML / DAT Ratio |
|-------------|---------------|--------------|-------------------|
| 1,000       | 0.1549        | 0.0636       | 2.44              |
| 10,000      | 1.2446        | 0.3631       | 3.43              |
| 100,000     | 12.5261       | 3.2238       | 3.89              |
| 1,000,000   | 129.1956      | 31.7636      | 4.07              |

```python
import time

from meshpy.core.mesh import Mesh
from meshpy.four_c.element_beam import Beam3rHerm2Line3
from meshpy.four_c.input_file import InputFile
from meshpy.four_c.material import MaterialReissner

try:
    from meshpy.mesh_creation_functions.beam_basic_geometry import create_beam_mesh_line

    is_dat = True
except Exception:
    from meshpy.mesh_creation_functions.beam_line import create_beam_mesh_line

    is_dat = False


def check_dump_time(n_el):
    """Create a beam model and print the dump time to the terminal."""
    mat = MaterialReissner()
    mesh = Mesh()
    create_beam_mesh_line(
        mesh, Beam3rHerm2Line3, mat, [0, 0, 0], [0, 0, 100], n_el=n_el
    )
    input_file = InputFile()
    input_file.add(mesh)
    start = time.time()
    if is_dat:
        input_file.write_input_file("test.dat")
    else:
        input_file.dump("test.yaml")
    end = time.time()
    print(f"elapsed time for write: {end - start} (n_el: {n_el})")


check_dump_time(1000)
check_dump_time(10000)
check_dump_time(100000)
check_dump_time(1000000)
```

davidrudlstorfer:
[...]

PR #361, opened 2025-06-02
Change default to enable input file validation by FourCIPP

davidrudlstorfer:
Part of #356

### DIFF
--- a/tests/create_cubit_input.py
+++ b/tests/create_cubit_input.py
@@ -318,7 +318,9 @@ def create_solid_shell_meshes(file_path_blocks, file_path_dome):
     brick.translate([3 * 4, 0, 0])
     input_file.add(brick)
 
-    input_file.dump(file_path_blocks, add_header_information=False)
+    input_file.dump(
+        file_path_blocks, add_header_information=False, validate_sections_only=True
+    )
 
     # Create the dome input
     cubit = CubitPy()
--- a/src/meshpy/four_c/input_file.py
+++ b/src/meshpy/four_c/input_file.py
@@ -149,7 +149,8 @@ class InputFile(_FourCInput):
         add_header_information: bool = True,
         add_footer_application_script: bool = True,
         sort_sections=False,
-        validate=False,
+        validate=True,
+        validate_sections_only: bool = False,
     ):
         """Write the input file to disk.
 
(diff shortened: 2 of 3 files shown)


davidrudlstorfer:
Uses the new feature from https://github.com/4C-multiphysics/fourcipp/pull/61 to allow for section only validation during dumping

isteinbrecher (dismissed):
Thanks for this @davidrudlstorfer, at the moment I think this definitely makes sense to activate by default.

Just one minor question, is the runtime for the validation relevant or does it vanish compared to the overall runtime?

davidrudlstorfer:
@isteinbrecher I've resolved the merge conflict and rebased onto latest main.

Good idea regarding the performance check!

I've taken the example from #358 and got the following results for 100000 elements

**with validation**

Screenshot from 2025-06-04 16-08-34

**no validation** (same result as in #358)
Screenshot from 2025-06-04 16-08-40

So validation approximately added 0.9s runtime. I think this is feasible. If one does really large input files with millions of elements the validation can still be simply disabled. But most of the times the examples are much smaller and validation is anticipated. I would merge it in its current form

tests/create_cubit_input.py
```diff
@@ -318,7 +318,9 @@ def rotate_list(original_list, n):
     brick.translate([3 * 4, 0, 0])
     input_file.add(brick)
 
-    input_file.dump(file_path_blocks, add_header_information=False)
+    input_file.dump(
```
isteinbrecher, line 321: Why do we need this change here (and in the other changed test case)? Because we don't automatically do this in CubitPy?

@knarfnitram should we also enable this default behaviour in CubitPy?

isteinbrecher (commented):
Thanks for looking into this (and also the nice plots!). That is an overhead I am definitely willing to take in general development.

tests/create_cubit_input.py
davidrudlstorfer: No - because these test files do not have the `PROBLEM TYPE` key in the input file and would otherwise fail validation. Note, that `PROBLEM TYPE` is a required parameter of an input file in 4C.

With doing the "section only validation", we validate each section independently and, therefore, ignore if sections which are required are not present

isteinbrecher: Ah, and we are adding the problem type in MeshPy... Thanks for the explanation.

isteinbrecher (approved):

PR #363, opened 2025-06-02
Improve performance of rotation copy

davidrudlstorfer:
Improves the performance of the rotation functionality.

Notes:

- I also tested
```
return copy.copy(self)
```
which already was faster. But the current approach is much faster

- If the copy is not independent of the original rotation, the tests fail. With this approach all tests succeed.


isteinbrecher (approved):
Thanks a lot for this change @davidrudlstorfer , it is nice to see how much good a single changed line of code can do.

PR #364, opened 2025-06-02
Improve performance by disabling converter during dumping

davidrudlstorfer:
Part of #358


davidrudlstorfer:
Screenshot from 2025-06-02 15-13-43

This now again improves performance by 2.9s or 14.8% (for example of #358)

Last big thing is the `get_rotation_vector` method as discussed in #358. Then I would say we picked all low hanging fruits

isteinbrecher (approved):
I stand by my comment from #363 😂

PR #371, opened 2025-06-17
Allow for splinepy Bezier curves

isteinbrecher:
Allow the creation of beam filaments along splinepy Bézier curves.

@fabian-key, with this you should be able to directly use Bézier curves with `create_beam_mesh_from_nurbs`.

### DIFF
--- a/src/meshpy/mesh_creation_functions/beam_splinepy.py
+++ b/src/meshpy/mesh_creation_functions/beam_splinepy.py
@@ -19,7 +19,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
-"""Create a beam filament from a NURBS curve represented with splinepy."""
+"""Create a beam filament from a curve represented with splinepy."""
 
 import numpy as _np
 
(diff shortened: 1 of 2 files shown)


davidrudlstorfer (commented):
> One TODO that could be done in a follow up PR is to adapt the name of create_beam_mesh_from_nurbs since it now can also take Bézier curves, not just NURBS. This should be done with caution, as some production cases already use that function - so maybe with a deprecation warning or something like that.

Could you rename this directly in this PR? I am not 100% sure if i understand our "production cases". A rename of an individual function should be pretty straight forward? I would hope that we can prevent deprecation warnings in MeshPy as a whole. I think then we would have to discuss the deprecations in each PR and our community is so small that this should not make a difference.

P.S. that's a weird action failure... I've not seen that before, the nightly pipeline succeeded with Python 3.13 on Windows

isteinbrecher:
> I simply restarted the action and it succeeded 😅 so this was a weird bug 🐛

Weird, also just happened in #372 . Let's see if this becomes a regular thing, then we should look into it. As far as I have seen it has something to do with the c-compiler on Windows...

Our production case is, that we have project partners who use a MeshPy/4C pipeline in a container environment. Our partners more or less use predefined functions for a simulation and only vary the parameters for optimisation/ML applications. Whenever they change anything in the pipeline the container is rebuild with the newest MeshPy/4C versions which then causes failing pipelines and our partners have to wait for us to fix it 😅. This does not mean we should not change the API, just to be careful. But I think, in this case you are right, and a deprecation warning might be overkill. I added a second commit that changes the naming of the function(s) and file.

davidrudlstorfer (approved):
Thanks for changing it directly :)

PR #373, opened 2025-06-20
Improve solid material class data structure

isteinbrecher:
This allows for an easy definition of custom materials in user scripts, without having to encode every possible material in MeshPy.


davidrudlstorfer (approved):
Thanks for the addition :)

One question to better understand it: this basically allows us to also define materials which do not fit in the youngs_modulus + nu + density approach?

isteinbrecher:
@davidrudlstorfer yes exactly, one PR where this is needed is just about to follow.

PR #374, opened 2025-06-20
Add splinepy support for surface volume nurbs

isteinbrecher:
Until now we could either use a `splinepy` NURBS curve to create beam meshes (with "standard" elements, not NURBS), or we could use `geomdl` objects for actually NURBS elements in 4C.

`geomdl` is no longer supported, and in my opinion `splinepy` is the best NURBS package for python I know.
Thus, this PR allows us to now use `splinepy` objects to create NURBS elements for FEM solvers.

@eulovi the `geomdl` functionality is completely retained, so you should be able to use you scripts without any changes.

Two last points:
- I converted one existing 2D NURBS test case to create a shell model in 4C. This should be moved out of the NURBS testing functionality when we refactor those tests.
- After looking into this, it took me a while to figure out how things work. I think the NURBS code can be simplified (and unified) in a lot of parts. But, the important thing, everything is working for now.


davidrudlstorfer:
> geomdl is no longer supported, and in my opinion splinepy is the best NURBS package for python I know.
> Thus, this PR allows us to now use splinepy objects to create NURBS elements for FEM solvers.

> @eulovi the geomdl functionality is completely retained, so you should be able to use you scripts without any changes.

These two points somehow contradict each other 😅 should/can we remove the geomdl functionality?

src/meshpy/mesh_creation_functions/nurbs_generic.py
```diff
@@ -31,6 +31,84 @@
...
+
+        For a volume, the following information is stored:
+            Vertices: 'vertex_u_min_v_min_w_min', 'vertex_u_max_v_min_w_min', 'vertex_u_min_v_max_w_min', 'vertex_u_max_v_max_w_min',
+                      'vertex_u_min_v_min_w_max', 'vertex_u_max_v_min_w_max', 'vertex_u_min_v_max_w_max', 'vertex_u_max_v_max_w_max'
+            Edges: 'line_v_min_w_min', 'line_u_max_w_min', 'line_v_max_w_min', 'line_u_min_w_min',
+                   'line_u_min_v_min', 'line_u_max_v_min', 'line_u_min_v_max', 'line_u_max_v_max'
+                   'line_v_min_w_max', 'line_u_max_w_max', 'line_v_max_w_max', 'line_u_min_w_max'
+            Surfaces: 'surf_w_min', 'surf_w_max', 'surf_v_min', 'surf_v_max', 'surf_v_max', 'surf_u_min'
+            Volume: 'vol'
+    """
```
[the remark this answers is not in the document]
isteinbrecher: You are completely right, I saw this immediately after opening the PR, this is now adapted.

isteinbrecher:
> These two points somehow contradict each other 😅 should/can we remove the geomdl functionality?

What I meant with "geomdl is no longer supported" was, that it is no longer maintained and the package is basically idle sine 5 years (although I just had a look and there seems to at least be some activity going on this year).

We can't remove the support at the moment, because a lot of @eulovi research builds upon geomdl.

eulovi (approved):
@isteinbrecher, thanks for the contribution to the NURBS code! I got a bit worried when I read that geomdl is not longer supported but later I read your next comment ;)

This looks good to me and I agree that the code could be simplified. We can take a look at this in the coming weeks if necessary.

src/meshpy/core/nurbs_patch.py
```diff
@@ -66,33 +66,36 @@ def get_nurbs_dimension(self):
             )
         return n_knots
 
+    def get_number_of_control_points_per_dir(self):
+        """Return the number of control points in each parameter direction of
+        the patch."""
```
davidrudlstorfer, line 71: Could you fix the docstring and type hints (for the return value) here as well?

isteinbrecher: Sure! Return type is clear to me, but how do you think the doctoring should look like?

davidrudlstorfer: The `Returns:` is missing

Maybe something along the lines

```suggestion
    def get_number_of_control_points_per_dir(self):
        """Determine the number of control points in each parameter direction of
        the patch.
        
        Returns:
            List of control points per direction"""
```

isteinbrecher: Good point, to be honest, the return doc string has usually been ignored by me 😬. But let's do this going further!

isteinbrecher:
> I got a bit worried when I read that geomdl is not longer supported but later I read your next comment ;)

Sorry, I really should have chosen my words more carefully 😂

PR #376, opened 2025-06-23
Only build docu and website after merge to main

### LINKED ISSUE #230: Discuss strategy when to build documentation and website
I just noticed that we build the documentation and website every night, e.g., https://github.com/imcs-compsim/meshpy/actions/runs/13273540969 https://github.com/imcs-compsim/meshpy/actions/runs/13273558085. Do we really want to do this? I think it makes sense to only build this for every change to `main`. In my opinion the nightly pipelines are just to test that nothing breaks with the dependencies (python packages, 4C, cubit, ...).

@davidrudlstorfer can you estimate how much we need to change for that?

davidrudlstorfer:
(Hopefully) Closes #230

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,7 +1,7 @@
 name: Build documentation
 
 on:
-  # Only build the documentation once the test suite has completed on the main branch (only after merged PR's)
+  # Only build the documentation once the test suite has completed on the main branch
   workflow_run:
     workflows: [Test suite]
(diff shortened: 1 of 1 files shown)


davidrudlstorfer:
This simple fix should hopefully work. As always I am not 100% sure if it works in the end but let's test it.

We only need this guard during the docu build because the website only gets deployed once the docu has finished building

isteinbrecher (approved):

PR #377, opened 2025-06-23
Automatically build binder image after merge

### LINKED ISSUE #229: Automatically build `mybinder` image after merge to main
The examples in mybinder work very nice, however, the internal docker image for mybinder hast to be rebuild for each MeshPy commit. This means that when one wants to start the blind examples and this is the first time that this is done for the current commit on main, one might have to wait 10 minutes until the examples start (compared to under a minute otherwise).

This could be solved by building the image regularly (e.g. daily) or for each commit to main. There is a GitHub action that should do exactly that: https://github.com/s-weigand/trigger-mybinder-build
Since mybinder only has limited resources we should chose an approach that results in a reasonable usage of their resources.

davidrudlstorfer:
(Hopefully) Closes #229

### DIFF
--- a/.github/workflows/build_binder.yml
+++ b/.github/workflows/build_binder.yml
@@ -1,0 +1,20 @@
+name: Trigger Binder build
+on:
+  push:
+    branches:
+      - main
+
+# Allow only one concurrent build, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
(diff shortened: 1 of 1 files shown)


davidrudlstorfer:
In the original issue #229 you proposed to use this read-to-use github workflow => https://github.com/s-weigand/trigger-mybinder-build

I now refrained from using it because it was not updated in years.

Basically, all we need is to once open the website and the build is triggered. Let's see if it (hopefully) works. If not we can still try the much heavier workflow

PR #378, opened 2025-06-23
Split up conftest

davidrudlstorfer:
In preparation to overhaul the conftest setup for our Abaqus tests (see #356) I would propose to split up the conftest setup. With all latest additions it got really long and messy

I propose to split the file for now into the general "conftest" setup, the performance stuff and the result comparison stuff.


tests/conftest.py
```diff
@@ -19,32 +19,23 @@
...
-from fourcipp.utils.dict_utils import compare_nested_dicts_or_lists
-from vistools.vtk.compare_grids import compare_grids
 
 from meshpy.core.conf import mpy
-from meshpy.core.mesh import Mesh
-from meshpy.four_c.input_file import InputFile
 
-PERFORMANCE_LOG = {}
+# Import additional confest files (split for better overview)
+pytest_plugins = [
```
isteinbrecher, line 35: Is this something that pytest automatically detects?

isteinbrecher (approved):

PR #381, opened 2025-06-24
Open diff after failed comparison

davidrudlstorfer:
This PR reactivates the diff in VSCode for failed assertions. Initially I tried to implement this for all other file types as well (".inp", ".json", ...) but it turned out to be a pretty substantial amount of work (formatting data, files, so many different scenarios, ...)

Therefore, I only implemented the diff for our ".4C.yaml" files (99% of reference files).

One drawback is, that our current reference files are not ordered in fashion, and, therefore, often a little bit more is shown in the diff.

Nevertheless, this PR is part of #356 and closes

> - [ ] improve testing infrastructure (test abaqus input files also with compare_nested_list_or_dicts, remove entire string comparison from our side, also open a diff in VSCode for the yaml files)

### DIFF
--- a/tests/conftest_result_comparison.py
+++ b/tests/conftest_result_comparison.py
@@ -39,18 +39,23 @@ from meshpy.four_c.input_file import InputFile
 
 
 @pytest.fixture(scope="function")
-def assert_results_equal() -> Callable:
+def assert_results_equal(tmp_path, current_test_name) -> Callable:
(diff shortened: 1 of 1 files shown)


isteinbrecher (commented):
Thanks for bringing this feature up to date.

Should we sort all existing yaml files to ease the diff?

Issue #184, opened 2025-01-13
Unify numerical comparisons of arrays


isteinbrecher:
Currently we use a number of different comparison mechanisms for arrays. We should check if it makes sense to provide one unified approach.

Current mechanisms:

- [x] `np.testing.assert_array_less`
- [x] `np.allclose`
- [x] `np.isclose`
- [x] `np.linalg.norm(a - b) < tol`
- [x] `np.array_equal`
- [x] `np.abs(1.0 - node.coordinates[0]) < 1e-10`

- ...

PR #382, opened 2025-06-24
Introduce global test tolerance

### LINKED ISSUE #192: Possibly rethink / discuss handling of tolerance in new testing framework
Initially the local vtk tests did not succeed for me because if we do not specify a tolerance for `assert_results_equal` the default `None` is put into the `compare_grids` function. This function is based on numpy isclose

With the new support of Numpy 2, Numpy can now handle tolerances=None

We should probably rethink this current args approach to really not put any tolerances into the vtk_compare function if we do not specify any.

@isteinbrecher what do you think? What would be your ideas? As I've seen you are a lot better with all the args/kwargs

davidrudlstorfer:
Closes #192 

This PR introduces a global testing tolerance and removes all hard coded tolerances from all tests (which use the assert_results_equal functionality)

In another PR the array comparison should be also centralized, but this is #184 

I needed to update 2 reference files so everything works locally, let's see how the pipelines are doing


davidrudlstorfer:
@isteinbrecher very interesting! The pipeline fails on macOS (especially ARM based OSs). Should we increase the tolerance for this one test or the global test tolerance? I am curious on your opinion:)

tests/reference-files/test_cosserat_curve_mesh_warp_transform_boundary_conditions.4C.yaml
isteinbrecher, line 1: Yes, that file makes the issues (I created it on macOS). I had a quick look and could not figure out what is going wrong, where the tolerances can be increased - I suspect different versions of the scipy interpolate functionality, but I am not sure. Anyway, this should not concern us, let's simply put both tolerances for this test to `1e-12`, then everything should pass. I would then also not adapt this result file.

tests/reference-files/test_four_c_beam_to_solid.4C.yaml
```diff
@@ -131,6 +131,6 @@ STRUCTURE ELEMENTS:
   - 6 SOLID HEX8 6 11 12 7 20 23 24 21 MAT 1 KINEM nonlinear
   - 7 SOLID HEX8 8 7 15 16 22 21 25 26 MAT 1 KINEM nonlinear
   - 8 SOLID HEX8 7 12 18 15 21 24 27 25 MAT 1 KINEM nonlinear
-  - 9 BEAM3R HERM2LINE3 28 30 29 MAT 1 TRIADS 0 -1.57079632679 0 0 -1.57079632679 0 0 -1.57079632679 0
-  - 10 BEAM3R HERM2LINE3 31 33 32 MAT 1 TRIADS 0 -1.57079632679 0 0 -1.57079632679 0 0 -1.57079632679 0
-  - 11 BEAM3R HERM2LINE3 33 35 34 MAT 1 TRIADS 0 -1.57079632679 0 0 -1.57079632679 0 0 -1.57079632679 0
+  - 9 BEAM3R HERM2LINE3 28 30 29 MAT 1 TRIADS -0.0 -1.5707963267948968 -0.0 -0.0 -1.5707963267948968 -0.0 -0.0 -1.5707963267948968 -0.0
```
isteinbrecher, line 134: Nice, I marked this file as ported to double precision in https://github.com/imcs-compsim/meshpy/issues/356

isteinbrecher (approved):
> @isteinbrecher when this is merged I would open an issue to further investigate the test_cosserat_curve_mesh_warp_transform_boundary_conditions test

Thanks!

PR #383, opened 2025-06-24
Unify numerical comparison across testing framework

### LINKED ISSUE #184: Unify numerical comparisons of arrays

davidrudlstorfer:
Closes #184


davidrudlstorfer:
1. The cosserat tests are currently failing because the global tolerance is too tight. Should we redo the results?

2. With the current changes I introduced the numerical comparison into the test framework. Should we instead introduce it into the general MeshPy utils and also use it in MeshPy?

3. Should we globally "reconfigure" the tolerances? my.eps?

davidrudlstorfer:
Looks like the cosserat tests were created on macOS because there they succeed :D

isteinbrecher (commented):
I really like how this improves the testing comparisons. (Actually quire strange what constructs could be found here 😂)

isteinbrecher:
> - The cosserat tests are currently failing because the global tolerance is too tight. Should we redo the results?

What exactly do you mean with "redo the results"?

> - With the current changes I introduced the numerical comparison into the test framework. Should we instead introduce it into the general MeshPy utils and also use it in MeshPy?

Without thinking to long about that, I would say yes, that would be a nice follow up.

> - Should we globally "reconfigure" the tolerances? my.eps?

You mean to unify them or lower the tolerances?

davidrudlstorfer:
@isteinbrecher yesterday I gave this PR and connected issue some thought and work again and I have the following points which are open for discussion:

- Here cosserat curve tests fail similar to #382 due to too tight tolerances. Here the tests succeed on macOS and fail on the other OS's (here it is exactly switched around compared to #382, probably these results were created on macOS). I would add this test to #410 if that is alright for you. To merge this PR I would increase the tolerance for this test

- After some thought I am not that happy with the name of the comparison function `assert_numerics_equal` - what do you think? First of all we do not test that the numerics are equal and second "numerics" is a weird name. I think it would make more sense to use something like `assert_array_values_close`? Do you have a better idea? The same would hold for our standard `assert_results_equal` function? Does it make sense? Maybe `assert_results_close`?

- Initially I wanted to add the function to the BeamMe source and then simply import it in the testing framework. This way we could also use it in the source of BeamMe. But yesterday I moved it and as far as I could tell we have close to zero use cases in our src.

We have not a single use of `np.testing.assert_array_less`, `np.allclose0` and `np.array_equal` in our source code. And the provided function would not really replace the following ones:  `np.isclose`, `np.linalg.norm(a - b) < tol` and `np.abs(1.0 - node.coordinates[0]) < 1e-10`.

Therefore, does it even make sense to provide this functionality in the source of BeamMe?

My proposed next steps would be:

- Adjust the test tolerance for the one cosserat curve test so it succeeds
- Add this test to #410
- Rename `assert_numerics_equal` to something more descriptive
- Maybe rename `assert_results_equal` to something more descriptive
- Merge :rocket:

Let me know what you think:)

isteinbrecher:
Thanks for the overview!

> Here cosserat curve tests fail similar to https://github.com/beamme-py/beamme/pull/382 due to too tight tolerances. Here the tests succeed on macOS and fail on the other OS's (here it is exactly switched around compared to https://github.com/beamme-py/beamme/pull/382, probably these results were created on macOS). I would add this test to https://github.com/beamme-py/beamme/issues/410 if that is alright for you. To merge this PR I would increase the tolerance for this test

Sounds good to me.

The names for the assert functions I would leave up to you, whatever you seem fit. My only thought:
- `assert_array_values_close` does this also work for scalars? Then I would simply name it `assert_values_close`

> Initially I wanted to add the function to the BeamMe source and then simply import it in the testing framework. This way we could also use it in the source of BeamMe. But yesterday I moved it and as far as I could tell we have close to zero use cases in our src.
> We have not a single use of np.testing.assert_array_less, np.allclose0 and np.array_equal in our source code. And the provided function would not really replace the following ones:  np.isclose, np.linalg.norm(a - b) < tol and np.abs(1.0 - node.coordinates[0]) < 1e-10.
> Therefore, does it even make sense to provide this functionality in the source of BeamMe?

Thanks for looking into this in so much detail, then let's leave this functionality in the testing.

So I think, once the points you listed are done, we can merge this!

davidrudlstorfer:
As it turned it this was much simpler than initially expected - we do not need the array/value comparison function at all because our standard function does the identical job 😅 Therefore, I've removed the new function again and centralized everything.

This would be ready - all points from above are done (and I've renamed one oversight during the renaming).

PR #384, opened 2025-06-24
Refactor entire website

davidrudlstorfer:
- no more symlinks to the readme, instead a script copies the file
- new structure of the website to allow for external links


davidrudlstorfer:
The website now looks like this:

Screenshot from 2025-06-24 15-13-06

Notably we now no longer have the weird symlinked readme inside our website directory but copy it with a python script. This also would easily allow us in the future to remove the hardcoded links from the readme and again use relative links (to images, etc.)

We now also have an available "header" of the website which contains the links to our "external" pages.

I am happy with the website in its current form and would leave it like this for now.

PR #391, opened 2025-06-25
Install cubit in GitHub hosted container

isteinbrecher:
Install cubit in the GitHub hosted container, this would allow us to not need self-hosted runners.

Approach is the following, we download cubit and activate it using a academic-testing license (cubit support suggested this approach).

Only downside, the installation has to download 1GB which could drastically slow down the pipeline - let's see.


isteinbrecher:
Close in favour of #393

PR #392, opened 2025-06-26
Only create 50000 elements with cubit

isteinbrecher:
Tests with #391 look very promising, that we can fully switch our testing to only use GitHub hosted runners.

This requires one preliminary change, as we will use a test version of Coreform Cubit in testing, which is limited to 50,000 elements. In our performance testing we test with solid 100,000 elements. This number is more or less arbitrary, but I would opt for keeping it.

Thus this PR proposes the following solution: We create the solid with CubitPy with 50,000 elements and then read it with MeshPy, duplicate it and write it to an input file again, thus we have an input file with 100,000 solid elements.

The performance tests should fail right now as we will have to tune them to the performance of the GitHub hosted runners when (or after) merging #391.

### DIFF
--- a/tests/test_performance.py
+++ b/tests/test_performance.py
@@ -44,10 +44,8 @@ def shared_tmp_path(tmp_path_factory):
     return tmp_path_factory.mktemp("performance_tests")
 
 
-def create_solid_block(file_path, nx, ny, nz):
+def create_solid_block(cubit, file_path, nx, ny, nz):
     """Create a solid block (1 x 1 x 1) with (nx * ny * nz) elements."""
-
-    cubit = CubitPy()
 
     # Create brick.
     brick = cubit.brick(1)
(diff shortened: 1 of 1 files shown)


tests/test_performance.py
```diff
@@ -175,18 +169,53 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
 @pytest.mark.performance
 def test_performance_cubitpy_create_solid(evaluate_execution_time, shared_tmp_path):
-    """Test the performance of creating a solid block using CubitPy."""
+    """Test the performance of creating a solid block using CubitPy.
+
+    The version of Cubit we use in testing only allows for 50,000, so we
+    create a mesh with exactly that.
+    """
+
+    cubit = evaluate_execution_time(
```
isteinbrecher, line 178: With Coreform the initialisation of CubitPy takes a while so I moved this out of the mesh creation. We can talk about if if makes sense to time this here.

davidrudlstorfer: Do you have a rough approximation on how long the initialization takes?

davidrudlstorfer (dismissed):
Looks good:)

Thanks for the change, only a few minor questions

tests/test_performance.py
```diff
@@ -30,10 +30,6 @@
 from meshpy.four_c.input_file import InputFile
 from meshpy.four_c.material import MaterialReissner
 from meshpy.four_c.model_importer import import_four_c_model
-from meshpy.geometric_search.find_close_points import (
```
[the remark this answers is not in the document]
davidrudlstorfer: see #394

tests/test_performance.py
isteinbrecher: I had a look at this and this time varies very much, since we time here how long it takes for cubit to verify its license, so let's skip this for now.

tests/test_performance.py
```diff
@@ -175,18 +169,53 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
-            "file_path": shared_tmp_path / "performance_testing_solid.4C.yaml",
+            "cubit": cubit,
+            "file_path": shared_tmp_path / "performance_testing_solid_half.4C.yaml",
             "nx": 100,
-            "ny": 100,
+            "ny": 50,
             "nz": 10,
         },
-        expected_time=10.0,
+        expected_time=1.0,
```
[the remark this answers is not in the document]
isteinbrecher: I think so, and mostly for the license check.

tests/test_performance.py
```diff
@@ -171,18 +169,51 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
+        mesh_translated = mesh.copy()
+        mesh_translated.translate([1, 0, 0])
+        input_file.add(mesh)
+        input_file.add(mesh_translated)
+        input_file.dump(
+            shared_tmp_path / "performance_testing_solid.4C.yaml", validate=False
+        )
+
+    evaluate_execution_time(
+        "MeshPy: Double solid element block from cubit", double_block, expected_time=5.0
```
davidrudlstorfer, line 216: Maybe
```suggestion
        "MeshPy: Load double solid element block from cubit", double_block, expected_time=5.0
```

to align with the other tests?

davidrudlstorfer (commented):
Looks good.

1. Do you want to adjust the performance test timings directly in this PR?

2. On paper the total timing should be identical with this change, am I right? So we have around 20-30% overhead in our mesh add function?

tests/test_performance.py
```diff
@@ -171,18 +169,51 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
+    our goal is 100,000 so we double the block with 50,000 elements here."""
+
+    def double_block():
+        """Load the block with 50,000 elements and double the mesh."""
+        input_file, mesh = import_four_c_model(
+            shared_tmp_path / "performance_testing_solid_half.4C.yaml",
+            convert_input_to_mesh=True,
+        )
+        mesh_translated = mesh.copy()
+        mesh_translated.translate([1, 0, 0])
```
davidrudlstorfer, line 208: or are these the computational expensive operations here?

I do not fully understand where the overhead comes from. With the previous test we had 10s runtime and now 15s which is a 50% decrease

isteinbrecher:
> Do you want to adjust the performance test timings directly in this PR?

I would do that after we run the performance tests on GitHub hosted runners, since we will likely have to adapt many if not all performance tests.

isteinbrecher:
Don't worry, I am honestly happy for each critical question - see my attempt to avoid regex this morning 😂

> so we could simply move them into the regular test suite

We could to that, but then we have the issue on how to handle failed performance tests. I like the approach that we see one failed test (and the cross in the PR) but still have a "passing" pipeline. But maybe there is a nice alternative solution.

PR #393, opened 2025-06-26
Use Cubit in GitHub hosted runners

isteinbrecher:
This PR allows to use GitHub hosted runners with Cubit. This is done by actually installing cubit from an official source and activating a testing version with GitHub secrets.

The download, unpacking and license setup of Cubit takes between 60 and 80 seconds, so it is in the same range as pulling the 4C image - I think this is a reasonable increase in runtime for the pipelines.

This brings a number of advantages:
- We now have one test which tests MeshPy with all dependencies (ArborX, CubitPy, 4C)
- Everything is done in GitHub, no need to run any testing machines
- In the future we can now also easily test the ArborX performance

The one super tricky part was using GitHub secrets in standard PR workflows. Secrets are needed to activate CubitPy. I am still not 100% sure if everything works as expected, but we will have to see.

One problem is that we enforce forks, however PRs from forks are not allowed do use secrets. Thus we have to trigger the pipeline with `pull_request_target` instead of `pull_request` this will (hopefully) allow to use the secrets in every PR - BUT it will use the workflow definition from `main`, so users can not try to figure out the secrets.

This should work fine, but leaves one question: how to test changes to the testing workflow. This is achieved by having a second branch `ci-testing` in the main MeshPy repository. Branches on the main repo are allowed to use secrets. Pushes to `ci-testing` now trigger the pipeline and allow full secret access. So the idea is to have all testing related work on this branch in the future. Leaves one final issue, when we merge from `ci-testing` to main we would theoretically trigger two pipelines (one for the PR (but with the testing setup from `main`, and one for a push to `ci-testing`) this should be avoided by if guarding all tests.

Depends on #392 

Follow up PRs: Think about performance testing ArborX


.github/actions/setup_virtual_python_environment/action.yml
```diff
@@ -1,17 +1,12 @@
 name: setup_virtual_python_environment
 description: Configure the python virtual environment
-inputs:
```
davidrudlstorfer, line 3: Thanks for the clean-up, hopefully sooner than later I will update the virtual env process in our workflows. We should also use a conda env everywhere. This will drastically simplify everything and also make it more stable. I used the same approach in the workflows in FourCIPP

.github/workflows/testing.yml
```diff
@@ -4,19 +4,18 @@ on:
...
-      - reopened
-      - synchronize
+    branches: [main, ci-testing]
+  pull_request_target:
+    types: [opened, synchronize, reopened]
   workflow_dispatch:
-    type: choice
+
+env:
+  CUBIT_DOWNLOAD_URL: https://f002.backblazeb2.com/file/cubit-downloads/Coreform-Cubit/Releases/Linux/Coreform-Cubit-2025.3%2B58709-Lin64.tar.gz
```
davidrudlstorfer, line 13: Do we have any process to update this link? Does Cubit somewhere provide a link to the latest release?

isteinbrecher:
@davidrudlstorfer I think this is ready. Seems like that `pull_request_target` now does not trigger, so we only ran the code checks. I would be fine with merging this and seeing if this works.

davidrudlstorfer (approved):
Thank you @isteinbrecher :)

Let's get this merged and check if everything works

PR #395, opened 2025-06-27
Temporarily use last 4C release for testing

isteinbrecher:
Check if we can temporarily fix the pipelines by using the latest 4C release image.


isteinbrecher:
As I thought, this does not work, because we use the most recent `fourcipp`. So let's close this and adapt the api soonish.

PR #398, opened 2025-06-30
feat: rename MeshPy to BeamMe (readme + workflows)

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher I've not yet updated the citation for beamme, let's discuss how we want the new citation to look like in person next week. I'll add it to the renaming issue

PR #401, opened 2025-06-30
feat: rename code to beamme

davidrudlstorfer:


isteinbrecher:
There it works, so seems like that `pull_request_target` might not be the best long term option - we will see. At least everything is tested as expected in `main`.

Issue #403, opened 2025-07-01
Fix testing in PRs


davidrudlstorfer:
With https://github.com/beamme-py/beamme/pull/393 the entire testing suite was switched to Github hosted runners. To activate Cubit, credentials are provided via Github secrets.

To safely provide those secrets the trigger/cause of PR's was switched to

`pull_request_target`

Unfortunately, this only uses the base branch of the pull request, i.e., main branch. Therefore, no changes of a PR are tested.

This should be fixed so changes of a PR are actually tested again.

davidrudlstorfer:
So the current fix from #404 basically works (except one thing I missed to update during the website deployment) but the downside is that we now have to also approve the workflow run on main and the scheduled run (which is not how it should be)

So let's see how we can solve this

davidrudlstorfer:
So I researched for a little bit longer and found our (maybe) final solutions:

- We can simply create a second environment which does not need an approval and use that to provide the secrets to the runs on main

- And so we do not always need to approve the workflows in the PR we can simply check which status the author has (https://michaelheap.com/github-actions-check-permission/). We can discuss which level is expected, and I'll implement it.

- With these changes the secrets are safe, we test the latest code changes, we have to approve the pipeline for external/other contributors and I think we are pretty flexibel with future expansions

davidrudlstorfer:
This is now solved with #407 #412 and #414

We now have two separate environments which keep the Cubit secrets

- one needs an approval - this is necessary for PRs which are not opened from `CONTRIBUTORS` of the beamme repo
- one does not need an approval - this is used for PRs of `CONTRIBUTORS`, scheduled runs (each day) and merges to main

One thing that is not yet tested what happens if external people open PRs and if they actually can open a PR? But this will be investigated when it is necessary

PR #404, opened 2025-07-01
fix: testing during PRs

davidrudlstorfer:
- we now use environment secrets for PRs so maintainers need to approve new contributors
- the regular testing pipeline runs always

First try on fixing #403


davidrudlstorfer:
Note that the "standard" test suite is now running twice because the workflow is reused from main for `pull_request_target`

isteinbrecher (commented):
> Note that the "standard" test suite is now running twice because the workflow is reused from main for pull_request_target

Can we add an if in the workflows that will prevent this "double" testing?

PR #411, opened 2025-07-10
Store reference files in full precision

isteinbrecher:
This should move almost all reference files to full precision (see https://github.com/beamme-py/beamme/issues/356).

The only one that is missing is `test_meshpy_nurbs_import_solid_mesh`.

### DIFF
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -21,9 +21,9 @@ jobs:
     name: ubuntu-latest with all dependencies
     runs-on: ubuntu-latest
     environment:
-      # Use the trusted environment only for PRs authored by an OWNER or MEMBER (no approval needed); all others use the untrusted environment (someone has to approve the workflow run).
+      # Use the trusted environment only for PRs authored by an COLLABORATOR (no approval needed); all others use the untrusted environment (someone has to approve the workflow run).
       # Otherwise (scheduled or merge triggers) use the trusted environment (no approval needed).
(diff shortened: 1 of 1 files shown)


davidrudlstorfer:
@isteinbrecher is this already ready for review?

tests/test_beamme.py
```diff
@@ -1368,13 +1432,20 @@ def test_nurbs_import(assert_results_equal, get_corresponding_reference_file_pat
...
 
     # Add result checks.
     displacements = [
         [
-            -5.14451531793581718e-01,
-            -1.05846397858073843e-01,
-            -1.77822866851472888e-01,
+            -5.14451531199392575e-01,
+            -1.05846397823837826e-01,
+            -1.77822866488512921e-01,
```
isteinbrecher, line 1448: These had to be adapted because the old results worked with the "limited precision" control point coordinates. Since we are now using the full precision coordinates the results changed slightly (~1e-12).

davidrudlstorfer (approved):
@isteinbrecher thanks for tackling this!

> I added a comment explaining that this test might be a candidate to be moved / split.
Could we track this somewhere other than the ToDo in the code?

> Should we check in the scripts to create these test cases in the BeamMe repo? I did this for a couple of tests, but not all of them. I think that could be a good starting point if someone wants to work with a 4C test case, then one can find the corresponding script here.

Unfortunately, I must admit that I do not fully understand your points here. Which scripts to you mean? As far as I know most of the beam tests are created with other tools (just thinking of all the fiber network tests, brownian dyn tests, beam potential input files)

We've already decided that we do not really want to store "user" scripts in BeamMe, should we change this decision?

Maybe it would make sense to quickly open an issue for this and discuss this there.

Other than that - looks good and let's get this merged 🚀

isteinbrecher:
@davidrudlstorfer I had to merge the current `main` to this branch before being able to click the merge button (there were no merge conflicts). I assume this is due to the fact that the workflow on main changed compared to this branch. Let's keep an eye on that.

PR #412, opened 2025-07-10
Use trusted environment for special users

davidrudlstorfer:
If this works as expected, this would be the last PR to finish #403.

Now we check if a PR author is either a owner or member of the organization/repo, i.e., we should add our team members to the organization beamme-py rather than the repo.

This now results in the following:

- If a user, who is not in the orga, opens a PR we have to approve the cubit workflow
- If a user, who is in the orga, opens a PR we do *not* have to approve the cubit workflow
- The scheduled and merge to main workflows must not be approved (as it is currently)

See https://michaelheap.com/github-actions-check-permission/

### DIFF
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -21,9 +21,9 @@ jobs:
     name: ubuntu-latest with all dependencies
     runs-on: ubuntu-latest
     environment:
-      # If this workflow is triggered in a PR we use the untrusted environment (someone has to approve the workflow run),
-      # otherwise (scheduled or merge triggers) we use the trusted environment (no approval needed).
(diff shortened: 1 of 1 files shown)


isteinbrecher (approved):
Thanks for looking into this, let's see if it works!

davidrudlstorfer:
@isteinbrecher now we are in the weird situation that we do not really have a way to test changes to this workflow 😆 as far as I know pull_request_target really only works if the changes are on main. Do you know any workaround?

isteinbrecher:
We could use something like the `ci-testing` branch approach I initially proposed with the switch to cubit runners.

For testing purposes, we could also switch to `pull_request_target` and then check if the correct environment gets selected.

In this case, I would also be fine merging this to main and seeing if it works.

PR #414, opened 2025-07-10
Fix environment selction if user is privileged

davidrudlstorfer:
Follows #412

### DIFF
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -21,9 +21,9 @@ jobs:
     name: ubuntu-latest with all dependencies
     runs-on: ubuntu-latest
     environment:
-      # Use the trusted environment only for PRs authored by an OWNER or MEMBER (no approval needed); all others use the untrusted environment (someone has to approve the workflow run).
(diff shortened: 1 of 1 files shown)


davidrudlstorfer:
Interesting, so it probably does not work because I am an collaborator myself

davidrudlstorfer:
@isteinbrecher because we use forks and the test runs from the fork we can only see the "CONTRIBUTOR" status.

If we add that exception automatically all people which are invited to beamme can automatically run these workflows. Would that be fine for you or not?

davidrudlstorfer:
`That would be fine with me. Can someone who is not invited open a PR?`

@isteinbrecher I do not know about that specifically - we would probably need to tune the settings of this repo. I have no 2nd github account at hand so maybe you could try that out at IMCS

isteinbrecher:
> @isteinbrecher I do not know about that specifically - we would probably need to tune the settings of this repo. I have no 2nd github account at hand so maybe you could try that out at IMCS

Sounds good. From my point of view this can be merged.

davidrudlstorfer:
Part of #403

PR #416, opened 2025-07-11
Refactor NURBS geometry set creation

isteinbrecher:
Currently we kind of hard code the creation of all possible return sets for NURBS. This is quire lengthy and repetitive. This PR proposes to change this to a more generalised approach that reuses code where possible for different NURBS dimensions.

Also, a unit test case for the created NURBS sets is added.


davidrudlstorfer (approved):
Thanks for the nice simplification! 🚀

Issue #368, opened 2025-06-04
Improve performance of `get_rotation_vector`


davidrudlstorfer:
#358 revealed that the method `get_rotation_vector` of the `Rotation` class is a significant performance bottle neck of MeshPy. This should be investigated and possibly refactored for better overall performance.

isteinbrecher:
I had a look at this, and a possible solution can be found in https://github.com/isteinbrecher/beamme/tree/performance_rot_vec. This basically reduces the time required to evaluate the rotation vectors to 0, but requires an increased complexity in the mesh dumping functionality.

Also, the tests fail since we currently take special care when dumping the rotation vectors, which the library numpy quaternions does not. This is not needed for the "results" but with our current approach we guarantee unique rotation vectors which eases the comparison in testing.

For now, I would stick with the current slower approach, if we really need to fine-tune performance at some point we can revisit this (although the overall performance gain from this is "only" 30%, so we won't gain orders of magnitude in performance).

davidrudlstorfer:
@isteinbrecher thanks a lot for your investigation!

As I've already mentioned I am totally fine with the current performance of BeamMe because I am only creating models with a maximum of 1e3 elements.

Therefore, feel free to close this issue. The approach to speed things up is documented here as well for future endeavors :rocket:

PR #421, opened 2025-07-15
Use numpy quaternions for global quaternion operations

isteinbrecher:
While looking into #368 I found some quite outdated code that we implement ourselves manipulating global quaternion arrays. Since we support the numpy-quaternion package, a lot of this code can be replaced. This is done in this PR.

Let's see what the performance tests say.

### DIFF
--- a/src/beamme/core/mesh.py
+++ b/src/beamme/core/mesh.py
@@ -380,7 +380,7 @@ class Mesh:
         rot1 = _get_nodal_quaternions(self.nodes)
 
         # Apply the rotation to the rotation of all nodes.
-        rot_new = _add_rotations(rotation, rot1)
+        quaternions_new = _quaternion.as_float_array(_add_rotations(rotation, rot1))
 
         if not only_rotate_triads:
             # Get array with all positions for the nodes.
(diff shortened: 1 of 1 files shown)


isteinbrecher:
So the performance tests look pretty bad:
```
BeamMe: Find close points (brute force Cython)                         0.019s < 0.025s
CubitPy: Create half solid block                                       2.402s < 2.700s
BeamMe: Load and double solid element block from cubit                 9.784s < 10.500s
BeamMe: Load solid mesh (no full import)                               1.666s < 2.500s
BeamMe: Load solid mesh (full import)                                  2.761s < 3.500s
BeamMe: Create large beam mesh                                         4.357s < 4.500s
BeamMe: Rotate large beam mesh                                         1.777s > 0.500s
BeamMe: Translate large beam mesh                                      0.204s < 0.250s
BeamMe: Reflect large beam mesh                                        1.826s > 0.500s
BeamMe: Wrap large beam mesh around cylinder                           2.783s > 1.750s
BeamMe: Wrap large beam mesh around cylinder without check             1.710s > 0.500s
BeamMe: Find close nodes in large beam mesh                            0.336s < 0.400s
BeamMe: Add large beam mesh to input file                              8.510s < 10.000s
BeamMe: Dump input file with large beam mesh                           3.552s < 5.500s
BeamMe: Write beam mesh to VTK                                         3.841s < 4.000s
BeamMe: Write beam mesh to VTK with more segments                      7.582s < 7.750s
BeamMe: Space-Time: Create mesh in space                               0.004s < 0.010s
BeamMe: Space-Time: Create mesh in time                                2.335s < 3.000s
```

Maybe the hand coded implementation is not so bad after all 😂. I will convert this to draft, let's see if we want to merge this at some point in a different state.

PR #425, opened 2025-07-15
Beamme logo

isteinbrecher:
The last thing on my list regarding the name BeamMe. This PR adds the logo and the script to create it. The script is added to the testing, but it is only tested if the script does not throw any errors.


isteinbrecher:
I was afraid that the testing approach does not work. @davidrudlstorfer do you have an idea how we can store the script somewhere and test it without it being part of the actual BeamMe package?

davidrudlstorfer (approved):

Issue #430, opened 2025-07-24
Update beam elements from 4C


davidrudlstorfer:
Currently we support the following beam elements from 4C:

- `BEAM3R LINE3 HERMITE_CENTERLINE true`
- `BEAM3R LINE2 HERMITE_CENTERLINE true`
- `BEAM3K LINE3`
- `BEAM3EB LINE2`

@isteinbrecher as we've already briefly discussed we should also add the higher order (non-hermitian) elements, i.e., `BEAM3R LINE4, `BEAM3R LINE5` (I would also try them out now for different reasons)

Maybe it would make sense to overhaul the classes and just create `LINE2`, `LINE3`, `LINE4` classes with optional variables for the type (Reissner, Euler-Bernoully, Kirchhoff) and discretization (hermitian centerline)? This would simplify the setup by a lot and one no longer has to import all different classes if one wants to change the discretization

isteinbrecher:
@davidrudlstorfer can we close this with #431 completed or am I missing something?

PR #431, opened 2025-07-30
General beam element

isteinbrecher:
Draft to unify the beam elements in core and for 4C.

#430

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -2,7 +2,7 @@ name: Code quality
 
 on:
   schedule:
-    - cron: '0 04 * * *'
+    - cron: '0 06 * * *'
   push:
     branches:
       - main
--- a/.github/workflows/check_meshpy_redirects.yml
+++ b/.github/workflows/check_meshpy_redirects.yml
@@ -2,7 +2,7 @@ name: Check MeshPy to BeamMe redirects
 
 on:
   schedule:
-    - cron: "0 4 * * *"
+    - cron: "0 06 * * *"
 
 jobs:
   check-redirects:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -2,7 +2,7 @@ name: Test suite
 
 on:
   schedule:
-    - cron: '0 04 * * *'
+    - cron: '0 06 * * *'
   push:
     branches:
       - main
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -2,7 +2,7 @@ name: Protected test suite (can access cubit secrets)
 
 on:
   schedule:
-    - cron: '0 04 * * *'
+    - cron: '0 06 * * *'
   push:
     branches:
       - main


src/beamme/core/element_beam.py
```diff
@@ -341,3 +341,26 @@ def get_vtk(
...
+    nodes_create = _np.linspace(-1, 1, num=n_nodes)
+
+    # Create the beam class which inherits from the base beam class.
+    return type(f"Beam{n_nodes}", (Beam,), {"nodes_create": nodes_create})
+
+
+Beam2 = generate_beam_class(2)
+Beam3 = generate_beam_class(3)
+Beam4 = generate_beam_class(4)
+Beam5 = generate_beam_class(5)
```
davidrudlstorfer, line 366: Do we maybe have a better name for these? Somehow the naming does not really convey the difference (at least for me)

Additionally, do we want to use a consistent naming scheme across the code base?

beam_2, Beam2, Beam_2?

I do not really have a preference here

And last but not least, would it make sense to initialize them somewhere else? Somehow, it feels a bit wrong to put such a central thing just on the bottom of this file.

src/beamme/four_c/element_beam.py
```diff
@@ -36,83 +41,89 @@
 )
 
 
-class Beam3rHerm2Line3(_Beam):
+def dump_four_c_beam_to_list(self) -> dict:
```
davidrudlstorfer, line 44: That's really nice!

src/beamme/four_c/element_beam.py
```diff
@@ -36,83 +41,89 @@
...
+                for i in _INPUT_FILE_MAPPINGS["n_nodes_to_node_ordering"][
+                    len(self.nodes)
+                ]
+            ],
+        },
+        "data": element_data,
+    }
+
+
+class Beam3rHerm2Line3(_Beam3):
```
davidrudlstorfer, line 83: Should we here directly switch to a

`4C_beam`

 which has the number of nodes, beam_type and centerline as an argument? This way one would prevent importing all the different beam classes. I think this could be easier for the code development

isteinbrecher: Do you mean a function that creates the beam object similar to what we do in `generate_abaqus_beam`? I think providing something like that would be a good approach - although I would still provide shortcuts with the present names for backwards compatibility.

src/beamme/core/element_beam.py
isteinbrecher: For the case style, I would stick with `BeamN` (it technically isn't a `class` but it is used like one). I personally think that the `Beam2`, `Beam3`, ... naming is ok, I chose this because it is similar to hex8, hex20, ... . But I am open for other suggestions.

The placement is a good question. In this file we can only put it at the end because the names depend on all previous definitions. We could add a central file where we provide certain shortcuts, but that would then go a little bit against the decision we made when we removed the shortcuts from the `__init__.py` file and required uses to import from the correct file. In my opinion this is fine, because almost no-one will ever interact with this file and simply import via the IDE, that hopefully correctly suggests these names first (for me with VsCode that is the case). But I am open for other options.

davidrudlstorfer: Sounds good for me:)

src/beamme/four_c/element_beam.py
```diff
@@ -36,154 +40,127 @@
...
+            "valid_materials": [_MaterialReissner, _MaterialReissnerElastoplastic],
+            "coupling_fix_dict": coupling_fix_dict,
+            "coupling_joint_dict": coupling_joint_dict,
+            "dump_to_list": dump_four_c_beam_to_list,
+        },
+    )
+
+
+Beam3rHerm2Line3 = get_four_c_reissner_beam(n_nodes=3, is_hermite_centerline=True)
+Beam3rLine2Line2 = get_four_c_reissner_beam(n_nodes=2, is_hermite_centerline=False)
```
davidrudlstorfer, line 111: A few minor points to this

1. Should we also add these to the bottom of the file as in the beam core file?

2. Should we update the names to the new 4C layout, i.e., `Beam3rLine3Hermite`?

davidrudlstorfer: 3. I think for consistency we should also create a ready to use Kirchhoff beam?

And I do not know the specifics: do we only have one discretization available for Kirchoff beams (no different node counts?)?

src/beamme/four_c/element_beam.py
```diff
@@ -36,154 +40,127 @@
...
+            "four_c_element_data": four_c_element_data,
+            "valid_materials": [_MaterialKirchhoff],
+            "coupling_fix_dict": coupling_fix_dict,
+            "coupling_joint_dict": coupling_joint_dict,
+            "dump_to_list": dump_four_c_beam_to_list,
+        },
+    )
+
+
+class Beam3eb(_generate_beam_class(2)):
```
davidrudlstorfer, line 153: Same for the Beam3eb beam; should we add the number of nodes into the class name?

src/beamme/four_c/element_beam.py
```diff
@@ -36,154 +40,127 @@
...
+                for i in _INPUT_FILE_MAPPINGS["n_nodes_to_node_ordering"][
+                    len(self.nodes)
+                ]
+            ],
+        },
+        "data": element_data,
+    }
+
+
+def get_four_c_reissner_beam(n_nodes: int, is_hermite_centerline: bool) -> type[_Beam]:
```
davidrudlstorfer, line 80: These functions are really nice!

Should we also create one `get_four_c_beam(beam_type, n_nodes, is_hermite, weak, is_rotvec, fad, ...)` which then can create all beam types?

This could also be done in a follow-up

src/beamme/four_c/input_file_mappings.py
```diff
@@ -22,9 +22,17 @@
 """This file provides the mappings between BeamMe objects and 4C input
 files."""
 
+from typing import Any as _Any
+
 from beamme.core.conf import bme as _bme
+from beamme.four_c.four_c_types import BeamType as _BeamType
 
-INPUT_FILE_MAPPINGS = {
+INPUT_FILE_MAPPINGS: dict[str, _Any] = {
```
davidrudlstorfer, line 30: Nice changes

davidrudlstorfer (commented):
Sorry for taking so long with the review

src/beamme/four_c/element_beam.py
isteinbrecher: Yes we should (`Beam3ebHerm2`).

davidrudlstorfer (approved):
Thanks for your nice changes and new additions :)

Let's merge it in its current form and just see what we could need in the future. I think with this approach we can quickly add new "short cuts" to new beam elements in 4C.

Thanks a lot! Let's get this merged:)

PR #432, opened 2025-07-31
Make default base mesh item data to be empty dictionary

isteinbrecher:
Make the default data for the base data item an empty dictionary.


davidrudlstorfer (approved):

davidrudlstorfer:
not it works as expected :)

isteinbrecher:
Now it works, weird. Let's look further into this if this happens again.

isteinbrecher:
That could be. Can we test this somehow?

isteinbrecher:
Sounds good. Can you do that? Or we can simply try with #433

PR #433, opened 2025-07-31
Move beam type to four c

isteinbrecher:
The `BeamType` enum is only relevant for 4C. This PR moves this enum form the core to the 4C namespace.


isteinbrecher:
For some reason the performance test for the beam element creation fails now...

davidrudlstorfer:
That's really weird. The changes in #432 did not change anything - then #432 would fail as well.

Could it be the case the new enum is instantiated now all the time? Because only the test fails where we create a large number of beam elements

For me it would be no problem to just increase the timeout to something like 5s

isteinbrecher:
> Could it be the case the new enum is instantiated now all the time? Because only the test fails where we create a large number of beam elements

I had the same thoughts, I will try a few things.

> For me it would be no problem to just increase the timeout to something like 5s

Agree, but I would still like to know why 😂

src/beamme/four_c/four_c_types.py
```diff
@@ -0,0 +1,38 @@
...
+    """Enum for beam types."""
+
+    reissner = _auto()
+    kirchhoff = _auto()
+    euler_bernoulli = _auto()
+
+
+# Each time we call BeamTypeEnum directly there is a small performance overhead.
+# Thus, we do this assignment here to avoid that.
+BeamType = BeamTypeEnum
```
isteinbrecher, line 38: Super weird, but this seems to be the issue - you learn something new every day 😂

davidrudlstorfer:
So the performance test fail continuously and always on main, I now tried multiple times

For me it was weird that the change should make any difference. Maybe we got lucky with the two pipelines here

(Also AI said that this change is not able to have any performance difference)

isteinbrecher:
Continues to get weirder...

If we think this is a useful change (I think it is, moves 4C stuff out of core), I would suggest to simply increase the reference time.

PR #434, opened 2025-08-01
Temporary PR to test the environment deployment

davidrudlstorfer:
This PR aims to test what happens / if it can happen that the environment deployment gets outdated and one no longer can merge.

@isteinbrecher maybe you can approve


davidrudlstorfer:
@isteinbrecher maybe you could approve here, then we can check if the deployment gets outdated

isteinbrecher (approved):

davidrudlstorfer:
So right now it is still mergable, let's leave it for another 10hours, then it would be a day

PR #435, opened 2025-08-01
Adapt value of performance tests

isteinbrecher:
Adapt value of performance tests and remove not needed performance workaround.


davidrudlstorfer (approved):
Have you checked the last failed attempts? I do not know for sure but I think I saw a few values between 4.8 and 4.9 as well, maybe 4.9 or even 5 makes more sense. But feel free to merge

isteinbrecher:
Actually I didn't check, I updated the PR to have it at 5.0.

PR #439, opened 2025-08-14
Allow for full material import from cubit/input files

isteinbrecher:
Up to now, when importing CubitPy instances or input files, the materials were imported as plain dictionaries containing data. This PR actually creates a solid material object for (fully) imported input files.

This also revealed that some test cases were missing solid material definitions and thus ended up with solid and beams pointing to the same material ID. This is fixed with the first commit of this PR.


davidrudlstorfer (approved):
I think if this branch is rebased it should work - currently it only fails due to the old branch and the changes I added with the beam potential :)

Looks good - thanks a lot:)

PR #442, opened 2025-08-25
Named space time sets

isteinbrecher:
Add the names for sets for the space-time mesh output. Depends on #431 as the testing module already used the generic beam elements introduced there.


davidrudlstorfer (approved):
Thanks for also adding an associated test :)

PR #443, opened 2025-08-27
Update nightly test time to work with new FourCIPP version

isteinbrecher:
As suggested by @davidrudlstorfer in #436 let's schedule the nightly tests once all nightly jobs of upstream packages (e.g., 4C and FourCIPP) are done.

FourCIPP builds the metadata at `- cron: "0 5 * * *"`, so - cron: "0 6 * * *" should be fine here.

In summer time that should be 8:00 local time in Germany - a little bit earlier would be nice, but let's think about that if the nightly pipeline starts to fail regularly (which I hope it won't) an the BeamMe developers start to work before 8:00.

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -2,7 +2,7 @@ name: Code quality
 
 on:
   schedule:
-    - cron: '0 04 * * *'
+    - cron: '0 06 * * *'
   push:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


davidrudlstorfer:
@isteinbrecher 

> @davidrudlstorfer any idea why the pipeline fails?

As far as Github says your branch is out of date, i.e., does not yet have the changes I made today. In the protected test suite we can only test your branch and this misses the updates to the beam potential which are necessary with the latest FourCIPP version

PR #444, opened 2025-08-29
Add advanced locsys example

isteinbrecher:
@knarfnitram and myself had some discussions on how to use advanced locsys conditions in 4C.

I had a look at a relatively simple example with some complex locusts conditions (@knarfnitram basically what we talked about yesterday) and it works. However, defining the respective input with BeamMe is a little bit more tricky than expected. Let's use this as a starting point to discuss if there is the need for some utility functions within BeamMe.

What we do with the test case is up for debate. I actually think this one would be suitable for 4C as well.

On a final note, we should have a look if we can also adapt locsys to work on rotational degrees of freedom (currently they only act on displacements), but that is completely a 4C issue.


tests/test_four_c_simulation.py
```diff
@@ -1226,3 +1228,162 @@ def test_four_c_simulation_beam_to_beam_contact_example(
...
+        mesh.add(displacements_locsys_functions)
+
+        # Set the boundary condition and locsys on the outer end.
+        mesh.add(
+            LocSysCondition(
+                beam_set["end"],
+                Rotation.from_rotation_vector(
+                    [1, 1, 1]
+                ),  # This is a hack for now to ensure that we directly use the values from the functions.
+                function_array=rotation_vector_component_functions,
```
isteinbrecher, line 1323: This was one thing I did not like. For complex rotations, it only makes sense to have the actually rotation vector components stored in `function_array`, not some scaling factors for the rotation vector components of `rotation`.

My first suggestion would be to allow either the `rotation` argument to `LocSysCondition` or the `function_array` argument. If `function_array` is given, then the rotation vector will be simply set to [1,1,1] (as I have done here, in this hacky manner).

isteinbrecher: In 4C it is handled the same as for normal BC, the scaling value is multiplied with the function value.

knarfnitram (dismissed):

knarfnitram (approved):

PR #447, opened 2025-09-01
refactor: improve beam potential

davidrudlstorfer:
After many months I am finally updating my personal stuff to the new BeamMe/Yaml layout (until now I was lazy enough and did a lot of manual work :smile:)

In the process it turned out that the beam potential approach was pretty convoluted because you needed to pass an input file and mesh into the function.

Now I simply return the header dict and BC's. This eases the workflow in an external package by a lot and should make this a lot more flexible in use.

### DIFF
--- a/src/beamme/four_c/beam_potential.py
+++ b/src/beamme/four_c/beam_potential.py
@@ -22,7 +22,11 @@
 """This file includes functions to ease the creation of input files using beam
 interaction potentials."""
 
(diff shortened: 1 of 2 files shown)


isteinbrecher (approved):
Looks good to me!

davidrudlstorfer:
@isteinbrecher thanks for the review :)

Issue #418, opened 2025-07-15
Publish BeamMe to PyPI


davidrudlstorfer:
One larger goal is to publish BeamMe to PyPI. This issue should be used to discuss/note things which are important before we publish the wheels.

A few things which come to my mind:

- How do we want to version BeamMe? Probably semantic versioning
- Do we want to increase the version for each PR?
- Do we want to update the wheels after each PR? (this would require the previous point)
- Which wheels do we want to build?

#335 also contains many notes/information regarding packaging

davidrudlstorfer:
Note: Before we tackle this it would probably be really helpful to overhaul our virtual environment in our workflows. I think it is a good idea to do this prior to publishing wheels to PyPI. In FourCIPP I simply use a conda based approach which is pretty simple and straightforward

isteinbrecher:
> Note: Before we tackle this it would probably be really helpful to overhaul our virtual environment in our workflows. I think it is a good idea to do this prior to publishing wheels to PyPI. In FourCIPP I simply use a conda based approach which is pretty simple and straightforward

Sounds good. The only thing I would like to keep, is that BeamMe can be installed and used with "plain" `pip`, i.e., we don't depend on certain conda packages.

davidrudlstorfer:
> Installing BeamMe in an extra step is a very good idea! I am not that familiar with conda, what would the advantage be over simply using actions/setup-python@v5?

Then we also test with our preferred installation approach, conda envs. Basically everything boils down to

```
      - name: Set up virtual environment
        uses: conda-incubator/setup-miniconda@v3
        with:
          auto-update-conda: true
          conda-remove-defaults: true
          activate-environment: fourcipp
          python-version: ${{ matrix.python-version }}
```
(see https://github.com/4C-multiphysics/fourcipp/blob/main/.github/workflows/run_testsuite.yaml for example)

> Sounds very good. Side not (not super important) can we then get the version number from within BeamMe?

Very good question - yes that works flawlessly ;) also thought about that already for some future logging 👀 

> You brought up a very good point - should we just replace all tests where we install in non-editable mode with tests where we test from wheels? That should take about the same time in installing (I assume that building the wheel + installing it is about the same as installing via pip). And then we would actually test the wheels for the min/max python versions and all OS.

Yes I've imagined something like this. This probably will be easier to discuss once everything is ready and we can really check how the pipeline evolves.

davidrudlstorfer:
The updates to our environments and installations are done (see #530 and #533)

[...]

PR #451, opened 2025-09-17
Install CubitPy from pip

isteinbrecher:
Install CubitPy from `pip`. With this, all prerequisites for #418 should be fulfilled.


davidrudlstorfer (approved):
Nice! Also thanks for the fix in the readme

PR #453, opened 2025-09-17
Set version to v0.1.0

isteinbrecher:
Set the BeamMe version to 0.1.0. With this we can push to `pip` again and have a recent working version there.

This is just a "test", as I would like to test if the wheels can be build by pip directly.


isteinbrecher:
@davidrudlstorfer I think in the future, we should take an approach like we discussed in Freising.

davidrudlstorfer:
If you just want to push temporarily to pip again - do we need a fixed version in the repo? You could just set the version temporarily?

Based on #418 I want to initially evaluate on how our versioning looks like and how we want to do it.

For example what we discussed in Freising is that we simply base our version on git tags (identical to the approach I introduced in FourCIPP). This would allow us to really quickly introduce new versions, really simply trigger workflows to publish based on new tags and we could also really simply release new versions on Github.

I am a bit hesitant to now introduce a hard-coded version to remove it very soon again.

PR #454, opened 2025-09-17
Add manifest file to allow packaging of Cython files

isteinbrecher:


davidrudlstorfer:
Because if we build the wheels beforehand, i.e., for Windows, Linux and macOS, the Cython code is already compiled and we do not need to ship the `.pyx` file?

PR #458, opened 2025-09-18
Improve creation of coupling pairs

isteinbrecher:
Up until now we had a bug that the `coupling_factory` did not work with given geometry sets. This is fixed. Also, a unitest for the factory is added.


davidrudlstorfer (approved):
Thanks for the fix and also for updating the type hints and docstrings!

I really like that we now improve the docs with each PR

PR #459, opened 2025-09-18
Remove add_sets option

isteinbrecher:
The `add_sets` option was removed a wile ago from the basic mesh creation functions. This PR now removes all occurrences, since this construct is obsolete.


davidrudlstorfer (approved):
Thanks for the cleanup :)

PR #462, opened 2025-09-18
Update reference time for large beam mesh creation

isteinbrecher:
The performance tests for the large beam creation failed again (https://github.com/beamme-py/beamme/actions/runs/17825063589/job/50676237446?pr=460), very very likely not due to a change on our side. So let's raise the expected time.


davidrudlstorfer (approved):

isteinbrecher:
> And now we are at 4.855s in this run :D

Maybe copilot is trying to have fun with us

PR #467, opened 2025-10-01
Getting started example

isteinbrecher:
Add a small getting started example to the README that shows the very basics of how to use BeamMe. This is not intended as an alternative to the examples (big TODO...) but just to have a very simple example in a very visible spot.

The second commit adds a way to test the snippet.

The README is getting more and more convoluted, we should likely restructure/split it after we have updated the installation instructions after #418


knarfnitram (approved):

davidrudlstorfer (commented):
@isteinbrecher first of all big thanks for the initiative and providing a first draft! I think this will be a really nice addition so users can directly see how to interact with BeamMe.

A few points I want to mention (maybe a discussion would make sense here)

1. I think we could improve the getting started by not duplicating so much code, for example it would be totally sufficient to only show the beam creation once. In your proposal we have 2 material definitions which are identical (only the radius differs) and then 2 x the creation of a line (which is almost identical) and then the arc segment 2d. I think it would make sense to simplify this by a lot but add more comments, i.e., what a beamme mesh can do, then add one beam element into there and also use some rotation/translation + operation on a single node.

2. if we add some specific method (display_pyvista) we could also add the mesh to an input file?

3. As you've already mentioned, the readme get's longer and longer. Instead of splitting it up, I would somehow shorten it in a few sections.

4. I would remove the created image entirely, in my opinion it does not really help with the example?

5. Regarding the testing approach - really nice that this works. I was wondering a few things here as well. Is the code block still recognized as "python code" with the added `test`. Or should we just copy the approach from Queens 1:1 ( see https://github.com/queens-py/queens/pull/91 I think its really nice)

6. We already have the entire binder setup up-to-date and readily available. Should we maybe move the getting started example directly into there? Then we do not have to think about testing/length of the readme. Then we would just link it, the user sees it and can directly try it out.

PR #470, opened 2025-10-15
Cleanup getting started script

isteinbrecher:
Some small cleanup of the getting stated example.

### DIFF
--- a/tests/test_readme.py
+++ b/tests/test_readme.py
@@ -23,14 +23,14 @@
 
 import os
 import re
-from pathlib import Path
 
 import pytest
 
 from beamme.core.mesh import Mesh
 
 
-def extract_code_snippets() -> tuple[dict[str, str], list[str]]:
+@pytest.fixture(scope="session")
+def extract_code_snippets(pytestconfig) -> tuple[dict[str, str], list[str]]:
     """Parse README.md for fenced code blocks marked as tests.
 
     Code fences may be written in two forms:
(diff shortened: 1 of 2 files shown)


tests/test_readme.py
```diff
@@ -30,7 +30,8 @@
 from beamme.core.mesh import Mesh
 
 
-def extract_code_snippets() -> tuple[dict[str, str], list[str]]:
+@pytest.fixture(scope="session")
+def extract_code_snippets(repo_root: Path) -> tuple[dict[str, str], list[str]]:
```
davidrudlstorfer: ... just use 


```suggestion
def extract_code_snippets(pytestconfig: Any) -> tuple[dict[str, str], list[str]]:
```

here and ...

tests/test_readme.py
```diff
@@ -50,7 +51,7 @@ def extract_code_snippets() -> tuple[dict[str, str], list[str]]:
               `test:`) to the code string.
             - **snippets_unnamed**: list of code strings from unnamed test blocks.
     """
-    readme_content = Path("README.md").read_text(encoding="utf-8")
+    readme_content = (repo_root / "README.md").read_text(encoding="utf-8")
```
davidrudlstorfer: ... just use 


```suggestion
    readme_content = (pytestconfig.rootpath / "README.md").read_text(encoding="utf-8")
```

here? I think the fixture is a bit overkill although it works 😅

tests/test_readme.py
isteinbrecher: Good point, that makes total sense, I remove the fixture and added `pytestconfig` here.

PR #473, opened 2025-10-16
refactor: simplify model importer

davidrudlstorfer:
Part of #356

This would be one of the last points of #356.

The basics were done by chatgpt and then the code was overhauled by me. All local tests pass.
Main objective was to simplify the code in general. I also wanted to get rid of the `get_geometry_set_indices_from_section` function. This can then be deleted once the mesh conversion is simplified (next step of #356).

Curious on what you think @isteinbrecher :)

### DIFF
--- a/src/beamme/four_c/model_importer.py
+++ b/src/beamme/four_c/model_importer.py
@@ -165,16 +165,11 @@ def _extract_mesh_sections(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
         if not items:
             continue
 
-        # Find geometry key for this section
-        geom_section_keys = list(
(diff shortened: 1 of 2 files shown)


src/beamme/four_c/model_importer.py
```diff
@@ -106,177 +100,123 @@ def import_four_c_model(
...
+
+        # Find geometry key for this section
+        geometry_key = None
+        for key, val in _INPUT_FILE_MAPPINGS[
+            "geometry_sets_geometry_to_condition_name"
+        ].items():
+            if val == section_name:
+                geometry_key = key
+                break
+        if geometry_key is None:
```
isteinbrecher:
```suggestion
        else:
```
I think that should work.

PR #474, opened 2025-10-20
Update kirchhoff beams

isteinbrecher:
With https://github.com/4C-multiphysics/fourcipp/pull/107 and https://github.com/4C-multiphysics/4C/pull/1402 we can make the 4C beam to input file during more consistent.


src/beamme/four_c/four_c_types.py
```diff
@@ -31,3 +31,17 @@ class BeamType(_Enum):
...
+
+    weak = _auto()
+    strong = _auto()
+
+
+class BeamKirchhoffParametrizationType(_Enum):
+    """Enum for type of Kirchhoff parameterization."""
+
+    rot = _auto()
+    tan = _auto()
```
isteinbrecher: @davidrudlstorfer I am not 100% sure if enums are the best way to go here, an alternative could be to simply require strings in `get_four_c_kirchhoff_beam`. Do you have an opinion on this?

davidrudlstorfer: I do not really have a string opinion here 😅 probably the enums are a bit overkill here and your suggested strings would be more than enough. But feel free to continue as you like 😅

davidrudlstorfer:
Testsuite on main now failed as expected due to the Kirchhoff beams so this should be good to go

PR #475, opened 2025-10-20
Enhance Path-Treatment in Inputfile

knarfnitram:
The Path within the Inputfile is currently not updated.
Hence by switching the `Path` within your application script and activating  `add_footer_application_script=True` an error occurs. Here are the steps to reproduce the error:

```python
import os
from pathlib import Path

from beamme.four_c.header_functions import set_header_static
from beamme.four_c.input_file import InputFile


if __name__ == '__main__':

    input_file = InputFile()
    set_header_static(
        input_file, total_time=1.0, n_steps=1, tol_increment=1e-4, tol_residuum=1e-5
    )
    input_file.dump(os.getcwd()+'/myfile.4c.yaml')
    os.mkdir("test")
    file2=Path(os.getcwd()).joinpath("test/myfile.4c.yaml")
    os.chdir(file2.parent)
    input_file2 = InputFile()
    set_header_static(
        input_file2, total_time=1.0, n_steps=1, tol_increment=1e-4, tol_residuum=1e-5
    )
    input_file2.dump(file2)
```
Within this MR the class `InputFile` is enhanced by introducing a function called `_resolve_application_path` to detect the path based on the main function of a script. 

Do we need to add a test case for that?

### DIFF
--- a/src/beamme/four_c/input_file.py
+++ b/src/beamme/four_c/input_file.py
@@ -23,7 +23,6 @@
 4C."""
 
 import os as _os
-import sys as _sys
 from datetime import datetime as _datetime
 from pathlib import Path as _Path
 from typing import Callable as _Callable
@@ -48,6 +47,7 @@ from beamme.four_c.input_file_mappings import (
     INPUT_FILE_MAPPINGS as _INPUT_FILE_MAPPINGS,
 )
 from beamme.utils.environment import cubitpy_is_available as _cubitpy_is_available
+from beamme.utils.environment import get_application_path as _get_application_path
 from beamme.utils.environment import get_git_data as _get_git_data
 
 if _cubitpy_is_available():
@@ -186,8 +186,9 @@ class InputFile(_FourCInput):
                     lines = ["# " + line + "\n" for line in _INPUT_FILE_HEADER] + lines
 
                 if add_footer_application_script:
-                    application_path = _Path(_sys.argv[0]).resolve()
-                    lines += self._get_application_script(application_path)
+                    application_path = _get_application_path()
+                    if application_path is not None:
+                        lines += self._get_application_script(application_path)
 
                 with open(input_file_path, "w") as input_file:
                     input_file.writelines(lines)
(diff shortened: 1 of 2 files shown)


src/beamme/four_c/input_file.py
```diff
@@ -454,7 +463,31 @@ def _get_application_script(self, application_path: _Path) -> list[str]:
...
 
-        with open(application_path) as script_file:
-            application_script_lines.extend("# " + line for line in script_file)
+        if application_path is not None:
+            with open(application_path) as script_file:
+                application_script_lines.extend("# " + line for line in script_file)
 
         return application_script_lines
+
+    def _resolve_application_path(self) -> _Path | None:
```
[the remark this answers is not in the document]
davidrudlstorfer: I am also a bit hesitated if this is a viable use case? The option to not append the application script can simply be disabled and, if I understand it correctly, the path is changed during runtime? I think this is not an approach which should be supported?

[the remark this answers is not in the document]
knarfnitram: I think that it's a valuable enhancement, since it increases the robustness of the BeamMe InputFile generation.
 
Currently we use `_sys.argv[0]` to estimate automatically the path of the initial beamme driver file and  you are right, the footer can be suppressed by:
```input_file.dump("myfile-name",add_footer_application_script=False)```

For cases where you change as presented in the minimal example the directory, the estimation `_sys.argv[0]` is not correct any more and leads this _annoying_ error message:
```
Traceback (most recent call last):
  File "/home/a11bmafr/software/py-libs/meshpy/test.py", line 40, in <module>
    input_file2.dump(file2)
  File "/home/a11bmafr/software/py-libs/flowdiverter_simulations/src/meshpy/src/beamme/four_c/input_file.py", line 219, in dump
    lines += self._get_application_script(application_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/a11bmafr/software/py-libs/flowdiverter_simulations/src/meshpy/src/beamme/four_c/input_file.py", line 500, in _get_application_script
    with open(application_path) as script_file:
```
By adapting the path settings and actually searching for the real main driver file, the application script
is correctly added to the input file.

File: src/beamme/four_c/input_file.py [hunk 2]
```diff
@@ -186,12 +186,29 @@ def dump(
...
-                    application_path = _Path(_sys.argv[0]).resolve()
-                    lines += self._get_application_script(application_path)
+                    application_path = self._resolve_application_path()
+                    if application_path is not None:
+                        lines += self._get_application_script(application_path)
 
                 with open(input_file_path, "w") as input_file:
                     input_file.writelines(lines)
 
+    def _resolve_application_path(self) -> _Path | None:
```
isteinbrecher: I would put this into `src/beamme/utils/environment.py`.

src/beamme/utils/environment.py
```diff
@@ -107,3 +108,20 @@ def get_git_data(repo_path: _Path) -> _Tuple[_Optional[str], _Optional[str]]:
...
+
+
+def get_application_path() -> _Path | None:
+    """Returns the application path which created this input file.
+
+    and ensures that the file exists.
+    Args:
+    Returns:
+        a path to the file, which created this input file
+        or None.
```
isteinbrecher:
```suggestion
    """Returns the application path which created this input file.
    and ensures that the file exists.
    
    Returns:
        A path to the file, which created this input file or None.
```

src/beamme/four_c/input_file.py
knarfnitram: I think it's a good idea to put it into a separate function , because we already have two calls and then we can make sure that the path is found same (and I'm still not giving up on extending the function  `get_application_path` such that it will return a valid path for my case ;) ) . 
Jokes aside, I see the problem regarding the return type, because we have to check for `None` again and in the ideal case I'd get here always a valid path. I think there is not an empty path object (or something similar) which we could return, such that it skips the execution. So if we stick to the `Path | None` as return type of the function, that's the unfortunately the best option we have.

isteinbrecher:
Weird, I can click on merge. Is this the correct state? Then I can just merge it.

Issue #478, opened 2025-11-06
Discuss deep copy of GeometryNames and Mesh


isteinbrecher:
We can deep copy a `Mesh` with `mesh.copy()`. One has to be careful with node set containers returned by mesh creation functions, as they will always link to the mesh that they are created with.
So when calling `mesh.copy()` sets returned from mesh creation functions are currently "lost".

Good news is that there is a very simple solution:
```python
import copy

from beamme.core.element_beam import Beam3
from beamme.core.material import MaterialBeamBase
from beamme.core.mesh import Mesh
from beamme.mesh_creation_functions.beam_line import create_beam_mesh_line

mesh = Mesh()
beam_set = create_beam_mesh_line(
    mesh=mesh,
    beam_class=Beam3,
    material=MaterialBeamBase(radius=0.05),
    start_point=[0, 0, 0],
    end_point=[1, 0, 0],
)

# Use deep copy directly and explicitly proved the mesh and the sets
mesh_copy, beam_set_copy = copy.deepcopy((mesh, beam_set))
```

Should we provide a utility function that does like this? Or maybe add an example to the documentation of `Mesh.copy` how this can be done?

PR #480, opened 2025-11-07
Update documentation of mesh.copy

isteinbrecher:
#478


isteinbrecher:
Good points @davidrudlstorfer . I played around a little bit with the `--doctest-modules` function of `pytest`, but I could not find a solution that worked as I wanted.

I will add the code as an individual test case for `mesh` - but after #476 is merged, so we can use the generalised test materials there directly.

PR #482, opened 2025-11-14
Reorder all 4C reference files

davidrudlstorfer:
This PR now finally reorders all 4C reference files in BeamMe

The scipt I used is the following

```python
from pathlib import Path
from fourcipp.fourc_input import FourCInput

directory = Path("/home/rudlstorfer/01_work/01_workspace/beamme/")

if __name__ == "__main__":
    for file_path in directory.rglob("*"):
        if file_path.is_file():
            if str(file_path).endswith(".4C.yaml") or str(file_path).endswith(".4C.yml"):
                
                print(f"Reordering file: {file_path}")
                inputfile = FourCInput.from_4C_yaml(file_path)
                #inputfile.validate(sections_only=True)
                inputfile.dump(file_path, validate_sections_only=True)
```

Part of #356


isteinbrecher (approved):
Thanks for tackling this @davidrudlstorfer. Looks really good!

Issue #483, opened 2025-11-17
CubitPy testing is broken


isteinbrecher:
The licence used to activate Cubit during testing has expired in the last days. I will have a look into this tomorrow (18.11.25), hopefully this is an easy fix.

isteinbrecher:
Turns out #484 was not the solution, let's see...

isteinbrecher:
@davidrudlstorfer I looked into this, and I am not sure yet how we will proceed with the Cubit tests since the behaviour of the new student licences that they hand out is differently. I think it will take some time to figure out a solution. In the meantime I would be fine with deactivating the CubitPy (and thus also some of the performance) test, so we can still be operational. What do you think, do you have any other ideas?

PR #486, opened 2025-11-18
Temporarily deactivate Cubit in testing

isteinbrecher:
To allow for a continuation of our PR workflow, pipelines should pass. Therefore, this PR deactivates the currently not possible CubitPy tests. I think that enforcing 93% test coverage should still pass.

#483


isteinbrecher:
Let's see if this works 🤞

PR #489, opened 2025-11-19
Move some old "beamme" tests to the integration test folder

isteinbrecher:


davidrudlstorfer (approved):
A pretty dumb question - I just checked the tests and you just moved them from the "old" to the new file.

But the 6 reference files were not moved but are new? Shouldn't then the old ones be deleted? Sorry if I'm not getting the point here 😅

isteinbrecher:
> But the 6 reference files were not moved but are new? Shouldn't then the old ones be deleted? Sorry if I'm not getting the point here 😅

Good point, these are actually new reference files. There were some todos, to not only compare two meshes to each other but also to a fixed reference file:
https://github.com/beamme-py/beamme/blob/14ef1d40d7ea35fc054528bef6b031b1b7ffd63a/tests/test_beamme.py#L239-L241

I took care of those along the way.

PR #492, opened 2025-11-21
Reactivate Cubit testing

isteinbrecher:
This PR reactivates cubit testing. The new license model allows for a single activation. If the activated license is not released, the license can't be re-obtained in about 36 hours. Thus, this PR adds an action that releases the cubit license, which is always run (even on canceled jobs). Also, the performance tests are now run after the test suite, since both require the license.

One case that can still occur is that multiple parallel jobs require the license. For this case there is a mechanism, that it is tried for a certain amount of time to acquire the license until the job is failed.

### DIFF
--- a/.github/actions/cubit_setup/action.yml
+++ b/.github/actions/cubit_setup/action.yml
@@ -10,6 +10,10 @@ inputs:
   cubit_password:
     description: Password to activate cubit
     required: true
+  cubit_activation_timeout:
+    description: Timeout to try cubit activation in seconds
+    required: false
+    default: "300"
 outputs:
   cubit_root:
     description: "Path to the Cubit installation"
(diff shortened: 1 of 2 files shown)


.github/actions/cubit_setup/action.yml
```diff
@@ -10,15 +10,19 @@ inputs:
   cubit_password:
     description: Password to activate cubit
     required: true
+  cubit_activation_timeout:
+    description: Timeout to try cubit activation in seconds
+    required: false
+    default: "300"
```
davidrudlstorfer: Does this value make sense? Our tests take around 5min, so if two are started simultaneously (and more are added afterwards) all fail anyways?

.github/actions/cubit_setup/action.yml
```diff
@@ -37,5 +41,25 @@ runs:
...
+          now=$(date +%s)
+          elapsed=$(( now - start ))
+          if [ $elapsed -ge $timeout ]; then
+            echo "Cubit activation timeout reached after $elapsed seconds."
+            break
+          fi
+
+          # Delay between retries.
+          sleep 5
+        done
```
davidrudlstorfer: This looks a little bit complicated

ChatGPT just suggested something like

```bash
# Activate cubit. Retry login until success or timeout.
timeout="${{ inputs.cubit_activation_timeout }}"
start=$(date +%s)

while :; do
  if $CUBIT_ROOT/bin/rlm_activate --login \
      "${{ inputs.cubit_email }}" "${{ inputs.cubit_password }}"; then
    echo "Cubit activation succeeded."
    break
  fi

  elapsed=$(( $(date +%s) - start ))
  if [ $elapsed -ge $timeout ]; then
    echo "Cubit activation timeout reached after $elapsed seconds."
    break
  fi

  sleep 5
done
```

But I do not really care here 😅

davidrudlstorfer (approved):
Looks really good! Thanks for investing the time to fix the testing!

.github/actions/cubit_setup/action.yml
isteinbrecher: I also got it from ChatGPT, i asked to check that the exit code is equal to 0, I guess that is where the difference comes from.

PR #501, opened 2025-11-28
Fix cubit license loop

### LINKED ISSUE #496: [testing] Waiting for Cubit License is not working
Theoretically, we have a mechanism that repeatedly checks the cubit License until a timeout is reached. The following workflow failed and it does not look like the license was checked multiple times: https://github.com/beamme-py/beamme/actions/runs/19699849615/job/56433190468

isteinbrecher:
Closes #496

### DIFF
--- a/.github/actions/cubit_setup/action.yml
+++ b/.github/actions/cubit_setup/action.yml
@@ -13,7 +13,11 @@ inputs:
   cubit_activation_timeout:
     description: Timeout to try cubit activation in seconds
     required: false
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):
Thanks for the improvement :)

Regarding time-out: I would just increase it to something like 10-15min. We do not really care how long the job lasts but it's more important that it works at all.

davidrudlstorfer (approved):

Issue #456, opened 2025-09-17
Improve usage of type converter when dumping to input file


isteinbrecher:
As discussed here, the way we use custom type converting when dumping to an input file should be reworked
https://github.com/beamme-py/beamme/pull/446#discussion_r2354782611

- [ ] Introduce own BeamMe type converter
- [ ] Apply the custom type converter selectively only at places where we know that there might be the need for type conversion
- [ ] Check if this can also be applied to CubitPy https://github.com/beamme-py/beamme/pull/570#issue-3869224551

davidrudlstorfer:
If I remember correctly we currently convert the stuff during the addition to the input file, we could also think about only converting once we dump the file.

This would probably need a small change in FourCIPP because currently each item is automatically converted during the addition.

If we could cope with the performance degradation we could also just enable the conversion to happen twice, i.e., during the addition and during dumping (this is currently hardcoded to false). Then we would need no additional changes in FourCIPP

isteinbrecher:
I added some todos to the issue description.

PR #506, opened 2025-12-03
feat: remove inheritance of beamme input file on fourcipp

### LINKED ISSUE #449: Rethink inheritance of FourCInput from FourCIPP
After a very fruitful discussion with @gilrrei (thanks for the discussion!) it became clear that it is not the best idea to inherit our `InputFile` directly from the `FourCInput` from FourCIPP.

We should rethink that and one better way would be to have the FourCIPP object as a class attribute.

We would need to wrap any FourCIPP functions in our class implementation but through this our usage of the FourCIPP object gets a lot clearer.

One solution could look like this

```
class InputFile:
    """An item that represents a complete 4C input file."""

    def __init__(self, sections=None):
        """Initialize the input file."""

        # Instead of super(), explicitly wrap an instance of _FourCInput
        self.fourc_input = _FourCInput(sections=sections)

        # Contents of NOX xml file.
        self.nox_xml_contents = ""

        # Register converters to directly convert non-primitive types
        # to native Python types via the FourCIPP type converter.
        self.fourc_input.type_converter.register_numpy_types()
        self.fourc_input.type_converter.register_type(
            (_Function, _Material, _Node), lambda converter, obj: obj.i_global + 1
        )

    def add(self, object_to_add, **kwargs):
        """Add a mesh or a dictionary to the input file.

        Args:
            object: The object to be added. This can be a mesh or a dictionary.
            **kwargs: Additional arguments to be passed to the add method.
        """

        if isinstance(object_to_add, _Mesh):
            self.add_mesh_to_input_file(mesh=object_to_add, **kwargs)

        else:
            self.fourc_input.combine_sections(object_to_add)

    def dump(self, file_obj):
        ...
```

davidrudlstorfer:
Closes #449

This took me much longer than anticipated because I tried many things that did not work out. With the current status one test still fails locally. I need to further investigate this - therefore, draft status.

Overall this also needs a bit of clean-up before merging. @isteinbrecher feel free to have a first look, for example I thought if we should move the "internal helper functions" into a dedicated file?

### DIFF
--- a/src/beamme/four_c/input_file.py
+++ b/src/beamme/four_c/input_file.py
@@ -22,15 +22,19 @@
 """This module defines the classes that are used to create an input file for
 4C."""
 
+from __future__ import annotations as _annotations
+
 import os as _os
 from datetime import datetime as _datetime
 from pathlib import Path as _Path
+from typing import Any as _Any
 from typing import Callable as _Callable
 from typing import Dict as _Dict
 from typing import List as _List
 
 from fourcipp.fourc_input import FourCInput as _FourCInput
 from fourcipp.fourc_input import sort_by_section_names as _sort_by_section_names
+from fourcipp.utils.not_set import NOT_SET as _NOT_SET
 
 from beamme.core.conf import INPUT_FILE_HEADER as _INPUT_FILE_HEADER
 from beamme.core.conf import bme as _bme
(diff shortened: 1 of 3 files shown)


davidrudlstorfer:
So yes, we have a severe performance degradation for the `from_4C_yaml` function. I still don't know why...

src/beamme/four_c/input_file.py
```diff
@@ -78,24 +80,102 @@ def get_geometry_set_indices_from_section(
...
             (_Function, _Material, _Node), lambda converter, obj: obj.i_global + 1
         )
 
+    def __contains__(self, key: str) -> bool:
+        """Contains function.
+
+        Allows to use the `in` operator.
+
+        Args:
+            item: Section name to check if it is set
```
isteinbrecher: The documentation and function signature don’t match.

isteinbrecher (commented):
Thanks for tackling this! I only have some very minor comments.

Looking at the performance tests, I assume the issue with the type converter is fixed? Maybe we should not use the converter of fourcipp but add a beamme implementation (should be only a few lines of code), then we could also directly get rid of the workaround we use in the element creation - but thats beyond the scope of this PR.

src/beamme/four_c/input_file.py
davidrudlstorfer: What do you mean here, unfortunately I do not fully get it. `__contains__` is necessary for the "in" operator

i.e., `if "STRUCTURAL DYNAMIC" in input_file`

davidrudlstorfer:
@isteinbrecher yes the performance issue is fixed, is was a wrong initialization of the input file with `from_4C_yaml`.

The underlying problem was again the usage of the converter, I talked with Gil and it's probably really time to switch the FourCIPP converter to a BeamMe internal implementation, then we can also solve #456

src/beamme/four_c/input_file.py
isteinbrecher: Sorry, the comment was very cryptic, I meant that the argument is called `key` and in the documentation `item`.

isteinbrecher (approved):

PR #507, opened 2025-12-04
refactor: simplify mesh to input file addition

davidrudlstorfer:
Finally, this would be the last step of our big .dat to .yaml meta issue #356 

This PR simplifies the mesh to input file conversion. The ground work was done by ChatGPT but it needed a lot of manual intervention

### DIFF
--- a/src/beamme/four_c/input_file.py
+++ b/src/beamme/four_c/input_file.py
@@ -29,7 +29,6 @@ from datetime import datetime as _datetime
 from pathlib import Path as _Path
 from typing import Any as _Any
 from typing import Callable as _Callable
-from typing import Dict as _Dict
 from typing import List as _List
 
 from fourcipp.fourc_input import FourCInput as _FourCInput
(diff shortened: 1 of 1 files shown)


src/beamme/four_c/input_file.py
```diff
@@ -275,202 +250,151 @@ def dump(
...
+            if not items:  # do not write empty sections
+                return
+            dumped: list[_Any] = []
+            for item in items:
+                _dump_item_to_list(dumped, item)
 
-        start_index_functions = _get_global_start_function()
-        _set_i_global(mesh.functions, start_index=start_index_functions)
+            # Go through FourCIPP to convert to native types
+            if section_name in self.sections:
```
isteinbrecher: This can be simplified when we use our own type converter, correct? Should we add a TODO here?

isteinbrecher (approved):
Thanks for going the extra mile :rocket:

PR #519, opened 2025-12-23
Allow import nested materials

isteinbrecher:
A follow up to #515 . We can now also import nested sub materials from input files.


tests/conftest.py
```diff
@@ -316,6 +316,34 @@ def _get_default_test_solid_material(
...
+                                        data={"YOUNG": 2.0, "NUE": 0.0},
+                                    )
+                                ],
+                                "DENS": 1.0,
+                            },
+                        ),
+                    ],
+                    "DENS": 1.0,
+                },
+            )
```
davidrudlstorfer: Does it make sense to specify this as a fixture? As far as I can see this is only used once. Maybe it makes more sense to then just define it in the corresponding test? If the code is only used once I do not really see the advantage of defining it here.

PR #521, opened 2026-01-05
Fix performance issues

isteinbrecher:
Check if numpy version is reason for performance issues.


isteinbrecher:
It seems like the performance issues with the vtk test `BeamMe: Write beam mesh to VTK with more segments` is caused by a new numpy version. With `2.3.5` (only available with python 3.13) the test passes (https://github.com/beamme-py/beamme/pull/521/checks) with the newest `2.4.0` it fails (see the current main branch).

I propose to simply increase the expected reference time.

davidrudlstorfer (approved):
Thanks for the investigation!

PR #523, opened 2026-01-08
Allow Cubit tests to fail

isteinbrecher:
The Cubit workflow has caused some issues recently. This PR allows for failures of the full dependency test (including CubitPy). To still actively test the other dependencies, a test including 4C and ArborX is added to the unprotected test suite.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -47,3 +47,35 @@ jobs:
         with:
           name: ${{ github.job }}-${{ matrix.os-version }}-python${{ matrix.python-version }}-${{ github.run_number }}
           path: ${{ env.PYTEST_TMPDIR }}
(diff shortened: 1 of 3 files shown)


davidrudlstorfer (commented):
Looks good overall but one question, why do we now test the 4C and ArborX tests twice? They could be removed/disabled from the cubitpy tests if I am not missing something?

davidrudlstorfer (commented):
And one further question: if the cubitpy tests fail for a prolonged period of time the performance tests are also completely disabled? I think we should also just add the cubitpy flags to the performance tests so at least the "normal" (without cubitpy) performance tests run?

davidrudlstorfer:
Sounds good - thanks for the investigation :)

PR #529, opened 2026-01-12
Add examples to website

isteinbrecher:
This PR adds a html representation of the examples to the website. This includes an optional static/3D view of the `mesh.display_pyvista()` function, similar to how it is done in the pyvista documentation.

For now this is WIP, thus marking as draft. Also, if this is merged to main at some point, the examples should be extended before that, so we don't have the empty dummy example prominently on the website.


davidrudlstorfer:
wow, that looks super cool already!

isteinbrecher:
@davidrudlstorfer this is now ready for review. I reworked the existing examples and will push follow up PR once this is merged.

.github/workflows/website.yml
```diff
@@ -27,13 +27,20 @@ jobs:
           python-version: '3.14'
       - name: Install dependencies
         run: |
+          sudo apt-get install -y libosmesa6-dev
+          sudo apt-get install -y pandoc
+          sudo apt-get install -y xvfb
```
davidrudlstorfer:
```suggestion
          sudo apt-get install -y libosmesa6-dev pandoc xvfb
```

davidrudlstorfer: could you maybe explain why we need those now? maybe also in a comment

why the apt installation of pandoc?

src/beamme/core/mesh.py
```diff
@@ -1102,7 +1103,10 @@ def display_pyvista(
                         "All beams must have a defined cross-section radius."
                     )
 
-                beam_tube_grid = beam_grid.extract_surface()
+                # TODO: The `algorithm=None` argument is used to avoid a warning from
+                # pyvista regarding changed default arguments. For future pyvista
+                # versions, this argument can be removed.
+                beam_tube_grid = beam_grid.extract_surface(algorithm=None)
```
davidrudlstorfer, line 1109: Should we open an issue for that?

src/beamme/utils/visualization.py
```diff
@@ -0,0 +1,92 @@
...
+    Args:
+        plotter: The PyVista plotter to show.
+        nbsphinx_export_3d_view: For nbsphinx documentation, whether to
+            export an interactive 3D view alongside the static image.
+    """
+    if not _is_nbsphinx():
+        plotter.show()
+    else:
+        # Path where static documents are stored for the website.
+        static_doc_path = _Path(_os.environ["PYVISTA_DOCS_STATIC"])
```
davidrudlstorfer, line 60: Does this make sense to store as an environment variable? Cannot we hand that into the function?

src/beamme/utils/visualization.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Get a unique identifier for the current plotter
+        plotter_uid = str(_uuid.uuid4().hex)
+
+        # Export a screenshot of the plotter
+        plotter.screenshot(static_doc_path / f"{plotter_uid}.png")
+        static_frame_html = f"""
+            <img src="../_static/pyvista/{plotter_uid}.png"
+                style="max-width:100%; border-radius:8px;">
+            """
```
davidrudlstorfer, line 70: Is it possible to also add the dark theme here in accordance to the remaining website?

(I also wanna update the one completely white screenshot in the readme)

src/beamme/utils/visualization.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if nbsphinx_export_3d_view:
+            # Export a html representation of the plotter
+            plotter.export_html(static_doc_path / f"{plotter_uid}.html")
+
+            interactive_frame = _IFrame(
+                src=f"../_static/pyvista/{plotter_uid}.html",
```
davidrudlstorfer, line 77: can this relative path break?

website/docs/source/conf.py
```diff
@@ -36,6 +39,7 @@
 # extensions
 extensions = [
     "myst_parser",  # to enable Markdown support
+    "nbsphinx",  # to enable Jupyter Notebook support
```
davidrudlstorfer, line 42: do we still need this later when we do not use jupyter anymore?

website/docs/source/conf.py
```diff
@@ -49,3 +53,38 @@
...
+      </a>.
+      <br>
+      Interactive online version:
+      <a href="https://mybinder.org/v2/gh/beamme-py/beamme/main?urlpath={{ binder_urlpath }}">
+        <img alt="Binder badge"
+             src="https://mybinder.org/badge_logo.svg"
+             style="vertical-align:text-bottom">
+      </a>
+    </div>
+"""
```
davidrudlstorfer, line 82: cannot this move into the example.rst?

davidrudlstorfer (commented):
Already looks really really good

Just a few small comments

I guess a majority of the sphinx stuff was AI generated, I sometimes do not really understand it and I didn't invest the time to fully understand it :D

.github/workflows/website.yml
isteinbrecher: They are needed for the offscreen rendering of pyvista. I have no idea why, but pandoc is also needed as a system app, otherwise it does not work.

src/beamme/utils/visualization.py
isteinbrecher: Unfortunately that does not work, because the path is sort of a global variable while building the website. The environment variable was the only way I could pass that path to this function.

src/beamme/utils/visualization.py
isteinbrecher: Good question, I don't know. We can definitely create the screenshots with black and white background. But how we would then link the correct ones with sphinx, I don't know. I would move that to a future issue/PR.

src/beamme/utils/visualization.py
isteinbrecher: I will have a look, if we can use the `static_doc_path` from above.

website/docs/source/conf.py
isteinbrecher: Yes, because we only store the notebooks in the py:percent format, but they are still rendered as jupyter notebooks.

website/docs/source/conf.py
isteinbrecher: I am not really familiar with all these files, but I think it belongs here, since we set the options for how the website should be generated in this file. As far as I understand, the `example.rst` file is only to specify the notebooks.

src/beamme/utils/visualization.py
isteinbrecher: Unfortunately this is not super easy to do. As long as we only build notebooks in the examples directory, this will work. If we extend that, let's see if there is a more general solution.

src/beamme/core/mesh.py
isteinbrecher: Yes, we can do that.

.github/workflows/website.yml
```diff
@@ -27,13 +27,21 @@ jobs:
           python-version: '3.14'
       - name: Install dependencies
         run: |
+          # Needed for PyVista to render offscreen
+          sudo apt-get install -y libosmesa6-dev
+          sudo apt-get install -y pandoc
+          sudo apt-get install -y xvfb
```
davidrudlstorfer: As I suggested I would move the apt installs in one line then they should be quicker AFAIK

Then let's merge :)

isteinbrecher: Sorry, I missed that.

davidrudlstorfer (approved):

isteinbrecher:
Looks like everything worked and the examples are now shown on the website!

One small issue, the visualisation in my binder does not work, because of a method we call in pyvista that does not exist any more. Likely this has not been working for a while and is not related to this PR. I will push a fix.

PR #530, opened 2026-01-13
feat: improve env creation in workflows

davidrudlstorfer:
First part of #418 

This is a draft on how fast the conda setup process is. Now all env creations of all workflows are centralized. If conda is (far) too slow I wanna try caching the setup between runs.

Additionally, I want to pull out the requirements install from the actuall testing process into a dedicated action.


davidrudlstorfer:
My big pros with the proposed workflow are the following:

- the whole env setup is centralized
- we can set a beamme wide default python version
- we have the output on the python version in every step
- we do no longer need to activate or play around with shells depending on conda or venv
- we can super quickly switch out conda and venv with just one boolean flag

(I will push a commit to check if it also works if conda is turned on by default, but I will undo that commit again)

.github/actions/setup_python_environment/action.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo "$CONDA_PREFIX/Scripts" >> "$GITHUB_PATH"
+          fi
+
+        else
+
+          echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
+
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            echo "$PWD/.venv/Scripts" >> "$GITHUB_PATH"
+          fi
```
isteinbrecher: I think this else case can be completely removed.

isteinbrecher:
@davidrudlstorfer I am seeing that windows is causing issues. I think the issue is that `venv` don't work exactly the same on windows as on unix systems. Before you invest too much time, I have an idea:

Add a third option `system_python` (the others `venv`, `conda`) to `setup_python_environment` and then for `system_python` only call `actions/setup-python@v5`, and don't create the `venv` (and no exporting to paths) - then we know that everything should work for Windows as this is the approach that we currently use.

davidrudlstorfer:
@isteinbrecher now this would be ready for review, thanks for your hint on the additional option - I included it.

A short explanation:

- we now have three ways to source python, via a system install, via the venv after the system install or via conda
- the system and venv install works out of the box on all systems (Ubuntu, macOS and Windows)
- the conda install works out of the box for Ubuntu and macOS, if it should be used on Windows in the future as well one needs to use the login shell (bash -l {0} instead of just bash). Then it also works

I think we can merge this as is so we can see that the other workflows also work on main.

I would then open another PR to split up the testing and dependency install if that's fine for you.

(I now switched our testing to the venv because it has no overhead in a performance/time perspective and then we finally test with an empty env rather than the system environment)

.github/actions/setup_python_environment/action.yml
```diff
@@ -0,0 +1,61 @@
+name: setup_python_environment
+description: Setup a Python environment with conda or venv with a configurable Python version
```
isteinbrecher:
```suggestion
description: Setup a Python environment either with conda, venv or the system python. The version can be configured.
```

PR #531, opened 2026-01-14
Update testing

isteinbrecher:
This PR adds three things:
- Use the minimal 4C image
- Add a timeout to the beamme tests using cubit
- Reorder the steps in the cubit tests, so cubit is activates just before the beamme tests are run

@davidrudlstorfer on a related note, we are currently installing beamme and running the tests in a single step - would it make sense to split that?


davidrudlstorfer (dismissed):
Nice improvement!

PR #534, opened 2026-01-14
feat: introduce beamme version

davidrudlstorfer:
third part of #418 

we now automatically retrieve and install the version from the tag. If the last version is connected to an older commit it is automatically v0.0.1.post

This is the same approach as in FourCIPP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ api-documentation/*
 # website folder which contains automatically copied files
 website/docs/source/md/*
 
+# BeamMe version file
+src/beamme/version.py
+
 
 ### Automatically created ignores ###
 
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -0,0 +1,2 @@
+[run]
+omit = src/beamme/version.py
```
isteinbrecher: Can we move any other options from the `pyproject.toml` here?
ChatGPT tells me:
```
[run]
source =
    src/beamme

omit =
    */tests/*
    src/beamme/version.py

[report]
fail_under = 90
```

davidrudlstorfer: Ok I take everything back. I can move the omitted file into pyproject.toml

That is now available with never versions. I will push that immediately

isteinbrecher (approved):

PR #537, opened 2026-01-15
feat: build wheels in CICD

davidrudlstorfer:
part of #418

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,0 +1,40 @@
+name: Build wheels
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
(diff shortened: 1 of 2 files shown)


davidrudlstorfer:
@isteinbrecher I would aim to merge the current status of this PR. I will need to add the testing of the wheels to the testing workflow. This will only be possible once it is merged to main because the 

```
workflow_run:
    workflows: [Build wheels]
    types: [completed]
```

only works once it's merged to main. 

We now basically build the wheels for python 3.13 during PR's, and nightly and on new tags for 3.10,3.11,3.12,3.13

the buliding works super fast and the wheels also work

isteinbrecher (approved):

PR #540, opened 2026-01-16
Fix ArborX

isteinbrecher:
Cubit works again, yet the tests fail due to ArborX not being able to load. 

Possible issues:
- ~[ ]~ Some symbols not available on the minimal 4C image
- [x] We now test on the 4C image with python 3.13 instead of 3.12, maybe there are some version issues with pybind11


davidrudlstorfer (approved):
Thanks for looking into this and providing the quick fix

PR #564, opened 2026-01-21
Move mesh tests

isteinbrecher:
Move some of the mesh creation tests.


davidrudlstorfer (approved):
Thanks for splitting things up and also for introducing the parametrize feature for the helix :)

PR #597, opened 2026-02-19
Add mesh representation

isteinbrecher:
This PR adds a mesh representation to BeamMe. This mesh representation allows for a common interface to BeamMe meshes. For now, we use this mesh representation when importing models in BeamMe.

### DIFF
--- a/src/beamme/core/mesh_representation.py
+++ b/src/beamme/core/mesh_representation.py
@@ -1,0 +1,201 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2018-2026 BeamMe Authors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+"""This module defines the mesh representation data structure."""
+
+from dataclasses import dataclass as _dataclass
+from typing import Any as _Any
+
+import numpy as _np
+import pyvista as _pv
+from numpy.typing import NDArray as _NDArray
+
+from beamme.core.conf import Geometry as _Geometry
+from beamme.core.conf import bme as _bme
+
+MESH_REPRESENTATION_MAPPINGS: dict[str, _Any] = {}
+MESH_REPRESENTATION_MAPPINGS[
+    "element_type_and_n_nodes_to_connectivity_mapping_beamme_to_vtk"
+] = {
+    # Only list the non-standard mappings
+    (_bme.element_type.solid, 20): [
+        0,
+        1,
+        2,
+        3,
+        4,
+        5,
+        6,
+        7,
+        8,
+        9,
+        10,
+        11,
+        16,
+        17,
+        18,
+        19,
+        12,
+        13,
+        14,
+        15,
+    ],
+    (_bme.element_type.solid, 27): [
+        0,
+        1,
+        2,
+        3,
+        4,
+        5,
+        6,
+        7,
+        8,
+        9,
+        10,
+        11,
+        16,
+        17,
+        18,
+        19,
+        12,
+        13,
+        14,
+        15,
+        24,
+        22,
+        21,
+        23,
+        20,
+        25,
+        26,
+    ],
+}
+MESH_REPRESENTATION_MAPPINGS[
+    "element_type_and_n_nodes_to_connectivity_mapping_vtk_to_beamme"
+] = {
+    # Only list the non-standard mappings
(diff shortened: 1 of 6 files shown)


src/beamme/core/mesh_representation.py
```diff
@@ -0,0 +1,201 @@
...
+        11,
+        16,
+        17,
+        18,
+        19,
+        12,
+        13,
+        14,
+        15,
+    ],
```
davidrudlstorfer: is there a way that we put those in one line and disable the formatter here?

src/beamme/four_c/model_importer.py
```diff
@@ -115,101 +158,119 @@ def _extract_mesh_sections(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
         A tuple (input_file, mesh). The input_file is modified in place to remove
         sections converted into BeamMe objects.
     """
+    if input_file.contains_external_mesh_based_geometry():
+        raise NotImplementedError(
+            "Importing external mesh-based geometry from 4C input files "
+            "is not yet implemented."
+        )
```
davidrudlstorfer, line 165: will this come in the future?

src/beamme/four_c/input_file.py
```diff
@@ -487,3 +487,12 @@ def _get_application_script(self, application_path: _Path) -> list[str]:
...
         return application_script_lines
+
+    def contains_external_mesh_based_geometry(self) -> bool:
+        """Check if the input file contains external mesh-based geometry.
+
+        Returns:
+            True if the input file contains external mesh-based geometry, False otherwise.
+        """
+
+        return False
```
davidrudlstorfer, line 498: why do we need this if it is always false?

src/beamme/four_c/model_importer.py
```diff
@@ -218,23 +279,159 @@ def _extract_mesh_sections(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
...
+    ):
+        if node_type == _bme.node_type.node.value:
+            mesh.nodes.append(_Node(node_coordinates))
+        else:
+            raise ValueError(
+                f"Mesh conversion for node type {_bme.node_type(node_type).name} is not implemented!"
+            )
+
+    # extract elements
+    ## first create the element types
```
davidrudlstorfer: i think here we can still use only one hashtag as comments
```suggestion
    # first create the element types
```

also applies below

tests/performance/test_performance_beamme.py
```diff
@@ -232,7 +232,7 @@ def double_block():
     ("log_name", "full_import", "expected_time"),
     [
         ("BeamMe: Load solid mesh (no full import)", False, 2.5),
-        ("BeamMe: Load solid mesh (full import)", True, 3.5),
+        ("BeamMe: Load solid mesh (full import)", True, 4.7),
```
davidrudlstorfer: what is the reason that this is now much slower? have you profiled it?

src/beamme/four_c/model_importer.py
```diff
@@ -115,101 +158,119 @@ def _extract_mesh_sections(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
...
+            "Importing external mesh-based geometry from 4C input files "
+            "is not yet implemented."
+        )
+    else:
+        (
+            mesh_representation,
+            element_type_id_to_data,
+            node_set_internal_id_to_legacy_id,
+        ) = _extract_mesh_representation(input_file)
+    return _create_mesh_from_mesh_representation(
```
davidrudlstorfer, line 172: why do we differentiate here between `_extract_mesh_representation` and `_create_mesh_from_mesh_representation`?

src/beamme/four_c/model_importer.py
```diff
@@ -115,101 +158,119 @@ def _extract_mesh_sections(input_file: _InputFile) -> _Tuple[_InputFile, _Mesh]:
...
-    mesh = _Mesh()
+def _extract_mesh_representation(
+    input_file: _InputFile,
+) -> tuple[_MeshRepresentation, dict[int, dict], dict[int, int]]:
+    """Extract the mesh representation from mesh data directly contained in the
+    input file.
 
-    # extract materials
-    material_id_map_all = {}
+    This will do an inplace removal the mesh data from the provided input file.
```
davidrudlstorfer:
```suggestion
    This will do an inplace removal of the mesh data from the provided input file.
```

src/beamme/core/mesh_representation.py
isteinbrecher: Yes we can do that with `# fmt: off` and `# fmt: on`, we already use that in some testing files. I am personally tending more to the approach to let the formatter take care of everything, but we can change that here - it would improve the readability of the code.

src/beamme/four_c/model_importer.py
isteinbrecher: Yes, this is just a placeholder until also the import from binary, e.g., exodus, meshes will be possible.

src/beamme/four_c/input_file.py
isteinbrecher: Similar to the previous point, at the moment we don't need it, but I already added some the basic functions that have to be implemented later on.

src/beamme/four_c/model_importer.py
isteinbrecher: Good point, I will take care of that.

tests/performance/test_performance_beamme.py
isteinbrecher: Thanks for the point, I did not profile it and just assumed that the increased time is due to the overhead that we now convert the 4C input to a mesh representation and then the mesh representation to the BeamMe mesh. Turns out that is not the case for the performance degration, actually there are some issues when iterating over the points saved in a pyvista grid - I will have a closer look at that.

src/beamme/four_c/model_importer.py
isteinbrecher: In the future, I also want to implement the import of exodus meshes (which is the desired use case for large simulations). With this structure, we only have to implement a `_extract_mesh_representation_from_exodus` and can reuse the `_create_mesh_from_mesh_representation` function.

tests/performance/test_performance_beamme.py
isteinbrecher: I had a closer look - this is indeed an issue with pyvista, the fix is relatively easy. However, I am enchanting more and more issues with pyvista here (strange effects when renaming data structures, no empyt meshes, ...) that we might want consider switching to an alternative format in the future.

davidrudlstorfer (approved):
Looks good, thanks for the explanations and the fixes regarding formatting and performance:)

isteinbrecher:
So the performance tests failed with `BeamMe: Load solid mesh (full import)                                  3.500s > 3.500s` 😂. But that makes sense, because there is a little overhead compared to how the import was done previously. I would still merge this, and monitor how that continues in the next days.

PR #619, opened 2026-04-09
feat: add two half pass approach to beam potential

davidrudlstorfer:
FYI: test pipeline does not fail because we do not write out the file during testing (the validation happens during the write), but the pre-commit hook where we verify the input file fails

Follows https://github.com/4C-multiphysics/4C/pull/1957


isteinbrecher (approved):

davidrudlstorfer:
pipeline now succeeds because the changes are in fourcipp

PR #620, opened 2026-04-13
feat: update to Python 3.14

davidrudlstorfer:
quick test if it fails / what fails

### DIFF
--- a/.github/actions/setup_python_environment/action.yml
+++ b/.github/actions/setup_python_environment/action.yml
@@ -5,7 +5,7 @@ inputs:
   python-version:
     description: "Python version to use"
     required: false
-    default: "3.13"
+    default: "3.14"
 
   env_type:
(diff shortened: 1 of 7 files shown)


davidrudlstorfer:
@isteinbrecher as expected this does not work out of the box

- https://github.com/PyCQA/docformatter/issues/340 is an upstream problem in docformatter (already fixed but a new version needs to be released)

- more importantly, splinepy still has no wheels released for Python 3.14, that's why each run is taking >> 10min because splinepy needs to be built from source

Once splinepy is available for Python 3.14 we can continue this.

davidrudlstorfer:
(The tests which are not run are set in our settings as expected, that's why they show up => Python 3.10 and 3.13)

Additionally, the performance tests have nearly doubled in speed.

Closing this until splinepy is available for Python 3.14

isteinbrecher:
@davidrudlstorfer thanks for having a look, I contacted the splinepy developers, a python3.14 release should be on the way.

PR #621, opened 2026-04-13
Add element to previously empty test case

isteinbrecher:
Until now, the reference file `tests/reference-files/test_integration_four_c_model_importer_non_consecutive_geometry_sets_input.4C.yaml` contains no element. This causes issues when switching to a mesh representation #597 #594 as vtu does not support unstructured grids with no cells. To fix this, this PR adds a single element to that test case.


davidrudlstorfer (approved):
Do we really want to switch all tests to vtu in the future or keep a few as text based input as well?

isteinbrecher:
> Do we really want to switch all tests to vtu in the future or keep a few as text based input as well?

I would keep all the tests we currently have and only add a few new tests to test the external mesh format feature.

davidrudlstorfer:
> I would keep all the tests we currently have and only add a few new tests to test the external mesh format feature.

But does then every import in the future need a solid element? Wouldn't it make sense to just throw an exception and catch/test that?

isteinbrecher:
> But does then every import in the future need a solid element? Wouldn't it make sense to just throw an exception and catch/test that?

Only for the cases where we convert the inout file to a BeamMe representation of the nodes and elements. If one just wants to read a plain text based input file, that will still work with no elements.

PR #624, opened 2026-04-14
Add cubit script for element types

isteinbrecher:
There was still one solid input file without the corresponding cubit script in BeamMe.


tests/other/test_other_create_cubit_input_files.py
```diff
@@ -112,3 +124,22 @@ def test_other_create_cubit_input_files_solid_brick(
...
     create_solid_brick(result_path, get_default_test_solid_material)
     assert_results_close(result_path, reference_file)
+
+
+@pytest.mark.cubitpy
+@pytest.mark.xfail(
+    condition=not is_cubit_coreform,
+    reason="Meshing of tetrahedra is highly sensitive - we only "
+    "require the test to pass for Cubit Coreform.",
+)
```
davidrudlstorfer, line 134: Could you explain this? What's the difference between cubit and coreform?

Furthermore, I am not super happy that we test cubitpy functionality in BeamMe more and more 😅

davidrudlstorfer: This is really a pure cubitpy test and nothing else

isteinbrecher: CubitCoreform is the new name of what used to be Cubit. We test with CubitCoreform. However, CubitCoreform does not support MacOS any more, so there one has to resort to older Cubit versions. For tet meshes they provide different node numbering thus the tests would fail. The `xfail` is only triggered if one runs this on a machine with an old Cubit version - in GitHub CI testing, we always require this to pass.

I am not so sure about this myself. We could simply remove this skip and always require the test to pass - then the tests will fail for people who are developing Cubit stuff on Mac - that is basically just me 😅. But since these tests are not part of daily work, I would be fine with removing this swtich/xfail.

> Furthermore, I am not super happy that we test cubitpy functionality in BeamMe more and more 😅
> This is really a pure cubitpy test and nothing else

We have some files in the repository that are required for testing purposes. Most of these files are created with some scripts and we test these scripts in the BeamMe pipeline. As you pointed out, these tests do not directly test BeamMe. I still prefer having them in the repo and tested. This allows to easily recreate the testing files in case we want/have to change something. If we would not keep the scripts around, adapting these referenced files can be a cumbersome effort.

davidrudlstorfer: > I am not so sure about this myself. We could simply remove this skip and always require the test to pass - then the tests will fail for people who are developing Cubit stuff on Mac - that is basically just me 😅. But since these tests are not part of daily work, I would be fine with removing this swtich/xfail.

Yes, would be nice if we could remove this entirely. It's a really specific point which should in my opinion not covered in the code. This could for example briefly be mentioned in the readme, i.e., add a sentence on the min required cubit version and that otherwise tests could fail.

tests/other/test_other_create_cubit_input_files.py
```diff
@@ -23,14 +23,26 @@
...
+if cubitpy_is_available():
+    from cubitpy import CubitPy, cupy
+
+    # TODO: We have to initialize CubitPy for the config to be loaded. Once this
+    # is improved in CubitPy, adapt this here.
+    cubit = CubitPy()
+    is_cubit_coreform = cupy.is_coreform()
+else:
+    is_cubit_coreform = False
+
```
davidrudlstorfer, line 45: Let's remove the xfail stuff and the checks for coreform, then let's directly merge this:)

(Maybe also the above mentioned readme extension)

tests/other/test_other_create_cubit_input_files.py
isteinbrecher: I removed the `xfail` part and now this is handled the same way as we do in the logo testing.

I decided not to add anything to the readme, because CubitPy (and therefore, BeamMe) still supports older Cubit versions - we just can't recreate all testing files with the older versions.

davidrudlstorfer (approved):
Thanks @isteinbrecher 🚀

PR #625, opened 2026-04-15
Update 2D NURBS API from 4C

isteinbrecher:
#616


davidrudlstorfer (dismissed):
Same question as always 😅 why is this not failing now?

And should we wait to merge this until the pipeline fails (probably tomorrow morning)? Then we would see if the pipeline shows our expected behaviour

isteinbrecher:
> Same question as always 😅 why is this not failing now?

This is not failing because the new input was already enabled a while back, with the new changes in 4C only the old elements were deleted. We can wait until tomorrow, then the pipeline will fail. I also want to add a few more improvements according to copilots remarks.

davidrudlstorfer:
https://github.com/beamme-py/beamme/actions/runs/24497665939

After a little problem in FourCIPP (the access tokens ended yesterday...) the beamme pipeline now "successfully" fails.

So this could be merged

davidrudlstorfer (approved):

PR #626, opened 2026-04-17
Change how geometry sets are numbered

isteinbrecher:
This PR changes how we index geometry sets in BeamMe. We currently assign ascending indices for geometry stets for each geometry type, i.e., the first point set and the first surface set will both have the ID 1. For the legacy input in 4C, this works fine. However, for mesh format input, the geometry sets indices should be unique across all geometry types. This PR changes the indexing scheme to assign unique indices for geometry sets across all geometry types.

This is a breaking change, however, only in the case that users are directly using the assigned geometry set indices. Ordering of all data structures in BeamMe is preserved. Likely, the only thing that will really break with this PR is if someone is testing their code by comparing the dumped yaml files. In that case, the IDs in the yaml files will have to be updated, as is done in this PR.

#594


src/beamme/core/mesh.py
```diff
@@ -272,7 +270,7 @@ def get_unique_geometry_sets(
         *,
         coupling_sets: bool = True,
         link_to_nodes: str = "no_link",
-        geometry_set_start_indices: _Optional[_Dict] = None,
+        geometry_set_start_index: int = 0,
```
davidrudlstorfer, line 274: Rather stupid question but I cannot find it quickly in this PR. If this defaults to 0 can we just completely remove this variable?

Is there any use case where we want to set it to something non zero?

davidrudlstorfer (dismissed):
@isteinbrecher thanks a lot for the breaking change note :D it will break my input scripts ;)

Just one question regarding the code

src/beamme/four_c/input_file.py
```diff
@@ -299,7 +300,7 @@ def add_mesh_to_input_file(self, mesh: _Mesh) -> None:
         # Add sets from couplings and boundary conditions to a temp container
         mesh.unlink_nodes()
         mesh_sets = mesh.get_unique_geometry_sets(
-            geometry_set_start_indices=start_indices_geometry_set
+            geometry_set_start_index=start_index_geometry_set
```
davidrudlstorfer, line 303: Aaah it's used here, then the above comment can be ignored

isteinbrecher: Yes, we need that here at the moment to account for already existing geometry sets in an input file. But your comment contained a very good idea, this is actually not really well placed in `Mesh`. At the moment we need it, but I am pretty sure we can remove this once we switch to an internal mesh representation. I added a todo for this.

davidrudlstorfer (approved):

davidrudlstorfer:
@isteinbrecher would it be possible that we add a "breaking change" label to such beamme PRs? It now took me a really long time to figure out what was going wrong in my repo and why the tests were failing :D I know you already added it to the description, but maybe a simple label would make sense.

I only updated beamme to the latest hash and all my tests failed because I compare input files :D

isteinbrecher:
Yes that makes a lot of sense. I will create such a label.

PR #628, opened 2026-04-23
Improve data structure for mesh representation

isteinbrecher:
This PR improves the data structure of the mesh representation. We now store the pure numpy arrays instead of the pyvista unstructured gird. We can still convert the data structures to an unstructured grid later on, but now we have full control over all the data and avoid some strange pyvista / vtk behaviour.

With this, we can also store mesh representations with zero points or cells, which causes issues in pyvista.

### DIFF
--- a/src/beamme/core/mesh_representation.py
+++ b/src/beamme/core/mesh_representation.py
@@ -23,9 +23,9 @@
 
 from dataclasses import dataclass as _dataclass
 from typing import Any as _Any
+from typing import Iterable as _Iterable
 
 import numpy as _np
-import pyvista as _pv
 from numpy.typing import NDArray as _NDArray
 
 from beamme.core.conf import Geometry as _Geometry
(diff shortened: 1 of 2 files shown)


src/beamme/core/mesh_representation.py
```diff
@@ -111,48 +111,59 @@ def __init__(
...
-                or (cell_data is not None and len(cell_data) > 0)
-                or (point_data is not None and len(point_data) > 0)
-            ):
-                raise ValueError(
-                    "If cell_connectivity is None or empty, all other data must be None or empty as well."
-                )
-            self._grid = None
-        else:
-            # We use a unstructured vtk grid as main mesh data structure
-            self._grid = _pv.UnstructuredGrid(cell_connectivity, cell_types, points)
```
davidrudlstorfer, line 130: Would it make it easier if we use the ustructured grid from vtk directly and not via pyvista?

davidrudlstorfer (dismissed):

davidrudlstorfer (dismissed):

src/beamme/core/mesh_representation.py
isteinbrecher: I don't think so, the pure `vtk` module in python is quite cumbersome to use.

davidrudlstorfer (approved):

PR #629, opened 2026-04-24
Fix numbering of NURBS patches and add test case

isteinbrecher:
This PR fixes a bug that lead to wrong numbering of patch IDs added to BeamMe input files.


tests/integration/test_integration_four_c.py
```diff
@@ -390,6 +391,70 @@ def test_integration_four_c_nurbs_import(
...
+    mat = get_default_test_beam_material(material_type="reissner")
+    beam_set = create_beam_mesh_line(
+        mesh_beam, Beam3rHerm2Line3, mat, [-1.0, 0.0, 0.0], [1.0, 0.0, 0.0], n_el=2
+    )
+    mesh_beam.add(beam_set)
+    input_file.add(mesh_beam)
+
+    input_file.add(mesh_2)
+
+    # Compare with the reference file
```
davidrudlstorfer, line 454: Not important here, but such comments are not super helpful and maybe easier to be just removed. We have a lot of them so we can also keep it here:)
```suggestion
```

davidrudlstorfer (approved):

tests/integration/test_integration_four_c.py
isteinbrecher: Good point, if we have many of them, I would remove them all together.

PR #630, opened 2026-04-24
Remove some deprecated functionality

isteinbrecher:
Remove some functionality that is never used.


davidrudlstorfer (approved):
I thought that was important functionality 😅 as far as I remember I wanted to remove that once but we still needed it 😅

isteinbrecher:
> I thought that was important functionality 😅 as far as I remember I wanted to remove that once but we still needed it 😅

No idea what we still needed it  😅 - but I had a look, and I am now pretty sure we don't need it any more.

PR #633, opened 2026-04-27
Mesh representation as internal input file storage

isteinbrecher:
This PR changes the 4C input file, such that all added mesh information (except legacy dat style strings) is stored inside a `MeshRepresentation` in the input file. For now nothing changes in the behaviour. In the future, this will allow us to chose at `dump` if the mesh information is written to the input file or stored as a external mesh format.

#594


src/beamme/core/element.py
```diff
@@ -63,6 +68,30 @@ def replace_node(self, old_node, new_node):
...
+        if valid_materials is None:
+            return
+
+        for material_type in valid_materials:
+            if isinstance(self.material, material_type):
+                break
+        else:
+            raise TypeError(
+                f"Element of type {type(self)} can not have a material of type {type(self.material)}!"
+            )
```
davidrudlstorfer: Is this actually necessary to check within beamme? I think at least for me it would be enough if this error surfaces in 4C and we could remove the entire logic from beamme

src/beamme/core/mesh.py
```diff
@@ -266,9 +277,8 @@ def get_unique_geometry_sets(
         self,
         *,
         coupling_sets: bool = True,
+        # TODO: Check if we need all of the following options.
```
davidrudlstorfer, line 280: one todo left here?

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+        dict[_NURBSPatch, int],
+    ]:
+        """Create a mesh representation for this mesh.
+
+        This function does not alter the mesh object. It assigns internal IDs to the mesh object and returns mappings between the objects and the IDs.
+
+        Returns:
+            mesh_representation: `MeshRepresentation` object for this mesh.
+            element_type_id_to_data: A dictionary that maps the element type id to the data of the element type.
+            geometry_sets_to_i_global: A dictionary that maps geometry sets to their global index in the mesh representation.
```
davidrudlstorfer, line 740: would it make sense to unify this "id" and "i", both of them are IDs

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+        all_materials = [
+            material
+            for mesh_material in self.materials
+            for material in _get_all_contained_materials(mesh_material)
+        ]
+        material_to_i_global: dict[_Material, int] = {}
+        for material in all_materials:
+            material_to_i_global[material] = len(material_to_i_global)
+        if len(all_materials) != len(material_to_i_global):
+            raise ValueError("Materials are not unique!")
```
davidrudlstorfer: How can this happen? Can this also happen if I just assign the same material to two different beam sets?

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+        # Extract nodes for the mesh representation and assign global IDs. We also extract
+        # other required information like the node type, the nodal rotation vector and
+        # control point weights information here. The weights are optional, so the array
+        # might be `None`.
+        n_nodes = len(self.nodes)
+        if n_nodes != len(set(self.nodes)):
+            raise ValueError("Nodes are not unique!")
+        points = _np.zeros((n_nodes, 3))
+        point_types = _np.full(n_nodes, -1)
+        control_point_weights = None
```
davidrudlstorfer, line 772: Could we maybe add nurbs to this variable name? I am not working a lot with them so it would be easier to differentiate and directly see that we handle nurbs stuff here

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+        nodal_rotation_vectors = _quaternion.as_rotation_vector(nodal_quaternions)
+
+        # Extract elements for the mesh representation
+        if len(self.elements) != len(set(self.elements)):
+            raise ValueError("Elements are not unique!")
+
+        #   We first have to loop over all elements, so we get the total
+        #   number of elements, as NURBS patches can contain multiple elements.
+        #   This is needed to initialize the numpy arrays with the correct size.
+        i_element = 0
```
davidrudlstorfer:
```suggestion
        # We first have to loop over all elements, so we get the total
        # number of elements, as NURBS patches can contain multiple elements.
        # This is needed to initialize the numpy arrays with the correct size.
        i_element = 0
```

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+
+        #   Now that we know the expected size, we can allocate the data arrays and
+        #   actually gather the element data.
+        element_type_to_id: dict[type, int] = {}
+        element_type_id_to_data: dict[int, dict] = {}
+        cell_connectivity = []
+        cell_types = _np.full(n_elements, -1)
+        cell_element_type_ids = _np.full(n_elements, -1)
+        cell_material_ids = _np.full(n_elements, -1)
+        cell_beamme_element_ids = _np.full(n_elements, -1)
```
davidrudlstorfer, line 820: Why do we "allocate" the data here? Cannot we just append the data? I think it doesn't make a huge difference in python compared to memory management in cpp?

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+            if type(element) not in element_type_to_id:
+                element_type_id = len(element_type_to_id)
+                element_type_to_id[type(element)] = element_type_id
+                element_type_id_to_data[element_type_id] = _copy.deepcopy(
+                    type(element).data
+                )
+            else:
+                element_type_id = element_type_to_id[type(element)]
+
+            # Some elements don't require a material, in this case we set the material id to -1.
```
davidrudlstorfer: Could we specify which elements do no require a material? I have no clue :D

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+            material_id = material_to_i_global.get(element.material, -1)
+
+            if isinstance(element, _NURBSPatch):
+                n_patch_elements = element.get_number_of_elements()
+                # We can to a cast here, since we know that we have set i_global previously
```
davidrudlstorfer:
```suggestion
                # We can do a cast here, since we know that we have set i_global previously
```
?

src/beamme/core/mesh.py
```diff
@@ -714,6 +721,227 @@ def get_nodes_by_function(self, *args, **kwargs):
...
+                )
+                geometry_sets.append(geometry_set_wrapper)
+
+        # Reset the previously set indices.
+        for i, node in enumerate(self.nodes):
+            node.i_global = None
+        for element in self.elements:
+            element.i_global = None
+
+        # TODO: Let's sort the cells by their block ID here - in most solvers, the
```
davidrudlstorfer, line 917: Another TODO

src/beamme/core/mesh_representation.py
```diff
@@ -203,13 +204,225 @@ def _store_geometry_set_data(
...
+
+    # Create merged data entries
+    #   Points
+    merged_points = _np.vstack(
+        (mesh_representation_a.points, mesh_representation_b.points)
+    )
+
+    #   Connectivity - before the connectivity can be merged, the IDs of the second mesh have
+    #   to be increased by the size of the first mesh.
+    #   First, we create a mask to identify the entries that have to be incremented.
```
davidrudlstorfer: Can we again clean the comments here to be consistent

src/beamme/core/vtk_writer.py
```diff
@@ -107,7 +107,9 @@ def add_point_data_node_sets(point_data, nodes, *, extra_points=0):
             raise TypeError("The geometry type is wrong!")
 
         # Add the data vector.
-        set_name = f"{geometry_name}_set_{_bme.vtk_node_set_format.format(geometry_set.i_global + 1)}"
+        set_name = (
+            f"{geometry_name}_set_{_bme.vtk_node_set_format.format(i_geometry_set + 1)}"
```
davidrudlstorfer, line 111: Would we now be able to drop those +1 indices? I think we talked about it some time ago

src/beamme/four_c/input_file.py
```diff
@@ -68,15 +59,16 @@ def __init__(self):
 
         self.fourc_input = _FourCInput()
 
-        # Contents of NOX xml file.
-        self.nox_xml_contents = ""
-
         # Register converters to directly convert non-primitive types
         # to native Python types via the FourCIPP type converter.
         self.fourc_input.type_converter.register_numpy_types()
-        self.fourc_input.type_converter.register_type(
```
davidrudlstorfer, line 77: Could you maybe briefly explain why we do not need it at all anymore? Would this also directly close #456

src/beamme/four_c/input_file.py
```diff
@@ -68,15 +59,16 @@ def __init__(self):
...
-            (_Function, _Material, _Node), lambda converter, obj: obj.i_global + 1
-        )
+
+        # Contents of NOX xml file.
+        self.nox_xml_contents = ""
+
+        # Mesh representation for this input file.
+        self.mesh_representation = _MeshRepresentation()
```
davidrudlstorfer, line 70: Just to understand, so every input file now automatically contains the mesh_representation at all times?

src/beamme/four_c/input_file.py
```diff
@@ -165,7 +157,7 @@ def add(self, object_to_add, **kwargs):
         """
 
         if isinstance(object_to_add, _Mesh):
-            self.add_mesh_to_input_file(mesh=object_to_add, **kwargs)
+            _dump_mesh_to_input_file(self, mesh=object_to_add, **kwargs)
```
davidrudlstorfer, line 160: Why did this function move out of this class. Do we need somewhere else?

src/beamme/four_c/input_file.py
```diff
@@ -215,21 +207,32 @@ def dump(
         # Make sure the given input file is a Path instance.
         input_file_path = _Path(input_file_path)
 
+        # Create a deep copy of the existing input sections - this function does not alter
+        # the present instance of InputFile
+        fourc_input = self.fourc_input.copy()
```
davidrudlstorfer, line 223: Why do we need this now?

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -146,32 +97,10 @@ def dump_coupling(coupling):
...
-                geometry_set.geometry_type
-            ],
-            "d_id": geometry_set.i_global + 1,
-        }
-        for node in nodes
-    ]
-
-
-def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
+def dump_nurbs_patch_knotvectors(input_file, nurbs_patch: _NURBSPatch) -> None:
```
davidrudlstorfer, line 103: thanks for adding more type hints

src/beamme/four_c/input_file_dump_item.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
 
 
-def dump_nurbs_patch_elements(nurbs_patch: _NURBSPatch) -> list[dict[str, _Any]]:
-    """Return a list with all the element definitions contained in this
-    patch."""
+def dump_item(item) -> dict[str, _Any]:
+    """General function to dump items to a 4C input file."""
+    if hasattr(item, "dump_to_list"):
+        return item.dump_to_list()
+    elif isinstance(item, _BoundaryCondition):
```
davidrudlstorfer: Wouldn't it make sense to just add the dump function to boundary conditions (and couplings down below) as well? Then it would be unified and we can completely delete this function

src/beamme/four_c/input_file_dump_item.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
-    if nurbs_patch.i_global is not None:
-        raise ValueError(
-            f"For NURBS patches, i_global shall not be set, got {nurbs_patch.i_global}."
-        )
-    if nurbs_patch.i_nurbs_patch is None:
-        raise ValueError("Expected i_nurbs_patch to be set for NURBS patch.")
+    Internally, we store the geometry information from the mesh in the mesh
+    representation of the input file. All other information, e.g., element
+    types, materials, boundary conditions and functions will be dumped to the
+    FourCIPP input file.
```
davidrudlstorfer:
```suggestion
    4C input file via FourCIPP.
```

src/beamme/four_c/input_file_dump_item.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
+
+    Args:
+        input_file: The input file where we want to add the mesh information.
+        mesh: The mesh to be added to the input file.
+    """
 
-    # Check the material
-    nurbs_patch._check_material()
+    # Compute starting indices
+    #   Element types
```
davidrudlstorfer: Formatting

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
+        _Function, lambda _, obj: function_to_i_global[obj] + 1
+    )
+    input_file.fourc_input.type_converter.register_type(
+        _Material,
+        lambda _, obj: material_to_i_global[obj] + 1 + start_index_materials,
+    )
+    input_file.fourc_input.type_converter.register_type(
+        _NURBSPatch,
+        lambda _, obj: nurbs_patch_to_i_global[obj] + 1 + start_index_nurbs_patches,
+    )
```
davidrudlstorfer, line 254: This looks a little bit hacky and tbh I do not fully understand it. Could you maybe explain?

src/beamme/four_c/input_file_dump_item.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
+        full_item_list.extend(dumped)
+        input_file.add({section_name: full_item_list})
+
+    # Dump materials
+    _dump("MATERIALS", material_to_i_global.keys())
+
+    # Dump couplings
+    #   If there are couplings in the mesh, set the link between the nodes
+    #   and elements, so the couplings can decide which DOFs they couple,
+    #   depending on the type of the connected beam element.
```
davidrudlstorfer: Formatting

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
-        patch_elements.append(text_based_input_dict)
-        j += 1
+    # Dump NURBS patch information.
+    for element in mesh.elements:
+        if isinstance(element, _NURBSPatch):
+            dump_nurbs_patch_knotvectors(input_file, element)
 
-    return patch_elements
 
+def dump_mesh_representation_to_input_file_legacy(
```
davidrudlstorfer, line 323: Same as last time, would text based or something similar make sense here? Legacy is not really connected to the text based input file

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer, line 1: Probably it would also make sense to rename this file, we are no longer dumping items but whole meshes + conversions + ...

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -217,70 +146,357 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
+            # We modify the dict to add the rotation vectors, thus we create a
+            # deep copy here to avoid modifying the original data structure.
+            four_c_data = _copy.deepcopy(four_c_data)
+            # The numpy quaternion package can return rotation vectors outside
+            # the range of -pi to pi, which can cause issues in testing comparisons.
+            # To avoid this, we convert the rotations to BeamMe internal rotations
+            # and extract the rotation vectors again, ensuring they are in the
+            # correct range.
+            # This is super slow, but for now keep it, as a mesh based output is to
+            # be preferred when performance is of importance.
```
davidrudlstorfer, line 433: Why do we not need this with the mesh based input? I guess we also want definite rotations?

tests/integration/test_integration_core_vtk_writer.py
```diff
@@ -140,7 +140,7 @@ def test_integration_core_vtk_writer_beam(
         binary=False,
         coupling_sets=True,
     )
-    assert_results_close(ref_file, vtk_file)
+    # assert_results_close(ref_file, vtk_file)
```
davidrudlstorfer, line 144: What's up with this? If it should stay a ToDo would make sense

isteinbrecher, line 144: @davidrudlstorfer with the restructuring, the vtk tests fail because the order of the geometry sets is different now also for the vtk output - however, since I want to change the way we write output to simply use the mesh representation, I decided not to fix the implementation and deactivate the result check of the tests. This does not change anything about the `display_pyvista` functionality.

tests/conftest_result_comparison.py
```diff
@@ -183,7 +186,12 @@ def convert_to_primitive_type(
         obj = input_file
 
     if isinstance(obj, InputFile):
-        return obj.sections
+        # Create a copy of the FourCIPP input file and dump the mesh representation to it.
+        fourc_input_file = obj.fourc_input.copy()
+        dump_mesh_representation_to_input_file_legacy(
+            fourc_input_file, obj.mesh_representation, obj.element_type_id_to_data
+        )
```
davidrudlstorfer: i do not really like this approach that we now need, i.e., that we need an external function call to convert the mesh representation to a "dumpable" input file

couldn't we just use some boolean parameter that does that automatically? For example if i call input_file.sections I would probably automatically expect the input file in the full text based format?

Also in the dump function I would probably just add a parameter like "dump_type" which can either be "text_based" or "binary" and depending on that everything is done automatically

It's a little bit intuitive that you now need external functions imported from another file to convert the mesh_representation

davidrudlstorfer: Especially because the argument is only consisting of input files if I see it correctly?

fourc_input file is an input file copied (why actually the copy)?

and then all arguments of the dump_mesh_representation_to_input_file_legacy are input files again.

In my mind this should simply be

```python
input_file.convert_mesh_repr_to_text_based_input()
```

Without any necessary arguments or whatsoever

davidrudlstorfer (commented):
@isteinbrecher that review took a little bit longer based on the really large amount of changes, I started yesterday and was able to finish with the review today :D

Overall this is already looking really amazing, a lot of points are just coding style or formatting. The only thing I do not really like currently is the "manual"/"external" function that needs to be called to convert the mesh representation of an input file (see below).

src/beamme/core/element.py
isteinbrecher: I think it is not really needed, this can be removed. I will add this to this PR.

src/beamme/core/mesh.py
isteinbrecher: This is actually a TODO for the future, to really evaluate if this flag can be removed, we first need to change the `vtk` output.

src/beamme/core/mesh.py
isteinbrecher: YES. That has actually been really bothering me with this PR, the different names we use for global IDs. I think I would prefer `gid` and try to change the ones where it is possible to be index-0 based. I am not sure about materials, but I think functions can not start with 0, because 0 indicates a constant value in 4C.

I would do this in a follow up PR.

src/beamme/core/mesh.py
isteinbrecher: This can happen if the user explicitly adds a material to the `mesh.materials` list more than once.

In the past this caused an inconsistency in the input file, because we assigned the GIDs to the material object and then we would have assigned a GID to the same material twice.

With the new way we assign material GIDs (`material_to_i_global`), this can't happen any more, so we actually could remove the error. We could also keep it, as it still indicates a state of the mesh where it likely should not be.

src/beamme/core/mesh.py
isteinbrecher: I intentionally put the comments like these, because they are sort of "lower level" comments to the preceding
```python
# Extract elements for the mesh representation
```
I personally like having this kind of comments as it adds some structure to the code (in python we don't have `{...}` as in C++). Do you have any other ideas how to achieve that?

src/beamme/core/mesh.py
isteinbrecher: We could simply append the data. However, we later on have to convert it to a numpy array anyways, so I think it makes sense to already allocate the memory here. I haven't looked at this from a timing point of view (lists in python are surprisingly fast), but this should be the cleanest approach from a memory point of view.

src/beamme/core/mesh.py
isteinbrecher: Rigid sphere elements in 4C don't need a material. I adapted the comment accordingly.

src/beamme/core/mesh.py
isteinbrecher: This is needed to satisfy mypy. I adapted the comment.

src/beamme/core/mesh.py
isteinbrecher: This one is also for the future.

src/beamme/core/vtk_writer.py
isteinbrecher: Yes, see my previous comment. I think we would have no issues with geometry sets being index 0 based.

isteinbrecher: Only thing is, that that will break all input files for testing again 😅

src/beamme/four_c/input_file.py
isteinbrecher: We still need the type converter, but in a different way than before. Previously, we set GIDs (`self.i_global`) for objects and then returned that index when dumping.

I never really liked that approach of adding the temporary `self.i_global` variable - now this is changed, that when dumping materials, functions and geometry sets to an input file, we don't add the GIDs to the objects, but rather create a map from the objects to their GIDs.

Later on in the code, we set the type converter to return the GIDs from the respective mappings.

Note: We still set the GIDs in the nodes and elements. We can think about also replacing this with mappings, but for that I would like to have a look at the performance - which I think is a lot easier in a standalone add-on PR.

src/beamme/four_c/input_file.py
isteinbrecher: Yes. Maybe I should have put this in the description.

The mesh representation is an abstract data structure that holds all geometry information of the input/file (except the one that is not converted from legacy input). In the future, we want to be able to `dump` to either the legacy format or some external mesh format. To do this in an efficient way, we can not decide on the export format until the `dump`, thus we use the mesh representation.

src/beamme/four_c/input_file.py
isteinbrecher: No, we don't need it somewhere else. I would actually prefer this here, but thought it might also make sense in the `input_file_dump_item.py` file - but I can move it back if we want.

src/beamme/four_c/input_file.py
isteinbrecher: Related to the previous comment on the mesh representation.

An input file now stores the input data (in the FourCIPP object) and the geometry information in the mesh representation (plus one map defining the element types).

Let't take the example where we want to dump to legacy format. We will decide this when calling `dump`. Then we could simply add all the legacy strings to the input file and write it to disk. However, that would then alter the input file object, i.e., the mesh representation is empty, because everything is dumped to the input file sections.

I thought about this for a long time and think, the proposed way is the one that makes the most sense. From a performance point of view, once we can read a mesh representation from external mesh formats, `fourc_input` should always be relatively small, so the `copy` comes at little cost.

Also, from a general point of view, I like the concept of `dump` not modifying the input file. All it should do is dump the contained information to a file.

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: See the previous comment on the type converter.

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: Is this not the way we call the dat-style strings in an inout file? I know we talked about this recently, but after revisiting this here, I think legacy actually makes more sense (though it might not be 100% technically correct).

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: You are right, this might lead to non-unique rotation vector in an external mesh file. However, in 4C this does not matter, since we convert it to quaternions anyways.

The question that remains is how we handle this in testing. There I would maybe simply add a conversion in the future that compares that array in a $SO(3)$ metric, where non-unique rotation vectors won't make a difference. (At least that is the plan, lets see if that works as expected)

tests/integration/test_integration_core_vtk_writer.py
isteinbrecher: I will add a todo that this is to be removed in the future.

tests/conftest_result_comparison.py
isteinbrecher: Also related to the previous comments. Up until now we could do the comparison by simply looking at `input_file.sections`. That worked, because when we added a mesh to the input file we dumped all mesh information to `input_file.sections`. So at this point all the information was already in the `input_file.sections`.

Now, we have `input_file.sections` and `input_file.mesh_representation` and want to compare this to input files on disk. So we have to add the mesh information to the `input_file.sections` here somehow so we can compare it.

What we could do is add a method
```python
def InputFile():
    ...
    def get_copy_with_legacy_mesh_sections(self) -> FourCInput:
        fourc_input_file_with_mesh = self.fourc_input.copy()
        dump_mesh_representation_to_input_file_legacy(
            fourc_input_file_with_mesh, self.mesh_representation, self.element_type_id_to_data
        )
        return fourc_input_file_with_mesh
```
and then call this method here to get the input file sections we need for testing comparisons.

isteinbrecher:
@davidrudlstorfer thanks for the thorough review! Feel free to close resolved discussions. And sorry for a maybe too short description for such a long PR.

src/beamme/four_c/input_file_dump_item.py
isteinbrecher: I think that would not work as expected, because we would then have to create 4C related classes for BC and coupling that derive from the base classes and add the `dump_to_list` function.

For materials we can do this, since we already derive 4C based materials.

This function only has one caller, so what I will do, is move this function to where it is actually called and we don't have this explicit function any more.

src/beamme/core/mesh.py
davidrudlstorfer: Should we open up an issue for that?

src/beamme/core/mesh.py
davidrudlstorfer: If it is not even able to happen anymore in a meaningful way I would remove it, it's more or less dead code then

src/beamme/core/mesh.py
davidrudlstorfer: This would still be an open point for me


```suggestion
        nurbs_control_point_weights = None
```

src/beamme/core/mesh.py
davidrudlstorfer: I get your point but I would still aim for a consistent commenting style. I do not really know how these comment play along with our formaters tbh.

When one reads the code from the top to bottom it is clear anyways that these parts fit together.

If a separation is really necessary it would probably make sense to put that stuff into its own function? Then it can be documented nicely and all the comments belong to this one "task"

Another option I would see is to start the first comment with some sort of headline/step

Something like 

`# Step 1: Extract elements for the mesh representation`

or 

`# First step; extract elements for the mesh representation`

and so on for the next steps.

Looking at the function being something like 230 lines long, some inner functions would probably make this even easier to read and differentiate

src/beamme/core/mesh.py
davidrudlstorfer: I still think the appending and then converting it to numpy where necessary is nicer. We have this approach of pre-allocating the arrays nowhere in our code base and we should stay consistent. There's also no performance benefit in doing so. In my opinion this only makes the code harder to read.

src/beamme/core/mesh_representation.py
davidrudlstorfer: Same as in the other comment, here there are already some "sub functions" so it would maybe also make sense for this code block

src/beamme/core/vtk_writer.py
davidrudlstorfer: So this will also be tackled with #634? I would have absolutely no problem with breaking all test files

src/beamme/four_c/input_file.py
davidrudlstorfer: On a second notice it makes sense in the other file so let's keep it :)

src/beamme/four_c/input_file.py
davidrudlstorfer: Hmmm I am still not sold on this approach

Cannot we somehow combine this a little bit more elegantly. I would have the following suggestion but I do not really yet know if it makes sense or not

The input file contains the following data

- input file sections (general 4C input stuff which is contained in the fourcipp object)
- mesh representation

Could we add maybe a third data structure

- mesh representation to text based string which is just a second fourcipp object?

With every dump we just overwrite the old mesh_representation->text_based_input structure and in the last step we merge the two fourcipp objects just for the dump (the functionality is already in fourcipp)

This way we do not need any copy stuff at all + we keep the mesh representation alive at all times. This way we do not loose anything and have the full flexibility

src/beamme/four_c/input_file_dump_item.py
davidrudlstorfer: same applies here with different sub functions

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: If it's fine for you we could add a todo here or open a new issue or add it to #456. I am not fully sold on this solution. Maybe we find something better after a brief talk about this

src/beamme/four_c/input_file_dump_item.py
davidrudlstorfer: Same as all above

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: So legacy are only the parts in 4C which are not converted to yaml style input functionality, such as the elements or nodes

But as far as I understand the mesh representation also holds stuff like functions or materials. So I think `dump_mesh_representation_to_text_input_file` or even `dump_mesh_representation_to_yaml` is easier to understand

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: Also this still applies

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: Should we also add this as a todo/issue?

tests/conftest_result_comparison.py
davidrudlstorfer: I would prefer not getting copies of data because this get's imho dirty quickly in python. I could still imagine some third data type in the input file

input_file.`sections_from_mesh_representation` which we fill with `input_file.mesh_representation_to_sections`. During the dump we can simply merge the two in a temporary fourcipp input file and write it to disk

davidrudlstorfer (commented):
I hope I have answered everything, please just ping me if I have forgotten any comment :)

This already is pretty close, I would love to resolve the comment formatting somehow and the copies of the input file. Maybe we find a good solution on both :)

src/beamme/core/mesh.py
isteinbrecher: I think this should be already covered in https://github.com/beamme-py/beamme/issues/634 - or did in miss anything?

src/beamme/core/mesh.py
isteinbrecher: Then let's remove it!

src/beamme/core/mesh.py
isteinbrecher: I am a little bit torn here, since we call them "control points" every where else in the code and 4C.

Somewhat related, I don't quite like that we now gather all different node types here with their specific data, e.g., rotation vectors for Cosserat nodes, weights for control points, arc length and time for space time nodes (in the future). I am thinking about splitting this up in the future - then I think also the naming won't be an issue any more - but I haven't come up with the right idea yet. I would like to look at these thins in follow up PRs.

isteinbrecher: I added a comment in #594 https://github.com/beamme-py/beamme/issues/594#issuecomment-4353840223 to keep track of these follow ups.

src/beamme/core/mesh.py
isteinbrecher: I had a look at a simple test script comparing preallocating with python lists for the global coordinates (kudos ChatGPT) and the runtime for python lists is about twice as long (that is actually not the big issue) but the memory consumption is about 10x as much. So I would keep this.

Just a side note, there are already a few places in BeamMe where we preallocate memory, e.g., `get_nodal_coordinates`, `get_nodal_quaternions`.

<details>
<summary>Allocation benchmark</summary>

```python
import time
import tracemalloc

import numpy as np


def generate_point(i):
    return float(i), float(i + 1), float(i + 2)


def list_then_numpy(n):
    points = []
    for i in range(n):
        points.append(generate_point(i))
    return np.array(points, dtype=np.float64)


def preallocated_numpy(n):
    points = np.empty((n, 3), dtype=np.float64)
    for i in range(n):
        x, y, z = generate_point(i)
        points[i, 0] = x
        points[i, 1] = y
        points[i, 2] = z
    return points


def benchmark(func, n):
    tracemalloc.start()
    t0 = time.perf_counter()
    arr = func(n)
    t1 = time.perf_counter()
    _, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    return t1 - t0, peak / 1e6, arr  # time (s), peak memory (MB)


if __name__ == "__main__":
    N = 1_000_000  # try 100k, 1M, 5M depending on RAM

    print(f"Benchmark with N = {N}\n")

    t_list, mem_list, arr1 = benchmark(list_then_numpy, N)
    print(f"List -> NumPy:")
    print(f"  Time: {t_list:.3f} s")
    print(f"  Peak memory: {mem_list:.1f} MB\n")

    t_pre, mem_pre, arr2 = benchmark(preallocated_numpy, N)
    print(f"Preallocated NumPy:")
    print(f"  Time: {t_pre:.3f} s")
    print(f"  Peak memory: {mem_pre:.1f} MB\n")

    # sanity check
    print("Arrays equal:", np.allclose(arr1, arr2))
```
<\details>

src/beamme/core/vtk_writer.py
isteinbrecher: Yes, this sill be part of #634.

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: Good point, I added a bulletpoint to https://github.com/beamme-py/beamme/issues/456 (and a link in #634)

src/beamme/core/mesh.py
isteinbrecher: You have a good point, this likely means that we should split this up and or use sub functions. Alternatively, we could simply highlight the "higher level" comment with `##`. I will have a look.

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: I added a todo in #594

src/beamme/four_c/input_file.py
isteinbrecher: > mesh representation to text based string which is just a second fourcipp object?

This part I don't fully understand. Why would we want to store the legacy/text mesh format at all? The contained information would be equivalent to the mesh representation, but the mesh representation is much more flexible and efficient. The only time we want to convert a mesh representation to legacy/text input is immediately before we want to write it to disk in legacy/text format. After the information is dumped, we don't need it any more.

---

The rest of what you are describing is more or less what is already happening in the code. Conceptually, these two workflows are more or less equal:
- mesh to data (`mesh_data`) -> merge `mesh_data` with `self.fourc_input` and directly dump to disk
- copy `self.fourc_input` (`copy`) -> mesh to data (`mesh_data`) -> add `mesh_data` to `copy ` -> dump `copy`

With the merge approach, I have to create temporary data structures where I add mesh information and whatever else (e.g., header, ...). Then this temporary data is merged with the existing FourCIPP data and directly dumped to disk.

With the copy variant, I take a copy of the FourCIPP data in the input file. In most cases that should be just parameters, BC, materials, functions, ... but no mesh information, so the copy is efficient. Then I add the mesh information and whatever else (e.g., header, ...) to the copy. In the end I dump the copy to disk and the copy will be destroyed once it goes out of scope.

I prefer the copy variant because it is much clearer what is happening. I am never running into the risk of accidentally modifying data I don't want to modify. 

---

So both approaches are very similar - with one exception:

Looking into the future, we want to output the mesh representation as an external data format (that is the whole reason for all of this 😅). In 4C the way we have to define BC for external meshes is slightly different from text/legacy based input files. So we will likely have to adapt the BC data when dumping depending on the choice of mesh format. With the copy variant this is no issue at all, with the merge approach this becomes more cumbersome.

---

Side note: We actually have a bug/inconsistency when we currently dump an input file:
```python
input_file = InputFile()

# Add stuff to input_file

input_file.dump("dump1.4C.yaml") # Dumps with header
input_file.dump("dump2.4C.yaml", add_header_information=False) # Dumps also with header
```
The first dump adds the header information to the FourCIPP data and then it is already there for the second dump, even though the second dump does not want the header.

With the proposed approach this is fixed directly.

tests/conftest_result_comparison.py
isteinbrecher: See my (way too long) comment from above: https://github.com/beamme-py/beamme/pull/633#discussion_r3169480338

src/beamme/four_c/input_file.py
davidrudlstorfer: Thanks for your long explanation, your proposed idea actually makes a lot more sense 👌

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer, line 1: Should we maybe update the filename to describe the contents better?

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -217,70 +146,352 @@ def dump_nurbs_patch_knotvectors(input_file, nurbs_patch) -> None:
...
+    for element in mesh.elements:
+        if isinstance(element, _NURBSPatch):
+            dump_nurbs_patch_knotvectors(input_file, element)
 
-    # Check the material
-    nurbs_patch._check_material()
 
-    patch_elements = []
-    j = 0
+def dump_mesh_representation_to_input_file_legacy(
```
davidrudlstorfer, line 323: And this also still applies in my opinion 

=> Old comment copied here again for new thread

> So legacy are only the parts in 4C which are not converted to yaml style input functionality, such as the elements or nodes
> 
> But as far as I understand the mesh representation also holds stuff like functions or materials. So I think dump_mesh_representation_to_text_input_file or even dump_mesh_representation_to_yaml is easier to understand

tests/conftest_result_comparison.py
```diff
@@ -183,7 +186,12 @@ def convert_to_primitive_type(
         obj = input_file
 
     if isinstance(obj, InputFile):
-        return obj.sections
+        # Create a copy of the 4C input file via FourCIPP and dump the mesh representation to it.
+        fourc_input_file = obj.fourc_input.copy()
+        dump_mesh_representation_to_input_file_legacy(
+            fourc_input_file, obj.mesh_representation, obj.element_type_id_to_data
+        )
```
davidrudlstorfer: Maybe we could provide the function you already proposed above to simplify this procedure, I think this could be needed in other situations as well

davidrudlstorfer (commented):
So the review comments are slowly getting a real mess with all the threads, comments etc :D

For the last three points I started completely new comments, then we retain the overview. I hope we can merge this large PR today :)

One last question:

The performance test fails with substantial performance degradation, dumping into the text based format now takes close to 3 times as long, have you profiled that in some sort? The overhead seems too large to me and I still think it would be nice if we could keep the performance roughly the same.

Thanks a lot for the huge changes in this PR:)

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: Yes. What about just `input_file_dump.py` or `input_file_dump_functions.py`?

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: Sorry, I missed that one.

> So legacy are only the parts in 4C which are not converted to yaml style input functionality, such as the elements or nodes

In the function `dump_mesh_representation_to_input_file_legacy`, we dump data to the following sections:
- `"NODE COORDS"`
- `"STRUCTURE ELEMENTS"`
- `"DNODE-NODE TOPOLOGY"`
- `"DLINE-NODE TOPOLOGY"`
- `"DSURF-NODE TOPOLOGY"`
- `"DVOL-NODE TOPOLOGY"`

All of them are contained in the list of legacy sections of FourCIPP (`FourCInput().legacy_sections_names`). That's why I originally chose the name legacy - and I still think that is a suitable name, because it basically refers to any mesh data that is contained directly in the input file.

But maybe I am missing something.

What I saw while looking in to this is that there are some variables, e.g., `node_set_internal_id_to_legacy_id`, that have to be renamed since there legacy is defiantly the wrong name.

tests/conftest_result_comparison.py
isteinbrecher: I added a method `get_fourcipp_input_with_mesh` to the input file. This is now called in testing.

isteinbrecher:
Thanks for the thorough review @davidrudlstorfer 

> The performance test fails with substantial performance degradation, dumping into the text based format now takes close to 3 times as long, have you profiled that in some sort? The overhead seems too large to me and I still think it would be nice if we could keep the performance roughly the same.

Yes that is the case. The "degradation" comes from the fact that I moved a lot of the stuff that was previously done during `add` to `dump` (e.g., rotation vector conversion, type converter). Now the `add` is much faster, but the `dump` much slower.
```
BeamMe: Add large beam mesh to input file                              0.474s < 10.000s
BeamMe: Dump input file with large beam mesh                           13.031s > 5.500s
```
I did not profile it, but I assume that the total runtime (`add`+`dump`) is still increased, but only by a small margin. I am fine with this, since for performance critical scripts, we will be able to dump directly to external mesh formats (vtu) in the future, which should be very fast compared to the current `dump`.
Once all of the refactoring of the output formats is done, I want to refactor the performance tests as well. 

---

I think this is close to getting merged, we only have to agree on a naming convention for the legacy/text-based input files and the "dump" file. Then I will add the last changes.

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: `input_file_dump_functions.py`

is a really good proposal :)

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: Oh sorry then I missed something here; I thought functions and materials also get dumped in here. Sorry for the faulty comment.

> What I saw while looking in to this is that there are some variables, e.g., node_set_internal_id_to_legacy_id, that have to be renamed since there legacy is defiantly the wrong name.

Should this be added to some follow-up?

tests/conftest_result_comparison.py
davidrudlstorfer: Thanks :)

src/beamme/four_c/input_file_dump_functions.py
davidrudlstorfer: Upsi - already done - thanks a lot!

davidrudlstorfer:
@isteinbrecher thanks a lot for addressing everything :)

Sorry again for not checking the performance degradation in more detail - with the times just being shifted to some other function it's totally fine.

Let's

- rename the one file as discussed above
- and my last question - does it make sense to directly adjust the performance test times here? Then the pipeline does not always fail.

Then we can merge this ASAP

isteinbrecher:
> - rename the one file as discussed above

Done!

> - and my last question - does it make sense to directly adjust the performance test times here? Then the pipeline does not always fail.

Yes it does, I adapted the times
```
BeamMe: Add large beam mesh to input file              10.000s -> 0.6s
BeamMe: Dump input file with large beam mesh           5.500s -> 15.0s
```
the performance tests still might fail, in the last days I have noticed the test `Find close nodes in large beam mesh` fail repeatedly, maybe that has to do with some change in `scipy` but I didn't have time to look into this.

So everything should be ready now. I rebased everything on the current `main` and tries to squash some commits where possible.

Thanks a lot for the review!

davidrudlstorfer (approved):
Thanks a lot for all the changes and your patience, sorry that it took so long.

Let's get this merged 🚀

PR #640, opened 2026-05-26
Refactor performance tests

isteinbrecher:
This PR refactors the performance tests. Currently, we create testing objects in tests that are then used again in later tests. This relies on the fact that the performance tests are run in the correct order. This works fine if all tests are run, but if only selected performance tests are run this might fail.

The new approach creates commonly used objects as fixtures on module level. If a fixture is required, it is evaluated one per module. For example if two tests require the large beam mesh fixture, it is only evaluated once and then given to the two tests. This eliminates the need for the correct ordering of the performance tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -132,9 +132,13 @@ def pytest_collection_modifyitems(config: Config, items: list) -> None:
         if not markers and not config.getoption("--exclude-standard-tests"):
             selected_tests.append(item)
 
-    deselected_tests = list(set(items) - set(selected_tests))
+    selected_tests_unique = list(
+        dict.fromkeys(selected_tests).keys()
+    )  # remove duplicates while preserving order
 
-    items[:] = selected_tests
+    deselected_tests = list(set(items) - set(selected_tests_unique))
+
+    items[:] = selected_tests_unique
     config.hook.pytest_deselected(items=deselected_tests)
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -110,26 +110,32 @@ def pytest_collection_modifyitems(config: Config, items: list) -> None:
         items: Pytest list of tests
     """
 
+    # Get all active markers for this pytest run
+    active_markers = set()
+    for flag, marker in zip(
+        ["--4C", "--ArborX", "--CubitPy", "--performance-tests"],
+        ["fourc", "arborx", "cubitpy", "performance"],
+    ):
+        if config.getoption(flag):
+            active_markers.add(marker)
+
     selected_tests = []
 
     # loop over all collected tests
     for item in items:
-        # Get all set markers for current test (e.g. `fourc_arborx`, `cubitpy`, `performance`, ...)
+        # Get all set markers for current test (e.g. `4C`, `ArborX`, `CubitPy`, ...)
         # We don't care about the "parametrize" marker here
-        markers = [
+        markers = set(
             marker.name
             for marker in item.iter_markers()
             if not marker.name == "parametrize"
-        ]
+        )
 
-        for flag, marker in zip(
-            ["--4C", "--ArborX", "--CubitPy", "--performance-tests"],
-            ["fourc", "arborx", "cubitpy", "performance"],
-        ):
-            if config.getoption(flag) and marker in markers:
+        if markers:
+            if markers.issubset(active_markers):
                 selected_tests.append(item)
 
-        if not markers and not config.getoption("--exclude-standard-tests"):
+        elif not config.getoption("--exclude-standard-tests"):
             selected_tests.append(item)
 
     selected_tests_unique = list(
--- a/tests/performance/test_performance_beamme.py
+++ b/tests/performance/test_performance_beamme.py
@@ -174,61 +174,78 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
     return mesh
 
 
-@pytest.mark.cubitpy
-@pytest.mark.performance
-def test_performance_beamme_cubitpy_create_solid(
-    evaluate_execution_time, shared_tmp_path
-):
-    """Test the performance of creating a solid block using CubitPy.
+@pytest.fixture(scope="module")
+def medium_solid_block(shared_tmp_path, evaluate_execution_time):
+    """Provide a solid mesh from cubit.
 
     The version of Cubit we use in testing only allows for 50,000, so we
     create a mesh with exactly that.
     """
-
     cubit = CubitPy()
-
+    input_file_path = shared_tmp_path / "performance_testing_solid_half.4C.yaml"
     evaluate_execution_time(
         "CubitPy: Create half solid block",
         create_solid_block,
         kwargs={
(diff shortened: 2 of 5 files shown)


tests/performance/test_performance_beamme.py
```diff
@@ -174,60 +174,74 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
             "ny": 50,
             "nz": 10,
         },
         expected_time=3.8,
     )
+    return input_file_path
 
 
+@pytest.mark.cubitpy
 @pytest.mark.performance
```
isteinbrecher, line 202: @davidrudlstorfer how do we want to handle the case where we have multiple `mark`s? I would say that all marks need to be activated for a test to run, i.e., for this case to run pytest would have to be called with both `--CubitPy --performance-tests`. What do you think?

davidrudlstorfer: yes that makes the most sense :)

is that already the current behaviour?

tests/performance/test_performance_beamme.py
```diff
@@ -174,60 +174,76 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
     return mesh
 
 
-@pytest.mark.performance
-def test_performance_beamme_cubitpy_create_solid(
-    evaluate_execution_time, shared_tmp_path
-):
-    """Test the performance of creating a solid block using CubitPy.
+@pytest.fixture(scope="module")
```
davidrudlstorfer, line 177: is this actually working? back then i had the problem that it was super slow with the scope (or something else 🧐)

davidrudlstorfer: looking at #369 the original problem was that it needed to be executed twice. Did you see this behaviour again?

tests/performance/test_performance_beamme.py
```diff
@@ -174,60 +174,76 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
+@pytest.mark.cubitpy
 @pytest.mark.performance
-def test_performance_beamme_double_solid_block(
-    evaluate_execution_time, shared_tmp_path
-):
+def test_performance_beamme_cubitpy_create_solid(medium_solid_block):
+    """Test the performance of creating a solid block using CubitPy.
+
+    The test is run in the fixture, so we don't need to do anything
+    here.
```
davidrudlstorfer, line 207: I am not fully getting this, could you explain this?

davidrudlstorfer: ah

would it make sense to add a `pass` or whatever to this function? The empty function is a bit strange

also I am not a super big fan of this behaviour tbh

tests/performance/test_performance_beamme.py
```diff
@@ -174,60 +174,76 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
     )
 
 
+@pytest.mark.cubitpy
+@pytest.mark.performance
+def test_performance_beamme_double_solid_block(large_solid_block):
+    """Test the performance of doubling a solid block from an input file.
+
+    The test is run in the fixture, so we don't need to do anything
+    here.
```
davidrudlstorfer, line 243: Same here

tests/conftest.py
davidrudlstorfer, line 1: I am not fully getting the changes in this file, what was the reason for that?

davidrudlstorfer (commented):
So I am not fully getting behind this change, what is the actual reason that we want this now?

The empty functions which are just called in the fixture are not really nice tbh

tests/performance/test_performance_beamme.py
isteinbrecher: No, I didn't see that behaviour. The fixture is run the first time it is requested and then the result is stored internally by pytest and returned the next time the fixture is called.

tests/performance/test_performance_beamme.py
isteinbrecher: Currently we don't have a "behaviour", because pytest fails if we have more than one mark on a test. This has to do with how we collect the tests in `pytest_collection_modifyitems`. At the moment, we would execute a test twice if it has two marks (and pytest is called with both flags), which can cause issues with pytest. The changes to `pytest_collection_modifyitems` now ensure that we get the expected behaviour (as described above).

tests/conftest.py
isteinbrecher: The changes basically enable the behaviour discussed in https://github.com/beamme-py/beamme/pull/640/changes#r3310894638

tests/performance/test_performance_beamme.py
isteinbrecher: Yes, this is indeed a little weird.

For example the fixture `large_beam_mesh` creates a large beam mesh that can then be reused in different performance tests (without it having to be recreated each time). However, we also want to time the creation of `large_beam_mesh` itself. We could do that by creating large beam mesh here again, but then we have to do that twice. I opted for the approach that we directly use `evaluate_execution_time` when creating the file in the fixture.

So lets say one just wants to run the performance test `test_performance_beamme_translate` with 
```bash
pytest --performance-tests -k test_performance_beamme_translate
```
This will internally call the fixture to create the large beam mesh (and time the creation) and then perform the translation. The performance test summary will then look like this:
```
==================== Performance Test Summary ======================
BeamMe: Create large beam mesh                                         2.118s < 2.500s
BeamMe: Translate large beam mesh                                      0.060s < 0.250s
```
I think this makes sense, since we already have to create the large beam mesh, then we can also time it.

Now, the empty test is simply there, so one can run just __that__ test, i.e.,
```bash
pytest --performance-tests -k test_performance_beamme_create_beams
```

I hope that helps. Regarding the `pass`, we can add that, I don't really care. The only thing is that some python style guides advice agains using `pass` when it is not required for a correct syntax.

tests/performance/test_performance_beamme.py
davidrudlstorfer: Ah, I didn't get that it is already implemented:)

tests/performance/test_performance_beamme.py
davidrudlstorfer: I would add the pass so we really see that this function is empty, I initially thought that it defined the next function within that function during the review

davidrudlstorfer:
Looks good, thanks for the explanations, let's add the pass and then merge this :)

isteinbrecher:
> Looks good, thanks for the explanations, let's add the pass and then merge this :)

Thanks for the comments, I added `pass` to the selected tests. I think this should be ready to merge.

davidrudlstorfer (approved):
Thanks for addressing the comments :)

Just a final question; the performance tests also passed once right? I did not check the old pipelines now but two tests took too long in the last run (I do not know if that is connected to the code changes)

isteinbrecher:
> Just a final question; the performance tests also passed once right? I did not check the old pipelines now but two tests took too long in the last run (I do not know if that is connected to the code changes)

I don't think so. But if this turns out to be a constant issue from now on we can investigate this.

PR #643, opened 2026-05-29
Add check that tests align with a certain naming convention

isteinbrecher:
This PR adds the functionality that pytest checks if tests align with our naming convention.


tests/conftest.py
```diff
@@ -93,9 +93,38 @@ def pytest_addoption(parser: Parser) -> None:
...
+        raise ValueError(
+            f"The test file {item.path} does not follow our naming convention. "
+            f"Expected to start with {prefix_from_subdirs}"
+        )
+
+    if not test_name.startswith(test_file_name):
+        raise ValueError(
+            f"The test {item.path}::{test_name} does not follow our naming convention."
+            f" Expected to start with {test_file_name}"
+        )
```
davidrudlstorfer, line 120: why do we have two separate error messages here? isn't this basically the same?

davidrudlstorfer (approved):
Nice feature :)

tests/conftest.py
isteinbrecher: The first one is for the test file and the second one is for the actual test, thus the slightly different error messages. I had a look and I would keep this as it is, generalising would lead to more complex code.

davidrudlstorfer: Ah sorry, I did not see that. Thanks for the explanation. Maybe changing the function name would fit then that it not only checks test names but also file names

isteinbrecher: Good point, I was to fast with merging, I opened a follow up PR #645

PR #646, opened 2026-06-01
Unify handling of 4C element data

isteinbrecher:
In 4C we use different data structures for elements in legacy format or external mesh format. This makes a direct unified output cumbersome. This PR adds a class `FourCElementData` that can handle both cases.

### DIFF
--- a/src/beamme/four_c/element_beam.py
+++ b/src/beamme/four_c/element_beam.py
@@ -22,7 +22,6 @@
 """This file implements beam elements for 4C."""
 
 import warnings as _warnings
-from typing import Any as _Any
 
 import numpy as _np
 
@@ -30,6 +29,7 @@ from beamme.core.conf import bme as _bme
 from beamme.core.element_beam import Beam as _Beam
 from beamme.core.element_beam import Beam2 as _Beam2
 from beamme.core.element_beam import generate_beam_class as _generate_beam_class
+from beamme.four_c.element_data import FourCElementData as _FourCElementData
 from beamme.four_c.four_c_types import (
     BeamKirchhoffConstraintType as _BeamKirchhoffConstraintType,
 )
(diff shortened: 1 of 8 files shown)


src/beamme/four_c/element_beam.py
```diff
@@ -86,18 +91,23 @@ def get_four_c_kirchhoff_beam(
...
 
     n_nodes = 3
+
     four_c_type = _INPUT_FILE_MAPPINGS["four_c_type_to_four_c_type"][
         _BeamType.kirchhoff
     ]
     four_c_cell = _INPUT_FILE_MAPPINGS["element_type_and_n_nodes_to_four_c_cell"][
         _bme.element_type.beam, n_nodes
     ]
-
```
davidrudlstorfer, line 95: Are these formatting changes on purpose?

src/beamme/four_c/element_data.py
```diff
@@ -0,0 +1,118 @@
...
+
+import numpy as _np
+from numpy.typing import NDArray as _NDArray
+
+from beamme.utils.data_structures import (
+    compare_nested_dicts_or_lists as _compare_nested_dicts_or_lists,
+)
+
+
+class FourCElementData:
```
davidrudlstorfer, line 37: Would this make sense as a data class?

i.e., 


```suggestion 
@dataclass
class FourCElementData:
```

Then you could remove the `__init__` (and some other benefits)

davidrudlstorfer: (ChatGPT proposes something like

```
@dataclass
class FourCElementData:
    """Class that contains the data for a 4C element block."""

    four_c_type: str
    four_c_cell: str
    element_technology: dict[str, Any] = field(default_factory=dict)
```
)

src/beamme/four_c/element_data.py
```diff
@@ -0,0 +1,118 @@
...
+                    {"MAT": element_material_id + 1}
+                    if element_material_id != -1
+                    else {}
+                ),
+                **self.element_technology,
+                **(additional_element_data or {}),
+            },
+        }
+
+    def __eq__(self, other) -> bool:
```
davidrudlstorfer, line 68: And with the dataclass this is also not necessary anymore

davidrudlstorfer (commented):
Looks good :)

Some ideas for the code

src/beamme/four_c/element_beam.py
isteinbrecher: Yes, because `n_nodes` is used also later on, I thought it would be better not to group it with the next code block.

src/beamme/four_c/element_data.py
isteinbrecher: Yes that would work. I uploaded a commit where this is switched to a `dataclass`. However, we still need a custom compare function anyway, since we use `compare_nested_dicts_or_lists` from FourCIPP in `__eq__`.

davidrudlstorfer (approved):
Thanks for using the data class :)

PR #647, opened 2026-06-01
Fix binder build

isteinbrecher:
The binder build has been failing for some time, this should fix it.


davidrudlstorfer (approved):
Thanks for looking into that!

PR #648, opened 2026-06-02
Fix material modularisation

isteinbrecher:
With #633 we use `MeshRepresentation` as the intermediate data structure between a mesh and an input file.

The changes from #633 import `from beamme.four_c.material` in `core/mesh.py`, which is not ideal with respect to modularisation. This PR fixes that by restructuring where the material ID lookup map is created.

Also the ID handling of materials is changed, now the BeamMe internal material IDs are index-0 based.


davidrudlstorfer (approved):
Thanks for also cleaning the docstrings :)

PR #649, opened 2026-06-04
Allow to write 4C input file with vtu mesh

isteinbrecher:
This PR adds the functionality to write a 4C input file where the mesh is stored in a vtu file.


src/beamme/four_c/element_data.py
```diff
@@ -41,6 +41,29 @@ class FourCElementData:
...
+        self, i_block: int, i_material: int, additional_data: dict | None = None
+    ) -> dict:
+        """Return the dictionary containing the data for this element block.
+
+        Args:
+            i_block: The index of the element block.
+            i_material: The index of the material. If -1, no material will be assigned to this element block.
+            additional_data: Additional data to add to the element block dictionary.
+
+        Returns: The dictionary to write this element block data to the input file for mesh based output.
```
davidrudlstorfer: Maybe add line breaks here

src/beamme/four_c/input_file.py
```diff
@@ -190,6 +195,12 @@ def dump(
         Args:
             input_file_path:
                 Path to the input file that should be created.
+            mesh_format:
+                The format in which the mesh information should be written.
+                Currently, "vtu" and "legacy" are supported.
```
davidrudlstorfer: I am not super sold on legacy, could we use 'yaml' or 'text-based'?

src/beamme/four_c/input_file.py
```diff
@@ -217,13 +228,21 @@ def dump(
         # Make sure the given input file is a Path instance.
         input_file_path = _Path(input_file_path)
 
+        # Base name of the input file without extension.
+        if not input_file_path.name.endswith(".4C.yaml"):
+            raise NameError(
+                "Input file must have a .4C.yaml extension, but got the "
+                f"path {input_file_path}"
+            )
```
davidrudlstorfer, line 236: Nice added check :)

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -484,3 +492,169 @@ def _dump(section_name: str, dictionary_list: _List):
...
+
+def dump_mesh_representation_to_input_file_vtu(
+    fourc_input: _FourCInput,
+    mesh_representation: _MeshRepresentation,
+    element_type_id_to_data: dict[int, _FourCElementData],
+    vtu_file_path: _Path,
+    binary: bool = False,
+) -> None:
+    """Dump the information contained in the mesh representation to the
+    FourCIPP input file, in vtu format.
```
davidrudlstorfer: That sounds a bit misleading, we are dumping the mesh representation into vtu and then add the necessary information into the fourcipp input file.

src/beamme/four_c/input_file_dump_functions.py
```diff
@@ -484,3 +492,169 @@ def _dump(section_name: str, dictionary_list: _List):
...
+        )
+    fourc_input.combine_sections(
+        {
+            "STRUCTURE GEOMETRY": {
+                "FILE": vtu_file_path.name,
+                "SHOW_INFO": "detailed_summary",
+                "ELEMENT_BLOCKS": element_blocks,
+            }
+        }
+    )
```
davidrudlstorfer, line 638: I did not check it further, but if I remember correctly the combine sections overwrites existing keys right? Should we check here if the entries exist before overwriting them?

tests/conftest_result_comparison.py
```diff
@@ -110,6 +110,89 @@ def _assert_results_close(
     return _assert_results_close
 
 
+@pytest.fixture(scope="function")
+def assert_four_c_input_file_close(
```
davidrudlstorfer: This functionality should also move into the `assert_results_close` function so we still call everything through the central comparison function

davidrudlstorfer (dismissed):
Thanks a lot for the amazing feature @isteinbrecher! Really nice how simple the switch to vtu input was in the end (although one should not forget all the PRs that came before that ;) )

Thanks for all the changes and the new functionality.

I only have a few minor comments; the biggest thing that needs change in my opinion is the result comparison in the tests. We should still use the general compare function rather than introducing a new one here.

tests/conftest_result_comparison.py
```diff
@@ -87,6 +91,37 @@ def _assert_results_close(
...
+                add_header_default=False,
+                add_header_information=False,
+                add_footer_application_script=False,
+            )
+            _assert_results_close(
+                input_file_reference_path, input_file_path, rtol=rtol, atol=atol
+            )
+            _assert_results_close(mesh_reference_path, mesh_path, rtol=rtol, atol=atol)
+            return
+
```
davidrudlstorfer: would it be possible to move all of this into `convert_to_primitive_type`?

Further, would it be possible to just grep the Structure geometry of the 4C input file and check in the code if the mesh is vtu?

Then we could also remove that variable again

src/beamme/four_c/input_file_dump_functions.py
isteinbrecher: I added a check at the beginning of the function.

tests/conftest_result_comparison.py
isteinbrecher: I played around a little bit and it was possible to move everything into the `convert_to_primitive_type` function.

We still need the new parameter `four_c_input_file_data_format`. Otherwise, the two calls to `convert_to_primitive_type` would be dependent on each other. Also, I prefer that we can explicitly state what the expected format is in the testing code and not implicitly rely on the reference file.

src/beamme/four_c/input_file.py
isteinbrecher: The two underlying functions are called `dump_mesh_representation_to_input_file_legacy` and `dump_mesh_representation_to_input_file_vtu`. I would keep this argument consistent with the naming in the code.

src/beamme/four_c/input_file_mappings.py
```diff
@@ -161,6 +168,14 @@
...
     ): "DESIGN SURF MORTAR CONTACT CONDITIONS 3D",
 }
+# The definition of boundary conditions in 4C depends on the type of mesh output
+# (e.g., legacy or external mesh). When the actual input file is dumped, the correct
+# definition for geometry sets has to be used. BeamMe does not know all possible
+# boundary condition sections in advance, therefore, we store all used boundary
+# condition sections here, so we can check against them when dumping the input file.
+INPUT_FILE_MAPPINGS["known_boundary_condition_sections"] = set(
+    INPUT_FILE_MAPPINGS["boundary_conditions"].values()
+)
```
isteinbrecher, line 178: I am not a fan of this current solution, this will be cleaned up in a future PR.

src/beamme/four_c/input_file.py
```diff
@@ -176,6 +179,8 @@ def dump(
         self,
         input_file_path: str | _Path,
         *,
+        mesh_format: str = "legacy",
```
isteinbrecher: @davidrudlstorfer what do we want to do with the default setting here? Personally, I would switch this to `vtu`, as this is the more performant and future proof variant. I think input files for 4C should work exactly the same with `vtu` and legacy input (the only exception is some CubitPy stuff, that will follow soon).

If we want to switch this default, the question is when, I would even be fine with doing it in this PR. The sooner people use this feature the sooner we should be able to fix some remaining bugs/corner cases.

davidrudlstorfer: I have absolutely zero problem with switching the default to vtu. To keep it easier I would do that in a follow-up PR because the testing probably also needs to be updated (just one line but nevertheless).

I will personally probably continue to use the text based input file with my small examples because they always only consist of a few elements and then I can directly tweak stuff in the input file. But it is just one line that I need to change in my personal repo

src/beamme/four_c/input_file.py
davidrudlstorfer: To phrase it differently, I would still rename both functions and this key here.

Could we maybe do "vtu" and "yaml"? Then we also directly see what is meant.

With legacy I am not super happy because it doesn't say what it is actually doing. If the real "legacy" input, i.e., dat style strings will be removed and only yaml style element read in is possible this is still called legacy although it is not legacy.

yaml would make a lot of sense in my opinion, so my favourite would be 

`mesh_format: "yaml"`

and 

`dump_mesh_representation_to_input_file_yaml`

tests/conftest_result_comparison.py
```diff
@@ -127,8 +149,24 @@ def convert_to_primitive_type(
...
+            # Return input file sections as dictionary
+            sections = (
+                InputFile().from_4C_yaml(input_file_path=obj).fourc_input.sections
+            )
+            if "STRUCTURE GEOMETRY" in sections:
+                # If external geometry is given, add the vtu file to the data structure
+                # for comparison.
+                mesh_file_name = Path(sections["STRUCTURE GEOMETRY"]["FILE"])
+                # Call the reference file function so we register this file in the used
+                # reference files.
```
davidrudlstorfer: If I get this correct we only do this here so the file is registered? This is not super straightforward and not easy to understand

This functionality should directly go into the `get_corresponding_reference_file_path` function imo.

There we should just check if a ["STRUCTURE GEOMETRY"]["FILE"] key is present, if yes => also register the vtu file as used reference file.

And here we just use the same name as the yaml input file?

(I wanna get rid of the get_corresponding_reference_file_path function in here)

isteinbrecher: Yes, this is needed to the test for the reference file passes.

> There we should just check if a ["STRUCTURE GEOMETRY"]["FILE"] key is present, if yes => also register the vtu file as used reference file.

I don't think this is a suitable alternative. We would need to load the yaml file inside `get_corresponding_reference_file_path` and then assume that the mesh file should be there.

I also don't like `get_corresponding_reference_file_path` being there. What we could to is simply add the file to `USED_REFERENCE_FILES` here.

isteinbrecher:
> I have absolutely zero problem with switching the default to vtu. To keep it easier I would do that in a follow-up PR because the testing probably also needs to be updated (just one line but nevertheless).

Sounds good, then let's to this in a follow up PR

> Could we maybe do "vtu" and "yaml"? Then we also directly see what is meant.

I added one commit where the naming of legacy is changed in all ob BeamMe to `yaml` and `vtu`.


> (I wanna get rid of the get_corresponding_reference_file_path function in here)

I added a second commit where this is done as described here https://github.com/beamme-py/beamme/pull/649#discussion_r3379417464

@davidrudlstorfer Let me know what you think.

---

Side question, is it possible to tell FourCIPP to write nodes/elements with the pure yaml data structure and not the dat style strings?

tests/conftest.py
```diff
@@ -282,9 +282,8 @@ def _get_corresponding_reference_file_path(
                 f"File path: {corresponding_reference_file_path} does not exist"
             )
 
-        # Track usage if flag is enabled
-        if request.config.getoption("--check-for-unused-reference-files"):
-            USED_REFERENCE_FILES.add(corresponding_reference_file_path.resolve())
+        # Track usage of reference files
+        USED_REFERENCE_FILES.add(corresponding_reference_file_path.resolve())
```
davidrudlstorfer, line 286: Could this be bad that we now always do that?

tests/conftest_result_comparison.py
davidrudlstorfer, line 1: This is now much less intrusive than the original implementation! Thanks for changing that.

Maybe there's a way we can get rid of the `four_c_input_file_data_format` but we can also check that in the future.

davidrudlstorfer (approved):
> Side question, is it possible to tell FourCIPP to write nodes/elements with the pure yaml data structure and not the dat style strings?

Regarding this - as far as I know 4C is not capable of doing that yet?

Thanks a lot for all the review changes, you can merge this :)

tests/conftest.py
isteinbrecher: I don't think so, as this only gets checked if the flag `--check-for-unused-reference-files` is activated.

isteinbrecher:
> Regarding this - as far as I know 4C is not capable of doing that yet?

You are right, this is not possible yet, I got that confused with the fact that we can read json files (with legacy input).

tests/conftest_result_comparison.py
isteinbrecher: Sounds good!

PR #650, opened 2026-06-08
Improve testing of vtu files

isteinbrecher:
This PR related to the testing in #649.

We used to have a custom compare mechanism for `vtu` files. Now this is integrated into the `convert_to_primitive_type` and `custom_fourcipp_comparison` functions.

### DIFF
--- a/tests/conftest_result_comparison.py
+++ b/tests/conftest_result_comparison.py
@@ -29,7 +29,7 @@ from typing import Any, Callable
 
 import numpy as np
 import pytest
-import vtk
+import pyvista as pv
 import xmltodict
 from vistools.vtk.compare_grids import compare_grids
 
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (dismissed):
really really nice improvement :) thanks for adding this, improves the readability of the comparison by a lot

davidrudlstorfer (approved):

davidrudlstorfer:
@isteinbrecher I just checked the performance tests, and somehow the logo test is now in there being executed 🧐

`116 tests/other/test_other_logo.py .`

isteinbrecher:
> @isteinbrecher I just checked the performance tests, and somehow the logo test is now in there being executed 🧐

I think this was introduced with #640, where we added the CubitPy marker to some performance tests. The testing flags thing is a little complicated. Maybe the best option would be to allow to call pytest with `and` `or` combination of testing markers, then it should be more flexibel. However, I am also thinking about removing the cubit dependency in the performance tests, so this issue might go away.

Anyway, this is not related to this PR, so I would continue with this one.

PR #651, opened 2026-06-12
Cubit exo import

isteinbrecher:
This PR allows to directly convert exodus files into a BeamMe mesh representation.

Some modifications to testing and dependencies are required:
- ~Tests that require CoreformCubit are now marked with `@pytest.mark.coreform`.~
- One exodus file is checked in to the repo to allow testing of the exodus conversion on all systems. This requires CubitPy, since we use some utility function of CubitPy in the exodus conversion. Thus, CubitPy (not CoreformCubit) is now a required dependency for standard testing.


davidrudlstorfer:
Is this already ready for review?:)

isteinbrecher:
@davidrudlstorfer this is now ready.

Don't mind the failing pipeline, once #658 is merged and CubitPy is updated, this error will go away.

Sorry for the huge PR, but most of the large changes are just testing restructuring. The main changes are the ones to the actually BeamMe code.

pyproject.toml
```diff
@@ -35,7 +35,10 @@ version_scheme = "post-release"
 local_scheme = "no-local-version"
 
 [project.optional-dependencies]
-cubitpy = ["cubitpy"]
+cubitpy = [
+  "cubitpy",
+  "netCDF4"
```
davidrudlstorfer, line 40: Shouldn't this be a direct dependency of cubitpy which is installed through it?

tests/conftest_result_comparison.py
```diff
@@ -47,9 +49,7 @@
 
 
 @pytest.fixture(scope="function")
-def assert_results_close(
-    tmp_path, current_test_name, get_corresponding_reference_file_path
-) -> Callable:
```
davidrudlstorfer, line 52: Nice that this is gone again!

tests/conftest_result_comparison.py
```diff
@@ -142,16 +140,42 @@ def convert_to_primitive_type(
         elif obj.suffix == ".vtu":
             return pv.read(obj)
 
+        elif obj.suffix in (".exo", ".e"):
```
davidrudlstorfer, line 142: Should we really test something like this in BeamMe? Wouldn't this make more sense to compare the exo stuff in cubitpy?

tests/conftest_result_comparison.py
davidrudlstorfer: On a second note, was this just unused? 😅

davidrudlstorfer (dismissed):
Really clean and easy to understand code :)

pyproject.toml
isteinbrecher: Yes, `netCDF4` is a direct dependency of CubitPy, but we also directly use it in BeamMe now, that's why I prefer to still explicitly list it here.

We could rename the dependency group like
```
exodus = [
  "cubitpy",
  "netCDF4"
```
but I am afraid that this will cause even more confusion.

tests/conftest_result_comparison.py
isteinbrecher: Yes, I think I accidentally added that some commits back.

pyproject.toml
davidrudlstorfer: True, sounds good. Let's leave it in the current form

davidrudlstorfer (dismissed):

tests/conftest_result_comparison.py
isteinbrecher: That is a good point, I will have a look if it is easily possible to move this data generation function to CubitPy.

isteinbrecher: I moved this to CubitPy, we now only import the functionality here.

This should now finally be ready!

davidrudlstorfer (approved):
Nice :)

PR #652, opened 2026-06-12
Add vtu mesh output to performance test

isteinbrecher:
This PR adds the dump to vtu functionality to the performance tests.

Also, the reference time for `Add large beam mesh to another` is increased since this has been failing for some time now (some other reference times might still have to be increased).

### DIFF
--- a/tests/performance/test_performance_beamme.py
+++ b/tests/performance/test_performance_beamme.py
@@ -319,7 +319,7 @@ def test_performance_beamme_add_beams_to_mesh(large_beam_mesh, evaluate_executio
         "BeamMe: Add large beam mesh to another",
         large_beam_mesh_copy.add,
         args=[large_beam_mesh],
-        expected_time=0.1,
+        expected_time=0.15,
     )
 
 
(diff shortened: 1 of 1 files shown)


tests/performance/test_performance_beamme.py
```diff
@@ -457,19 +457,32 @@ def test_performance_beamme_add_mesh_to_input_file(large_beam_input_file):
 
 
 @pytest.mark.performance
+@pytest.mark.parametrize(
+    ("log_name", "mesh_format", "expected_time"),
+    [
+        ("BeamMe: Dump input file with large beam mesh (yaml)", "yaml", 15),
+        ("BeamMe: Dump input file with large beam mesh (vtu)", "vtu", 0.65),
```
isteinbrecher: This shows how much performance can be gained by using the `vtu` output.

@maxfirmbach

davidrudlstorfer: wow :O

davidrudlstorfer (dismissed):

PR #653, opened 2026-06-12
Refactor visualisation

isteinbrecher:
With the mesh representation we can directly represent a BeamMe mesh in vtu format. This makes the old vtk writer obsolete. It is much cleaner and easier to simply use the mesh representation / vtu data structure for visualisation.

This PR replaces the old vtk writer and uses mesh representation / vtu directly for visualisation.

Note: The functionality to display smooth beams with `beam_centerline_visualization_segments` is removed - as far as I know this has never been by anyone in practice. Besides that, the "smooth" curve was just a guess by BeamMe and not the actual geometry in the finite element code (which depends on the actual implementation, which BeamMe does not know about).


davidrudlstorfer:
Let me know once I should review this :)

isteinbrecher:
@davidrudlstorfer This should now be ready for review.

tests/performance/test_performance_beamme.py
```diff
@@ -487,39 +487,14 @@ def test_performance_beamme_dump_input_file(
...
-        kwargs={
-            "output_name": "performance_testing_beam",
-            "output_directory": tmp_path,
-            "beam_centerline_visualization_segments": 5,
-        },
-        expected_time=8.1,
+        "BeamMe: Write beam mesh to VTU",
+        large_beam_mesh.write_vtu,
+        args=[tmp_path / "performance_testing_beam.vtu"],
+        expected_time=2.0,
```
isteinbrecher: We now create the output of a mesh that is 4 times as large and the runtime is about halved.

isteinbrecher:
```
BeamMe: Write beam mesh to VTU                                         0.829s < 2.000s
```
Actually, even faster, I will update the reference time.

src/beamme/core/mesh.py
```diff
@@ -716,7 +718,7 @@ def get_nodes_by_function(self, *args, **kwargs):
         return _get_nodes_by_function(self.nodes, *args, **kwargs)
 
     def get_mesh_representation(
-        self, material_to_i_global: dict[_Material, int]
+        self, material_to_i_global: dict[_Material, int] | None = None
```
davidrudlstorfer, line 721: I do not know if it makes sense here (Chatgpt was not sure itself haha)


```suggestion
        self, material_to_i_global: dict[_Material, int] | None = {}
```

but I guess we could just initialize this as {} directly instead of the first two lines below?

tests/integration/test_integration_core_mesh_visualization.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def test_integration_core_mesh_visualization_beam(
+    assert_results_close,
+    get_default_test_beam_material,
+    get_corresponding_reference_file_path,
+):
+    """Create a sample mesh and check the VTU output."""
+
+    # Create the mesh.
```
davidrudlstorfer: Should we still keep those comments?

They are completely self explained

tests/integration/test_integration_core_mesh_visualization.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Convert the solid mesh to beamme objects.
+    _, mesh = import_four_c_model(
+        input_file_path=get_corresponding_reference_file_path(
+            reference_file_base_name="test_other_create_cubit_input_files_tube"
+        ),
+        convert_input_to_mesh=True,
+    )
+
+    # Compare the vtu files.
```
davidrudlstorfer: Also these ones

tests/performance/test_performance_beamme.py
davidrudlstorfer: Crazy how fast this is now 🤩

tests/reference-files/test_integration_four_c_solid_shell_thickness_direction_visualization.vtu
davidrudlstorfer, line 1: I haven't checked this is detail; why is this now so much more data?

tests/conftest_result_comparison.py
```diff
@@ -210,6 +210,9 @@ def str_to_float(string: str) -> str | float:
             for fragment in line.split(",")
         ]
 
+    if isinstance(obj, pv.UnstructuredGrid):
+        return obj
+
```
davidrudlstorfer, line 215: Nice how easy this is now ;)

davidrudlstorfer (dismissed):
Really clean and nice PR! Only a few comments from my side that should be able to be resolved really quickly :)

- Nice that so much code could be deleted from the elements itself :)

- Also thanks for updating and cleaning so many docstrings :)

- One question; are the copilot comments of interest?

src/beamme/core/mesh.py
isteinbrecher: In this case, it should be fine to directly use it in the function definition. But, I try to not use mutable data types in keyword definition, because that can results in super hard to catch bugs.

davidrudlstorfer: Then let's leave it like that :)

tests/integration/test_integration_core_mesh_visualization.py
isteinbrecher: Good point, in these small test cases, they don't serve any purpose, I removed them.

tests/conftest_result_comparison.py
isteinbrecher: Yes, it is really nice how universal and jet simple this functionality is.

tests/reference-files/test_integration_four_c_solid_shell_thickness_direction_visualization.vtu
isteinbrecher: We used to split the vtu files into solid and beam parts. Now we have everything in one file (as we also need this for 4C). Thus, we need to store the beam information, e.g., rotation vectors, base vectors, cross-section diameter, for all points/cells, also ones that are only related to solid nodes/elements.

That is the reason this reference file now carries more data. If the vtu file is stored in binary format, these data arrays are compressed and result in a not much increased overall file size.

We have the data vectors that allow us to split the vtu grids into beam and solid parts, so users can use this in their post processing scripts. However, for BeamMe I personally like this unified approach.

isteinbrecher:
Thanks for the review @davidrudlstorfer 

> One question; are the copilot comments of interest?

I added some, the comments regarding the space-time elements are relevant, however, I would adapt the whole space time functionality to the new output format in a stand alone PR.

davidrudlstorfer (approved):
Thanks for the changes, let's get this merged :)

PR #656, opened 2026-06-17
Add get named node set functionality

isteinbrecher:
Add a method to `Mesh` that returns all geometry sets with a given name.

### DIFF
--- a/tests/beamme/core/test_beamme_core_mesh.py
+++ b/tests/beamme/core/test_beamme_core_mesh.py
@@ -28,14 +28,13 @@ from beamme.core.conf import bme
 from beamme.core.coupling import Coupling
 from beamme.core.element_beam import Beam, Beam3
 from beamme.core.geometry_set import GeometrySet
+from beamme.core.material import MaterialBeamBase
 from beamme.core.mesh import Mesh
 from beamme.core.node import Node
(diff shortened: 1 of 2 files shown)


tests/beamme/core/test_beamme_core_mesh.py
```diff
@@ -121,3 +119,60 @@ def test_beamme_core_mesh_multiple_couple_nodes(get_default_test_beam_material):
         assert len(coupling_nodes) == 2
         assert beam_set_1["end"].get_points()[0] in coupling_nodes
         assert beam_set_2["start"].get_points()[0] in coupling_nodes
+
+
+def test_beamme_core_mesh_get_named_geometry_sets():
+    """Test the `get_named_geometry_sets` method."""
+
+    # Create mesh object
```
davidrudlstorfer: Same with the comments here as before

davidrudlstorfer (dismissed):

tests/beamme/core/test_beamme_core_mesh.py
isteinbrecher: Good point, I need to cut down on my comments in the tests 😅

davidrudlstorfer (approved):

PR #657, opened 2026-06-17
Improve naming of CubitPy / Coreform in testing

isteinbrecher:
This is a perquisite for #651.

We need to split up the dependency on CubitPy and the dependency on an actual CoreformCubit installation for testing. This is done in this PR, where the test marker `@pytest.mark.cubitpy` is replaced with `@pytest.mark.coreform`.

Tests will fail since this is still run with the test suite on `main`, but they should pass once they are merged, as I only did a global replace and adapted the readme.

### DIFF
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Run the test suite and upload results on failure
         uses: ./.github/actions/run_tests
         with:
-          additional-pytest-flags: "--4C --ArborX --CubitPy --check-for-unused-reference-files --cov-fail-under=93"
+          additional-pytest-flags: "--4C --ArborX --Coreform --check-for-unused-reference-files --cov-fail-under=93"
           cubit-root: ${{ steps.cubit.outputs.cubit_root }}
       - name: Free cubit license
         if: ${{ always() }}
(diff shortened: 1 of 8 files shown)


davidrudlstorfer (dismissed):
Just to quickly understand, what's the difference between the two?

Cubitpy is the IMCS repo and Coreform is the programm behind that?

isteinbrecher:
> Cubitpy is the IMCS repo and Coreform is the programm behind that?

Good question! [Coreform Cubit](https://coreform.com/coreform-cubit/) is an FE meshing software (previously also known as Cubit or Trellis). Coreform provides a python interface from within the Coreform GUI.

[CubitPy](https://github.com/imcs-compsim/cubitpy) is a pure python package that wraps the Coreform python interpreter for easy usage in any python environment. It also contains a some utility functions for converting exodus in python to 4C useable input files.

With #651, we allow direct handling of exodus files in BeamMe, for that some of the CubitPy utility functions are needed (but this does not require an actual working Coreform Cubit installation). We will have some tests that depend on CubitPy but not Coreform Cubit. These tests should also run during the standard testing. Thus, the idea is to install CubitPy for all testing environments (similar to how we do it with FourCIPP) and mark the tests that need Coreform Cubit and not CubitPy.

This PR simply changes the naming in the testing environment.

davidrudlstorfer (approved):

PR #658, opened 2026-06-18
Use CubitPy config file during testing

isteinbrecher:
New versions of CubitPy need a configuration file. This PR adds this in the BeamMe testing workflow.

This closes #584

### DIFF
--- a/.github/actions/cubitpy_config/action.yml
+++ b/.github/actions/cubitpy_config/action.yml
@@ -1,0 +1,25 @@
+name: cubitpy_config
+description: Setup a CubitPy config file
+inputs:
+  cubit-root:
+    description: Path to the cubit installation
+    required: true
+outputs:
+  cubitpy_config_path:
+    description: "Path to the CubitPy config file"
+    value: ${{ steps.cubitpy_config.outputs.cubitpy_config_path }}
+runs:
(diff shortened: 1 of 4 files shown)


isteinbrecher:
@davidrudlstorfer I think the only way to check if this works is to merge to main 😅

.github/actions/cubitpy_config/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    required: true
+outputs:
+  cubitpy_config_path:
+    description: "Path to the CubitPy config file"
+    value: ${{ steps.cubitpy_config.outputs.cubitpy_config_path }}
+runs:
+  using: composite
+  steps:
+    - name: Create the CubitPy config file
+      id: cubitpy_config
```
davidrudlstorfer: do we need this id?

davidrudlstorfer (dismissed):
Looks good :)

Let's get this merged and see if it works. Small changes can still be done in some of the next PR's 😅

.github/actions/cubitpy_config/action.yml
isteinbrecher: I think we don’t need it. I will remove it a then merge this.

PR #659, opened 2026-06-18
Skip CubitPy warnings

isteinbrecher:
This PR skips all warnings coming from CubitPy in testing.

Ideally, we rework the code that causes the warnings, however, for testing purposes this is not always possible.

Depends on:
- #658
- https://github.com/imcs-compsim/cubitpy/actions/runs/27759889949/job/82131196366?pr=137 + a new CubitPy release

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,3 +71,8 @@ markers = [
   "coreform: tests in combination with Coreform Cubit",
   "performance: performance tests"
 ]
+filterwarnings = [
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (dismissed):
Should we maybe add a comment to this?

Like which warnings are suppressed and why

isteinbrecher:
Good point, I added a comment why we need to skip the CubitPy warnings.

davidrudlstorfer (approved):
Thanks :)

PR #660, opened 2026-06-18
Fix CubitPy config action

isteinbrecher:
This PR fixes the action that creates the CubitPy configuration file.


isteinbrecher:
Looking good now, should work on `main` as well.

@davidrudlstorfer turns out we needed the `id` https://github.com/beamme-py/beamme/pull/658#discussion_r3437347680 😂

davidrudlstorfer:
Thanks for trying 😅

PR #666, opened 2026-07-06
fix: protected workflow by allowing unsafe pr checkout

### LINKED ISSUE #664: Broken PR CI pipeline
After updating from `actions/checkout@v6` to `actions/checkout@v7` (https://github.com/beamme-py/beamme/pull/662) the PR workflows seem to be broken with the following error message (in the step "Checkout PR code", https://github.com/beamme-py/beamme/actions/runs/28011857030/job/82906664329):

> Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

davidrudlstorfer:
closes #664

### DIFF
--- a/.github/workflows/testing_protected.yml
+++ b/.github/workflows/testing_protected.yml
@@ -37,6 +37,8 @@ jobs:
           # For PR runs, checkout PR code; otherwise fallback to default branch or ref
           ref: ${{ github.event.pull_request.head.ref || github.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          allow-unsafe-pr-checkout: true # required to checkout PR code from forks, PRs from external people need to be approved anyways
+          persist-credentials: false
       - name: Setup python environment
         uses: ./.github/actions/setup_python_environment
         with:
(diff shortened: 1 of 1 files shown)


davidrudlstorfer:
@isteinbrecher so i looked into the details of the new checkout@v7 workflow

Basically we allow access to our GITHUB TOKENS and SECRETS with the pull_request_target workflow. If people execute code they could get the TOKENS and SECRETS (in our case the cubit credentials).

We already have it setup in a way that people without collaborator status need to be approved manually anyways (see dependabot PRs).

I think we can trust ourselves enough that we two do not do anything to leak those credentials, and external people cannot run the workflow anyways.

I think this should resolve this

I also had a brief chat with ChatGPT (https://chatgpt.com/share/6a4b8d90-c4f8-83eb-af4f-53e58e1ae651) which more or less supported this.

isteinbrecher (approved):
Thanks for looking into this and the explanations @davidrudlstorfer !

PR #667, opened 2026-07-08
Temporary test to check pipeline

davidrudlstorfer:


davidrudlstorfer:
@isteinbrecher so the fix from #666 worked, I was unsure because your pipelines still failed but I guess this was due to an old base branch

isteinbrecher:
Thanks for checking this! I also saw that the other pipelines did not pass, that is why I force merged the other two open PRs, should all be fine now.

davidrudlstorfer:
So the checkout succeeded, therefore closing this PR again

Issue #463, opened 2025-09-19
Updates after supporting python >=3.11


isteinbrecher:
This issue is meant for gathering topics that should be addressed when we only support python >=3.11.

isteinbrecher:
See https://github.com/beamme-py/beamme/pull/461#discussion_r2358381203 - we should replace the type hints for types `Type[BeamMeType]` with `Self`.

davidrudlstorfer:
Closed with #671

Issue #423, opened 2025-07-15
Increase documentation quality


davidrudlstorfer:
This would be the last (_famous last words_) big meta issue for BeamMe.

With the switch to pytest, a src based layout and the overall switch to .yaml based input the code quality increased by orders of magnitudes.

The last step would be to increase the documentation quality (which 1:1 gets piped into our API docu).

For that we should consider these points:

- [ ] Add type hints (https://github.com/4C-multiphysics/fourcipp/pull/54 could be helpful, https://github.com/gilrrei/MonkeyType could be helpful)
- [ ] Enforce type hints (https://github.com/4C-multiphysics/fourcipp/pull/80 could be helpful)
- [ ] Update/Add docstrings according to Google Style across the entire code base (@isteinbrecher I think we can close #305 as this is an issue for only one file). I think it would make sense to do them file wise. This way the review and implementation effort is limited)
- [ ] agree on one global "standard" for file docstrings (these are the module descriptions within the API docu), sometimes we use "this file defines", "this module defines", ...

@isteinbrecher let me know what you think:)

isteinbrecher:
I agree with all points! In the long term that is definitely something where we should aim to get to.

PR #671, opened 2026-07-08
feat: update to Python 3.14

isteinbrecher:
Reopen https://github.com/beamme-py/beamme/pull/620/changes to check if python 3.14 works (splinepy is now available ein 3.14 https://github.com/isosuite/splinepy/pull/494, thanks @clemens-fricke).

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os-version: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.11", "3.14"]
     runs-on: ${{ matrix.os-version }}
     steps:
       - name: Checkout PR code
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,15 +32,15 @@ jobs:
       - name: Install BeamMe
         uses: ./.github/actions/install_beamme
         with:
-          # Test coverage and editable install with Python 3.10, otherwise we use a
+          # Test coverage and editable install with Python 3.11, otherwise we use a
           # non-editable installation and turn off coverage, because the coverage
           # only works in editable mode.
           install-command: >-
-            ${{ matrix.python-version == '3.10' && '-e .[cubitpy,dev,fourc]' || '.[cubitpy,dev,fourc]'}}
+            ${{ matrix.python-version == '3.11' && '-e .[cubitpy,dev,fourc]' || '.[cubitpy,dev,fourc]'}}
       - name: Run the test suite and upload results on failure
         uses: ./.github/actions/run_tests
         with:
           # The single space in the empty string is required, otherwise GitHub
           # evaluates the if clause wrong.
           additional-pytest-flags: >-
-            ${{ matrix.python-version == '3.10' && ' ' || '--no-cov' }}
+            ${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,7 +36,7 @@ jobs:
           # non-editable installation and turn off coverage, because the coverage
           # only works in editable mode.
           install-command: >-
-            ${{ matrix.python-version == '3.11' && '-e .[cubitpy,dev,fourc]' || '.[cubitpy,dev,fourc]'}}
+            ${{ matrix.python-version == '3.11' && '-e .[dev]' || '.[dev]'}}
       - name: Run the test suite and upload results on failure
         uses: ./.github/actions/run_tests
         with:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -42,5 +42,15 @@ jobs:
         with:
           # The single space in the empty string is required, otherwise GitHub
           # evaluates the if clause wrong.
-          additional-pytest-flags: >-
-            ${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}
+          additional-pytest-flags: "${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}"
+
+  all-tests-pass:
+    name: All tests passed
+    runs-on: ubuntu-latest
+    needs: [beamme-testing-from-source-code]
+    if: always()
+    steps:
+      - name: Check test matrix result
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
--- a/README.md
+++ b/README.md
@@ -219,11 +219,11 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
 
 ### Python environment
 
-BeamMe is tested with, and supports Python versions 3.10-3.13. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
+BeamMe is tested with, and supports Python versions 3.10-3.14. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
 - A [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) environment can be created and loaded with
   ```bash
   # Create the environment (this only has to be done once)
-  conda create -n beamme python=3.13
+  conda create -n beamme python=3.14
   # Activate the environment
   conda activate beamme
   ```
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
 
 ### Python environment
 
-BeamMe is tested with, and supports Python versions 3.10-3.14. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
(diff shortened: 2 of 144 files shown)


davidrudlstorfer:
This is the corresponding issue to the fail in the code check https://github.com/PyCQA/docformatter/issues/340

davidrudlstorfer:
There is a newer version available of docformatter already, so we just need to update.

@isteinbrecher If you want i can update all pre-commit hooks for this PR

davidrudlstorfer:
Additionally I would include a small "ensure all tests pass" workflow like in 4C and FourCIPP, then we can remove the individual requirements for the workflow

And I would also directly include #463 to close it

isteinbrecher:
> @isteinbrecher If you want i can update all pre-commit hooks for this PR

Yes please. Other than that, I think we can merge this.

isteinbrecher:
> Additionally I would include a small "ensure all tests pass" workflow like in 4C and FourCIPP, then we can remove the individual requirements for the workflow

We can do that, I just don't fully understand what requirements we would not need anymore.

> And I would also directly include https://github.com/beamme-py/beamme/issues/463 to close it

Sounds good to me! So we drop 3.10 support? (Totally fine)

isteinbrecher:
@davidrudlstorfer you can directly push to the branch on my fork or reopen your original PR, whatever is easier.

davidrudlstorfer:
@isteinbrecher could you maybe briefly look into the new mypy warnings?

```bash
src/beamme/core/mesh.py:910: error: Argument 1 to "append" of "list" has incompatible type "int | None"; expected "int"  [arg-type]
src/beamme/space_time/beam_to_space_time.py:279: error: Unsupported operand types for + ("None" and "int")  [operator]
src/beamme/space_time/beam_to_space_time.py:279: note: Left operand is of type "int | None"
src/beamme/four_c/model_importer.py:599: error: "type" has no attribute "element_type"  [attr-defined]
```

I think you would be rather quick to fix this

Other than that i updated all pre-commit hooks. A lot of changes occur due to the new docstring formatting (changed to black line length which is compatible with ruff)

Other than that I needed to add to comments due to this issue https://github.com/PyCQA/docformatter/issues/350

isteinbrecher:
> I think you would be rather quick to fix this

Done, I pushed the changes.

davidrudlstorfer:
One question: 

I currently installed a new python 3.14 env for testing and i only installed dev and fourc (as in the readme stated)

Then it fails because cubitpy is necessary in the result comparison and if this is fixed it also fails because netcdf is missing.

This should probably be guarded somehow? Or is cubitpy now a required dependency? I would aim that this is not a required dependency

isteinbrecher:
> Then it fails because cubitpy is necessary in the result comparison and if this is fixed it also fails because netcdf is missing.
> This should probably be guarded somehow? Or is cubitpy now a required dependency? I would aim that this is not a required dependency

With the latest changes, the packages installed with `cubitpy` are required for general testing (just the python packages, not Coreform). Similar to how we need the `fourc` packages (i.e., FourCIPP) for testing. In the actually BeamMe code we guard agains usage of CubitPy without it being installed (we actually don't do that for FourCIPP, but that is another topic).

So CubitPy is basically handled the same as FourCIPP. Not needed for the core BeamMe functionality, but required for testing.

I see three options:
- Add guards in testing for not installed packages. But that would mean that we would have to further split up which tests are run when. Not my favourite.
- Simply change the one part of the readme to something like `pip install -e .[dev,fourc,cubitpy]`
- Add all packages that are required for testing to the `dev` package list. This would be my preferred solution.

I am curious on your thoughts. But I would adapt this in a separate PR to not make this one too broad.

README.md
```diff
@@ -219,11 +219,11 @@ interaction, <https://athene-forschung.unibw.de/143755>, (2022)</span>
 
 ### Python environment
 
-BeamMe is tested with, and supports Python versions 3.10-3.13. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
+BeamMe is tested with, and supports Python versions 3.10-3.14. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
```
isteinbrecher:
```suggestion
BeamMe is tested with, and supports Python versions 3.11-3.14. It is recommended to use a virtual Python environment such as [Conda](https://anaconda.org/anaconda/conda)/[Miniforge](https://conda-forge.org/download/) or [venv](https://docs.python.org/3/library/venv.html).
```

isteinbrecher (commented):
I had a look and are there even any cases of the issue https://github.com/beamme-py/beamme/issues/463 left in the code base?

davidrudlstorfer:
I will push the fixes for #463 and the remaining stuff in a second

davidrudlstorfer:
So now i changed quite a lot and I think it is done

- please just check the individual commits, they should be pretty self explanatory :)

- the biggest change is the formatting change that no longer an empty line is there after a docstring, this is actually Python standard

- I updated the deps also in this PR directly (following https://github.com/pypa/pip/issues/11296)

- I also introduced the two new all tests and all test build wheels success stages. Now the requirements for the PRs/branch protection simplified a lot and we do no longer need to change the Github settings once we change the Python versions or something else

I am still unsure on how to continue with our dev requirements. I think that beamme should be testable also without cubitpy (at least a minimum set of tests like the core). Currently one cannot even execute pytest because it fails immediately.

Now also #463 is also closed (see commits)

And I think this would then be a new BeamMe release :partying_face:

.github/workflows/testing.yml
```diff
@@ -43,4 +43,14 @@ jobs:
           # The single space in the empty string is required, otherwise GitHub
           # evaluates the if clause wrong.
           additional-pytest-flags: >-
-            ${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}
+            ${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }} #magic___^_^___line
```
davidrudlstorfer: this is a real WTF for me

this always get's added from the yaml formatter 

see https://github.com/google/yamlfmt/issues/84

davidrudlstorfer: solved by changing to a normal quoted string

davidrudlstorfer:
@isteinbrecher ready from my side, I cannot request your review because you are the PR author and my review is worthless here

so you can force merge once you are happy :)

src/beamme/core/geometry_set.py
```diff
@@ -19,11 +19,12 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
-"""This module implements a basic class to manage geometry in the input
-file."""
+"""This module implements a basic class to manage geometry in the input file."""
 
-from typing import KeysView as _KeysView
-from typing import Sequence as _Sequence
+from __future__ import annotations as _annotations
```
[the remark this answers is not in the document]
davidrudlstorfer: still worked with python 3.11, before it did not work

src/beamme/mesh_creation_functions/applications/beam_stent.py
```diff
@@ -50,39 +50,39 @@ def create_stent_cell(
...
-    Return
-    ----
-    mesh: Mesh
-        A mesh with this structure
+    """Create a cell of the stent.
+
+    This cell is on the x-y plane.
+        Args
+        ----
+        beam_class: Beam
```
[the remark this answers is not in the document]
davidrudlstorfer: this is just the old docu style that now gets formatted like that with the new ruff version, this should be updated with #423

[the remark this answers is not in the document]
davidrudlstorfer: updated in latest commit

davidrudlstorfer (approved):

pyproject.toml
```diff
@@ -42,6 +42,8 @@ cubitpy = [
 ]
 fourc = ["fourcipp"]
 dev = [
+  "beamme[cubitpy]",
```
isteinbrecher, line 45: Nice that this works!

.github/workflows/testing.yml
```diff
@@ -42,5 +42,15 @@ jobs:
         with:
           # The single space in the empty string is required, otherwise GitHub
           # evaluates the if clause wrong.
-          additional-pytest-flags: >-
-            ${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}
+          additional-pytest-flags: "${{ matrix.python-version == '3.11' && ' ' || '--no-cov' }}"
+
+  all-tests-pass:
```
isteinbrecher, line 47: Could you shortly explain why we need this?

isteinbrecher (commented):
@davidrudlstorfer thanks for all the changes, look very good, this definitely should lead to another BeamMe release (0.3.0?)

pyproject.toml
davidrudlstorfer: that took some time to figure out :D

.github/workflows/testing.yml
davidrudlstorfer: in the Github repo settings we specify the branch protection rules on which workflows need to be set so we can merge

Previously we needed to add all the workflows individually, i.e. tests python windows 3.10, tests python windows 3.13, .....

With this we just check if all of those tests are successfully passed and if yes this job is successfull

Therefore we only check in our branch protection rule if this one job is successfull. It is done identical in 4C so we do not need to add every chunked test to the requirements for the pull request.

You can see that down below in the successfull tests => required tests (logo on the right). Now only the all dependency test, and the two summary checks are required

isteinbrecher: Sounds good!

davidrudlstorfer:
@isteinbrecher AMAZING that all pipelines including publishing pipelines ran through 😍 

really happy about that:)

PR #674, opened 2026-07-08
Improve dumping of boundary conditions

isteinbrecher:
For 4C, the boundary condition sections in the input file look different if the mesh is in legacy yaml format or in vtu.

Thus, we need to create the correct data structure when dumping the input file. This required us to abuse the entry `INPUT_FILE_MAPPINGS["known_boundary_condition_sections"]` to track which sections in the input file are boundary conditions.

This behaviour is now cleaned up. When dumping a mesh to an input file, we convert each BeamMe boundary condition in a suitable data structure that can then easily be used for dumping to yaml/vtu later on.

### DIFF
--- a/src/beamme/four_c/boundary_condition_data.py
+++ b/src/beamme/four_c/boundary_condition_data.py
@@ -1,0 +1,60 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2018-2026 BeamMe Authors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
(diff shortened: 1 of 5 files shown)


src/beamme/four_c/boundary_condition_data.py
```diff
@@ -0,0 +1,60 @@
...
+
+    geometry_set_id: int
+    data: dict
+
+    def dump_to_input_file_yaml(self) -> dict:
+        """Dump the boundary condition data to a dictionary that can be written to a 4C
+        input file with the mesh in yaml format."""
+        return {"E": self.geometry_set_id, **self.data}
+
+    def dump_to_input_file_vtu(
```
davidrudlstorfer, line 43: Would it make sense to just provide one "dump_to_input_file" function and separate via an input flag "input_file_format"?

davidrudlstorfer (dismissed):
Just one minor comment

src/beamme/four_c/boundary_condition_data.py
isteinbrecher: Since the signature for the two functions is different (one requires a list of all the geometry sets in the mesh) I would keep two functions.

davidrudlstorfer (approved):

Issue #639, opened 2026-05-05
Failing performance tests


isteinbrecher:
In the last couple of days the performance tests have been failing, mainly `BeamMe: Find close nodes in large beam mesh`

This has likely to do with some changed TPL.

I re-ran a performance test that passed a couple weeks ago and now if failed: https://github.com/beamme-py/beamme/actions/runs/24628202231/job/72028452931

<details>

<summary>The packages that are different:</summary>

```
certifi
  new: 2026.4.22
  old: 2026.2.25

click
  new: 8.3.3
  old: 8.3.2

cyclopts
  new: 4.11.2
  old: 4.10.2

FourCIPP
  new: 1.102.0
  old: 1.92.0

idna
  new: 3.13
  old: 3.11

ipython
  new: 9.13.0
  old: 9.12.0

jedi
  new: 0.20.0
  old: 0.19.2

jsonschema_rs
  new: 0.46.4
  old: 0.46.1

jupyter-events
  new: 0.12.1
  old: 0.12.0

jupyter_server
  new: 2.18.1
  old: 2.17.0

jupyterlab
  new: 4.5.7
  old: 4.5.6

matplotlib
  new: 3.10.9
  old: 3.10.8

mistune
  new: 3.2.1
  old: 3.2.0

notebook
  new: 7.5.6
  old: 7.5.5

packaging
  new: 26.2
  old: 26.1

parso
  new: 0.8.7
  old: 0.8.6

pip
  new: 26.1.1
  old: 26.0.1

pre_commit
  new: 4.6.0
  old: 4.5.1

python-discovery
  new: 1.3.0
  old: 1.2.2

pyvista
  new: 0.48.0
  old: 0.47.3

scooby
  new: 0.11.2
  old: 0.11.0

trame-client
  new: 3.12.1
  old: 3.11.4

trame-server
  new: 3.11.0
  old: 3.10.0

trame-vtk
  new: 2.11.8
  old: 2.10.2

trame-vuetify
  new: 3.2.2
  old: 3.2.1

tzdata
  new: 2026.2
  old: 2026.1

virtualenv
  new: 21.3.1
  old: 21.2.4

wcwidth
  new: 0.7.0
  old: 0.6.0
```

</details>

I suspected either `numpy` or `scipy`, but the versions did not change. So maybe another package is the reason, or some change in testing hardware that specifically affects this test.

davidrudlstorfer:
Note: also the logo is still getting tested in the performance test suite

[...]

Issue #562, opened 2026-01-21
Investigate and fix warnings that are thrown during testing


davidrudlstorfer:
During testing a few warnings are thrown which

1. do come from 3rd party packages
2. could be deactivated because we purposefully get them during testing

Therefore, the warnings should be checked and "removed"

During our regular test suite the following warnings are there:

```bash
=============================== warnings summary ===============================
<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type swigvarlink has no __module__ attribute

tests/integration/test_integration_core_mesh.py::test_integration_core_mesh_transformations_with_solid[True-0.2-False-context2]
  /home/runner/work/beamme/beamme/src/beamme/core/mesh.py:538: UserWarning: There are elements which are not parallel to the y-z or x-y plane. This will lead to distorted elements!
    _warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

On windows also this warning is thrown:

```bash
[7](https://github.com/beamme-py/beamme/actions/runs/21199690791/job/60982692361#step:5:99)
tests/other/test_other_examples.py::test_other_examples_notebooks[examples/example_1_finite_rotations.ipynb]
  C:\hostedtoolcache\windows\Python\3.13.11\x64\Lib\site-packages\zmq\_future.py:718: RuntimeWarning: Proactor event loop does not implement add_reader family of methods required for zmq. Registering an additional selector thread for add_reader support via tornado. Use `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())` to avoid this warning.
    self._get_loop()
```

During our full test suite (with cubit) we get this additional warning:

```bash
tests/other/test_other_logo.py::test_other_logo
  /__w/beamme/beamme/src/beamme/mesh_creation_functions/beam_parametric_curve.py:156: IntegrationWarning: The maximum number of subdivisions (50) has been achieved.
    If increasing the limit yields no improvement it is advised to analyze 
    the integrand in order to determine the difficulties.  If the position of a 
    local difficulty can be determined (singularity, discontinuity) one will 
[...]

PR #675, opened 2026-07-08
Remove Coreform from performance testing

isteinbrecher:
This PR proposes to remove the part in performance testing that requires Cubit Coreform. A simple script is added that creates a structured grid for a block in exodus format. This is then created on the fly in testing and used to assess the performance of reading that file with BeamMe. This, it is not required to have a working Coreform installation for the performance tests.


.github/workflows/testing_performance.yml
davidrudlstorfer, line 1: Is there a reason to create a new workflow file for this?

I would just put that into the other testing workflow.

Originally we just split it because the other one is a workflow target

tests/create_structured_exodus_hex_mesh.py
```diff
@@ -0,0 +1,265 @@
...
+                        idx(point_index_x + 1, point_index_y + 1, point_index_z + 2),
+                        # volume center
+                        idx(point_index_x + 1, point_index_y + 1, point_index_z + 1),
+                    ]
+
+                else:
+                    raise ValueError("element must be 'hex8' or 'hex27'")
+
+                i_elem = i_x * ny * nz + i_y * nz + i_z
+                connectivity[i_elem] = conn
```
davidrudlstorfer, line 144: should this somehow be tested?

davidrudlstorfer (commented):
Looking good overall, so this means that we no longer need Coreform for performance tests. 

Is there something similar on the horizon for the remaining tests that use Coreform?

And this closes #639 right?

Regarding the test times, this is now slower because the write out takes longer? But then why do downstream tests fail, I thought we "cache" the object in between?

tests/create_structured_exodus_hex_mesh.py
isteinbrecher: That is a good point, I will add an integration test for this.

isteinbrecher:
> Looking good overall, so this means that we no longer need Coreform for performance tests.

Yes exactly, the performance tests can run on a standard runner.

> Is there something similar on the horizon for the remaining tests that use Coreform?

No, we still have actual CubitPy (with Coreform) functionality that we need to test. We have one single test `test_integration_four_c_model_importer_solid_element_types_from_cubitpy ` that checks code in `src/` (i.e., `import_cubitpy_model`).
All other test that require Coreform are somewhat secondary:
- We have tests that check the scripts that create the reference files based on Coreform (in `test_other_create_cubit_input_files.py`)
- The logo tests uses CoreForm to create the 2D Mesh

> And this closes https://github.com/beamme-py/beamme/issues/639 right?

The issue that the logo test is executed in the performance tests is solved, I did not have a look at the underlying performance test yet.

> Regarding the test times, this is now slower because the write out takes longer? But then why do downstream tests fail, I thought we "cache" the object in between?

I saw that too, and I have no idea why. I initially thought that there might be some difference between the 4C container and the default ubuntu container, but this is not the case. If we look at some of the tests:
```
BeamMe: Load solid mesh (no full import)                               0.868s < 2.500s
```
This is actually quire remarkable sine the solid mesh got about 10 times as large (the reason for the fast runtime are the changes in #651).
```
BeamMe: Load solid mesh (full import)                                  13.891s > 3.500s
```
This is ok, since the solid mesh got about 10 times as large.
```
BeamMe: Add large beam mesh to another                                 0.151s > 0.150s
```
Small runtime, but this one did not fail before.
```
BeamMe: Couple nodes in large beam mesh                                9.638s > 7.000s
```
I think this one already caused issues, see #639
```
BeamMe: Add large beam mesh to input file                              0.632s > 0.600s
BeamMe: Dump input file with large beam mesh (yaml)                    15.392s > 15.000s
BeamMe: Space-Time: Create mesh in time                                3.003s > 3.000s
```
No idea why these fail, nothing changed for these tests.

.github/workflows/testing_performance.yml
isteinbrecher: There is no reason for this, we can put this into the standard workflow.

isteinbrecher:
@davidrudlstorfer I had a look at this, and I have no idea why the performance tests change... The only ones that seem to work are the ones in the protected environment.

For now, I left the performance tests in the protected environment (but removed the Coreform dependency). I don't have a better idea than to merge this and see if the performance tests work as expected on main (some reference times need to be adapted due to the changed size of the solid mesh, but the other tests should hopefully pass as expected).

davidrudlstorfer:
Is this already ready for review? What did now change that the timings suddenly work again?

isteinbrecher:
> Is this already ready for review? What did now change that the timings suddenly work again?

Yes this is now ready. I don't know what changed, but now the test times seem better. I adapted some of the reference times, this should now work hopefully. We can also close #639. I think we just have to see how the performance tests behave in the future and maybe adapt the times again.

One side note, seems like `netCDF` has hot been updated to `numpy2.5` yet, as there are now new warnings. I added this to #562.

tests/other/test_other_create_structured_exodus_hex_mesh.py
```diff
@@ -0,0 +1,34 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+"""Test the functionality to create structured hex meshes in exodus format."""
```
davidrudlstorfer, line 22: Should the exodus maybe also carry over in the file name?

test_other_create_structured_exodus_hex_mesh?

tests/performance/test_performance_beamme.py
```diff
@@ -172,80 +84,23 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
+    return create_exodus_input_file(shared_tmp_path, n_hex8=100, n_hex27=50)
 
 
-@pytest.mark.coreform
 @pytest.mark.parametrize(
     ("log_name", "full_import", "expected_time"),
     [
-        ("BeamMe: Load solid mesh (no full import)", False, 2.5),
-        ("BeamMe: Load solid mesh (full import)", True, 3.5),
+        ("Load solid mesh (no full import)", False, 1.0),
```
davidrudlstorfer, line 104: Thanks for removing all the names and updating the times:)

tests/performance/test_performance_beamme.py
```diff
@@ -172,80 +84,23 @@ def create_beam_mesh(n_x, n_y, n_z, n_el):
...
+    """Create a large solid input file containing two blocks (hex8 and hex27) in exodus
+    format.
+
+    In total the mesh contains:
+    - 2,060,602 nodes
+    - 1,000,000 hex8 elements
+    - 125,000 hex27 elements
     """
-    pass
+    return create_exodus_input_file(shared_tmp_path, n_hex8=100, n_hex27=50)
```
davidrudlstorfer: This doesn't make too much sense to me because the numbers are contradicting.

Maybe rename the variable to n_hex8 per dir? Or hand over the total amount of elements and then do the cubit root (and raise an error if it does not work)

But the current way is a bit complicated to say we have 1mio elements and then set n_hex8 to 100

tests/create_structured_hex_mesh.py
```diff
@@ -0,0 +1,320 @@
...
+    points = np.concatenate([points_1, points_2])
+
+    # Combine the node sets and adjust the node indices.
+    node_sets = {
+        name: np.concatenate([node_sets_1[name], node_sets_2[name] + len(points_1)]) + 1
+        for name in node_sets_1
+    }
+
+    with netCDF4.Dataset(exo_path, "w") as nc:
+        # Create the actual exodus file
```
davidrudlstorfer: this comment could move above the code line

davidrudlstorfer (commented):
Really cool that we no longer need coreform for the performance tests!

Just a few small comments:)

tests/performance/test_performance_beamme.py
isteinbrecher: Nice catch, that makes total sense. I renamed the parameters to `n_per_dir_hex8` and `n_per_dir_hex27`.

tests/other/test_other_create_structured_exodus_hex_mesh.py
isteinbrecher: Good point, I changed it.

davidrudlstorfer (dismissed):
Nice :) thanks for the last changes

davidrudlstorfer (approved):

PR #678, opened 2026-07-10
Convert notebooks to py:percent format

isteinbrecher:
This allows for a much easier tracking of changes to the notebooks in git/GitHub, see [jupytext](https://jupytext.org/formats/scripts/).

Follow up PR to #529


davidrudlstorfer:
This would need a rebase on latest main for merging/reviewing:)

isteinbrecher:
This should now be ready.

.pre-commit-config.yaml
davidrudlstorfer: This can then be removed right? I think we do not have any other jupyter notebooks

pyproject.toml
davidrudlstorfer, line 22: Can this then also go? Probably this is then a requirement of jupytext itself right?

Then jupytext should probably just move to the regular requirements rather than dev. Back then we said that we wanna be able to execute the jupyter stuff with a regular install

davidrudlstorfer (commented):
Should the readme be updated (where "notebook" is written?)?

.pre-commit-config.yaml
isteinbrecher: Yes, this can go.

pyproject.toml
isteinbrecher: Good point, I moved `jupytext` to the main dependencies. The other ones have so stay as is. `jupytext` is basically just an extra layer that allows to store the notebooks in pure python format. Internally it converts it to `ipynb` structure and then passes this structure to the other libraries.

examples/utils/__init__.py
davidrudlstorfer, line 1: Why do we suddenly need this?

We did not add this on purpose before (but unfortunately I do not remember anymore why we did that...)

I checked the coverage report and it does not pop up there so it is fine. But I would still prefer not creating "modules" outside of the src folder.

Before this PR we also did not need this

examples/example_1_finite_rotations.ipynb
```diff
@@ -1,487 +0,0 @@
...
-    "\n",
-    "pv.set_jupyter_backend(\"trame\")\n",
-    "\n",
-    "\n",
-    "# Utility functionality for this example\n",
-    "from utils.example_1_utils import (\n",
-    "    PyVistaPlotter,\n",
-    "    add_cube_plot,\n",
-    "    print_matrix,\n",
-    "    print_rotation_matrix,\n",
```
davidrudlstorfer, line 233: Here we were able to import the file without the init file

examples/utils/__init__.py
davidrudlstorfer: Is the utility file still used somewhere? I was not able to find it quickly

website/docs/prepare_docs.py
```diff
@@ -45,13 +49,20 @@ def prepare_docs():
...
         shutil.rmtree(examples_target_dir)
-    file_extensions_to_copy = {".ipynb", ".py"}
-    for file in examples_source_dir.rglob("*"):
-        if file.is_file() and file.suffix.lower() in file_extensions_to_copy:
-            rel_path = file.relative_to(examples_source_dir)
+    for file in examples_source_dir.rglob("*.py"):
+        # Copy all python files in the example directory. The notebooks are directly
+        # located in ./examples and are in py:percent format. We need to convert them
+        # into ipynb format for the sphinx build.
+        rel_path = file.relative_to(examples_source_dir)
```
davidrudlstorfer, line 60: Ok now I understand :D So this get's converted on the fly so we can render it, nice comment:)

website/docs/prepare_docs.py
```diff
@@ -45,13 +49,20 @@ def prepare_docs():
...
+        if file.parent == examples_source_dir:
+            dest_path = examples_target_dir / rel_path.with_suffix(".ipynb")
+            dest_path.parent.mkdir(parents=True, exist_ok=True)
```
davidrudlstorfer, line 63: why do we need this line? isn't dest_path a file?

Then a folder with blabla.ipynb is created right?

davidrudlstorfer (commented):
Some last questions

examples/utils/__init__.py
isteinbrecher: If we don't have this `mypy` complains. Previously it worked, because `mypy` does not check `*.ipynb` files, but now they are `*.py` files and they are checked.

> Is the utility file still used somewhere? I was not able to find it quickly

Yes, it is still used in example 1, only the name of the utils file changed.

website/docs/prepare_docs.py
isteinbrecher: Yes, `dest_path` is a file, but we create `dest_path.parent`, so we can be sure that the directory exists.

Above, we delete the examples folder in the website directory. Then we have to create it again so we can store the converted notebooks. Since all notebooks are located in the same directory, only once is a directory created here.

examples/utils/__init__.py
davidrudlstorfer: If it is not too much work, could you send the mypy error?

I would somehow prefer to silence the mypy error rather than initializing a new module

website/docs/prepare_docs.py
davidrudlstorfer: Thanks for the explanation:)

examples/utils/__init__.py
isteinbrecher: Sure, that is the error
```
examples/utils/example_1_utils.py: error: Source file found twice under different module names: "example_1_utils" and "utils.example_1_utils"
examples/utils/example_1_utils.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
examples/utils/example_1_utils.py: note: Common resolutions include:
examples/utils/example_1_utils.py: note:     a) adding __init__.py somewhere,
examples/utils/example_1_utils.py: note:     b) using --explicit-package-bases or adjusting MYPYPATH
```

If it is easily possible I would also prefer not to have a module. But on the other hand I also don't want obscure visible comment to suppress the warning in the example.

davidrudlstorfer: So one thing could be to just turn mypy off for the examples?

Or chatgpt proposed 

```
[tool.mypy]
explicit_package_bases = true
```

we should be able to just put that setting into the pre-commit hook

isteinbrecher: If that works, that is a good solution. Could you add it and push to the branch?

davidrudlstorfer:
@isteinbrecher sorry that I was not able to look further into this on Friday, this took me much longer than originally anticipated

I pushed one commit on your branch proposing a different and much more elegant solution. In my opinion we could just move the util to the top level. This results in no need for changes to mypy (that was actually a bad idea), no necessary init files and no weird sys path injections in the testing anymore.

We could still rename the utils file in the examples directory to distinguish it better but I think this is a pretty solid workflow. For example we could rename it to "utils_example_1.py".

isteinbrecher:
@davidrudlstorfer thanks for the changes, looks good to me. I don't mind having all the files next to each other. From my point of view, we can merge this.

davidrudlstorfer:
Feel free to merge :)

davidrudlstorfer (approved):

PR #681, opened 2026-07-10
Make vtu mesh output the default for 4C

isteinbrecher:
This PR proposes to change the default mesh format for 4C input files to `vtu`.

@davidrudlstorfer I thought about the `auto` option, but I am not sure if this really makes sense as it is quire arbitrary. If the user wants a pure yaml output for debugging purposes, one can simply activate it.

Note, this should work fine for all discretizations, except NURBS. Furthermore, this might be a breaking change for some downstream scripts.

### DIFF
--- a/tests/integration/test_integration_four_c_simulation.py
+++ b/tests/integration/test_integration_four_c_simulation.py
@@ -111,7 +111,8 @@ def run_four_c_test(tmp_path: Path) -> Callable:
         input_file: InputFile,
         n_proc: int = 2,
         restart: list[int | str | None] = [None, None],
-        **kwargs,
+        mesh_format: str | None = None,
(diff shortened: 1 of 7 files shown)


tests/integration/test_integration_four_c_simulation.py
```diff
@@ -111,7 +111,8 @@ def _run_four_c_test(
         input_file: InputFile,
         n_proc: int = 2,
         restart: list[int | str | None] = [None, None],
-        **kwargs,
```
davidrudlstorfer, line 114: always nice when kwargs are gone :D

davidrudlstorfer (approved):
Nice change, thanks for thinking about the auto function, I will just add the yaml string to my project :)

PR #688, opened 2026-08-28
Improve comparison of reference files for failed tests

isteinbrecher:
This PR reworks how we open the VSCode comparison for failed tests.

This now works for input files with a `vtu` mesh format (the actually mesh data is not compared, but this could be implemented in the future). Previously this resulted in unreadable diffs.

Now the order of `reference` and `result` in `_assert_results_close` does not matter any more.

Also, this PR enables the diff of more type of objects than before. However, this can still be improved in the future by adding more suitable comparisons, e.g., `vtu` files, ...


tests/conftest_result_comparison.py
```diff
@@ -105,7 +107,14 @@ def _assert_results_close(
...
             )
         except AssertionError as error:
-            handle_failed_assertion(tmp_path, current_test_name, reference, result)
+            handle_failed_assertion(
+                tmp_path,
+                current_test_name,
+                reference if isinstance(reference, Path) else None,
+                converted_reference,
+                result if isinstance(result, Path) else None,
+                converted_result,
```
davidrudlstorfer:
```suggestion
                tmp_path = tmp_path,
                current_test_name = current_test_name,
                reference_path = reference if isinstance(reference, Path) else None,
                converted_reference = converted_reference,
                result_path = result if isinstance(result, Path) else None,
                converted_result = converted_result,
```

Although it duplicates a lot of information, maybe it would be easier to directly type the variables of the function here. For me the function call is not super clear. But this is not really important

tests/conftest_result_comparison.py
```diff
@@ -274,38 +285,50 @@ def handle_failed_assertion(
...
+
+        data_fourcipp.combine_sections(data)
+        data_fourcipp.dump(path)
+
+    # We try to convert the data to a FourCInput object and dump it to disk. If this
+    # fails, we don't do any comparisons here.
+    try:
+        if reference_path is None:
+            reference_path = tmp_path / (current_test_name + "_reference")
+            save_converted_data_to_file(cast(dict, converted_reference), reference_path)
```
davidrudlstorfer: Honestly I find that a bit rough that we just throw every type of data at FourCIPP :D

Cannot we determine beforehand (with some high probability) if it is a 4C input file? For example if one of the paths contains 4C.yaml or if there is some buzzword in the name?

The function is called with

`int | float | dict | list | np.ndarray | pv.UnstructuredGrid`

and on a quick guess, fourcipp only accepts dicts, right? Then we could at least change the check above to 


```suggestion
    # We try to convert the data to a FourCInput object and dump it to disk. If this
    # fails, we don't do any comparisons here.
    try:
        if reference_path is None and reference is dict:
            reference_path = tmp_path / (current_test_name + "_reference")
            save_converted_data_to_file(cast(dict, converted_reference), reference_path)
```

tests/conftest_result_comparison.py
isteinbrecher: Good point, I will add that.

tests/conftest_result_comparison.py
isteinbrecher: Yes, we basically throw the data at FourCIPP and if it sticks, we compare it. If not the `try` will take care of it and simply continue. I would not add this check here, because then we would end up at the compare call and the path will be non-existent. It should not matter due to the `try`, but now the error is actually thrown where changes are needed, i.e., in the `save_converted_data_to_file` function. If we want to add more functionality, it should be done there.

However, at least from my experience, 99% of the actually use cases for this comparison function are input files / meshes, so I am not sure if it is worth the effort to generalise this even further.

I updated the documentation in the code to better represent this.

tests/conftest_result_comparison.py
```diff
@@ -274,38 +285,59 @@ def handle_failed_assertion(
...
-
-    print(f"Result string saved to: '{result_path}'.")
-
-    # open VSCode diff tool if available
-    if shutil.which("code") is not None:
-        child = subprocess.Popen(
-            ["code", "--diff", result_path, reference],
-            stderr=subprocess.PIPE,
+        For now, we try to load the data with FourCIPP and store it to disk. In the case
+        that the inout file contains a vtu mesh reference, there will be a unstructured
```
davidrudlstorfer:
```suggestion
        that the input file contains a vtu mesh reference, there will be a unstructured
```

davidrudlstorfer (dismissed):
Thanks for the changes, only a typo

tests/conftest_result_comparison.py
isteinbrecher: Nice catch. Should be fixed. Could you reapprove for a final time?

davidrudlstorfer (approved):

PR #690, opened 2026-08-31
Remove not needed casts

isteinbrecher:
With the new version of vistools, a lot of casts can be removed from the logo function (some could even be removed without the new vistools version).


davidrudlstorfer (approved):
Thanks for the clean-up :)

